PR #8, opened 2022-07-15
Add custom playback screen to the example app

### LINKED ISSUE #3: Implement custom playback screen in the example app
### Description

The `example/` app is still missing the actual custom playback screen. Right now, it only shows a form to input the custom stream data but nothing is actually played (`play` button does nothing). So the actual playback screen must be implemented and start the stream configured in [`CustomPlaybackForm`](https://github.com/bitmovin/bitmovin-player-react-native/blob/main/example/src/screens/CustomPlaybackForm.tsx) as result to the `play` action. Stream data must be passed between screens as navigation parameters (see [react-navigation docs](https://reactnavigation.org/docs/params/)).

pogist:
closes #3 

### Preview
**Important:** In preview, the license key was omitted on purpose to avoid leaks. That's why it stays blank all the time.

custom_playback_preview

### DIFF
--- a/example/src/App.tsx
+++ b/example/src/App.tsx
@@ -1,12 +1,14 @@
 import React from 'react';
 import { NavigationContainer } from '@react-navigation/native';
 import { createNativeStackNavigator } from '@react-navigation/native-stack';
+import { SourceType } from 'bitmovin-player-react-native';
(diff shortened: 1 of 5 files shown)


cardoso (approved):
LGTM 👍

Issue #9, opened 2022-07-18
Add support for playback configuration


jonathanm-tkf:
### Description

Possibility to add custom properties to the playback such as isAutoplayEnabled, this would be a perfect way to autoPlay the video once it is loaded, without the need to handle event ``` player.play();```

### Implementation possibility

At the start of initialization, we are using react hook

```
  const player = usePlayer(
    {
      licenseKey: 'LICENCE_KEY',
    },
    {
      isAutoplayEnabled: true, // possibility to add more funcionalities startOffset, etc
    }
  );
```

### Possible validation 

![2022-07-18_14-49-30 (1)](https://user-images.githubusercontent.com/67608970/179599080-a9368432-d12b-4f7d-a797-e00332f09c43.gif)

### Questions

- When implementing the solution in some cases in android we have the following detail, regarding the buffering? which occurs in a non-constant way, only happens on android as it works perfectly well on ios.

![screen-20220718-150418 (3)](https://user-images.githubusercontent.com/67608970/179600098-f4de8696-60e5-44a4-9630-45e39e084d8f.gif)

![2022-07-18_15-11-24](https://user-images.githubusercontent.com/67608970/179599685-f9656a3a-8d2c-4476-8c18-43368b23aaa5.png)

I look forward to possible comments and/or suggestions, and I am happy to help in any way I can, thanks for the effort

Best

jonathanm-tkf:
Hi @pogist 

We can close this isse?

pogist:
@jonathanm-tkf Sure, I'll close it now. Also thanks a lot for your contribution, it is remarkable 🙏

PR #10, opened 2022-07-18
feat: add custom playback configuration

jonathanm-tkf:
This PR it's related to issue #9

thanks and regards


pogist:
Hello @jonathanm-tkf, thank you very much for your contribution 😁. That's a nice feature to have indeed, but there are a few issues with your code I need you to tweak before I can approve:

- `PlaybackConfig` should be part of `PlayerConfig`:
```typescript
export interface PlayerConfig {
  // ...other properties.

  /**
   * Configures playback behavior. A default `PlaybackConfig` is set initially.
   */
  playbackConfig?: PlaybackConfig;
}
```

- Remove the `playbackConfig` parameter from `Player.constructor`, `usePlayer` and `initWithConfig` (native sources);
- Just handle playback config initialization on `JsonConverter` and `RCTConvert+BitmovinPlayer`, not directly on the `PlayerModule`;
- Add as many props from the native `PlaybackConfig` object as possible, not just `isAutoplayEnabled`. Other common props between android/iOS are `isMuted` and `isTimeShiftEnabled`;

jonathanm-tkf:
Thanks, @pogist in principle I had thought about it that way but I didn't want to affect the current functionality, as soon as I have some time I'll try to implement it, if I have any doubt I'll comment on the issue :)

Regarding question

When implementing the solution in some cases in android we have the following detail, regarding the buffering? which occurs in a non-constant way, only happens on android as it works perfectly well on ios.

What would be the best way to solve this problem?

Thanks again

jonathanm-tkf:
@pogist I don't understand what you mean by this, how would it be better? Could you give me an example?

- Just handle playback config initialization on JsonConverter and RCTConvert+BitmovinPlayer, not directly on the PlayerModule;

jonathanm-tkf:
> @pogist I don't understand what you mean by this, how would it be better? Could you give me an example?
> 
> * Just handle playback config initialization on JsonConverter and RCTConvert+BitmovinPlayer, not directly on the PlayerModule;

Maybe you referred to something like that

```swift
  static func playerConfig(_ json: Any?) -> PlayerConfig? {
      let playerConfig = PlayerConfig()
      guard let json = json as? [String: Any?] else {
          return playerConfig
      }
      if let licenseKey = json["licenseKey"] as? String {
          playerConfig.key = licenseKey
      }
      // ...other logic
      if let playbackConfig = json["playbackConfig"] as? [String: Any?] {
          if let isAutoplayEnabled = playbackConfig["isAutoplayEnabled"] as? Bool {
              playerConfig.playbackConfig.isAutoplayEnabled = isAutoplayEnabled
          }
      }

      return playerConfig
  }

```

jonathanm-tkf:
Hi @pogist 

I have changed the implementation could you check it and let me know if it is to your liking?

Greetings.

jonathanm-tkf:
Thank you very much for the support @pogist, yes correct, and I totally agree I just added it as a new screen because I needed to test and not interfere with your original example, as soon as I get free I will be modifying and if you like maybe I can modify the readme with an example so it does not interfere with the BasicPlayback, that is what you think thank you very much.

jonathanm-tkf:
Hi @pogist how are you?
I caught up with your recommendations and suggestions if you have time to review them, let me know if you agree or need any updates.
Thanks and regards.

README.md
```diff
@@ -9,15 +9,21 @@ Official React Native bindings for Bitmovin's mobile Player SDKs.
...
-- [Contributing](#contributing)
+- [Bitmovin Player React Native](#bitmovin-player-react-native)
+  - [Installation](#installation)
+    - [Add package dependency](#add-package-dependency)
+    - [Setup iOS Player SDK](#setup-ios-player-sdk)
+    - [Setup Android Player SDK](#setup-android-player-sdk)
+  - [Getting Started](#getting-started)
+    - [Setting up a license key](#setting-up-a-license-key)
+      - [Configuring through code](#configuring-through-code)
+    - [Setting up a playback configurations](#setting-up-a-playback-configurations)
```
pogist, line 20: Please, rename this section. It has some mistakes.

```suggestion
    - [Setting up the playback configuration](#setting-up-the-playback-configuration)
```

README.md
```diff
@@ -9,15 +9,21 @@ Official React Native bindings for Bitmovin's mobile Player SDKs.
...
+      - [Configuring through code](#configuring-through-code-1)
+      - [Configuring `Info.plist`](#configuring-infoplist)
+      - [Configuring `AndroidManifest.xml`](#configuring-androidmanifestxml)
```
pogist, line 23: Those two topics are related to the `Setting up a license key` section, not the playback configuration.

README.md
```diff
@@ -198,6 +204,43 @@ const player = new Player({
...
+#### Configuring through code
+
+```typescript
+// Simply pass the `playbackConfig` property to `PlaybackConfig` when instantiating a player.
+
+// With hooks
+import { usePlayer } from 'bitmovin-player-react-native';
+const player = usePlayer({
+  playbackConfig: {
+    // Optionally isAutoplayEnabled: Specifies whether autoplay is enabled. Default is false.
```
pogist, line 220: Repeating the prop names in comments feels a little bit cumbersome here. You can just leave the description and default value.

```suggestion
    // Specifies whether the playback starts immediately after loading a source or not. Default is false.
```

README.md
```diff
@@ -198,6 +204,43 @@ const player = new Player({
...
+    isAutoplayEnabled: true,
+    // Optionally isMuted: Specifies if the player should start muted. Default is false.
```
pogist, line 222:
```suggestion
    // Specifies if playback starts muted. Default is false.
```

README.md
```diff
@@ -198,6 +204,43 @@ const player = new Player({
...
+    isMuted: true,
+    // Optionally isTimeShiftEnabled: Specifies if time shifting (during live streaming) should be enabled. Default is true.
```
pogist, line 224:
```suggestion
    // Specifies if time shift for live streamings should be enabled. Default is true.
```

README.md
```diff
@@ -198,6 +204,43 @@ const player = new Player({
...
+    isTimeShiftEnabled: true,
+    // Optionally isBackgroundPlaybackEnabled: Specifies if isBackgroundPlaybackEnabled should be enabled. Only in iOS. Default is false.
```
pogist, line 226:
```suggestion
    // Whether background playback is enabled or not. Default is false.
    // Only available for iOS.
```

README.md
```diff
@@ -198,6 +204,43 @@ const player = new Player({
 });
 ```
 
+### Setting up a playback configurations
+
+If you need more option to playback configuration you can add new parameter `playbackConfig`
+
+#### Configuring through code
```
pogist, line 211: You can remove this sub-section since it is the only one present under the main headline.

```suggestion
```

README.md
```diff
@@ -198,6 +204,43 @@ const player = new Player({
 });
 ```
 
+### Setting up a playback configurations
+
+If you need more option to playback configuration you can add new parameter `playbackConfig`
```
pogist, line 209:
```suggestion
If needed, the default player behavior can be configured through the `playbackConfig` key when initialized.
```

pogist (changes requested):

src/player.ts
```diff
@@ -38,6 +38,95 @@ export interface PlayerConfig {
    * ```
    */
   licenseKey?: string;
+  /**
+   * Configures playback behavior. A default `PlaybackConfig` is set initially.
```
pogist, line 42: Actually, we don't set a default value for `PlaybackConfig` so this doc comment could be a little misleading.

```suggestion
   * Configuration options for player's default playback behavior.
```

src/player.ts
```diff
@@ -38,6 +38,95 @@ export interface PlayerConfig {
...
+   */
+  playbackConfig?: PlaybackConfig;
+}
+
+/**
+ * Optionally PlaybackConfig Object used to configure a new `PlaybackConfig` instance.
```
pogist, line 48:
```suggestion
 * Configuration options for player's default playback behavior.
```

src/player.ts
```diff
@@ -38,6 +38,95 @@ export interface PlayerConfig {
...
+   *
+   * @example
+   * ```
+   * const player = new Player({
+   *   {
+   *     isAutoplayEnabled: true,
+   *   }
+   * })
+   * ```
+   */
```
pogist, line 64:
```suggestion
   * Whether the player starts playing automatically after loading a source or not. Default is `false`.
   *
   * @example
   * ```
   * const player = new Player({
   *   playbackConfig: {
   *     isAutoplayEnabled: true,
   *   },
   * });
   * ```
   */
```

src/player.ts
```diff
@@ -38,6 +38,95 @@ export interface PlayerConfig {
...
+   * Default is `false`.
+   *
+   * @example
+   * ```
+   * const player = new Player({
+   *   {
+   *     isMuted: true,
+   *   }
+   * })
+   * ```
```
pogist, line 78:
```suggestion
   * Whether the sound is muted on startup or not. Default value is `false`.
   *
   * @example
   * ```
   * const player = new Player({
   *   playbackConfig: {
   *     isMuted: true,
   *   },
   * });
   * ```
```

src/player.ts
```diff
@@ -38,6 +38,95 @@ export interface PlayerConfig {
...
+   * Default is `true`.
+   *
+   *  @example
+   * ```
+   * const player = new Player({
+   *   {
+   *     isTimeShiftEnabled: false,
+   *   }
+   * })
+   * ```
```
pogist, line 93:
```suggestion
   * Whether time shift / DVR for live streams is enabled or not. Default is `true`.
   *
   *  @example
   * ```
   * const player = new Player({
   *   playbackConfig: {
   *     isTimeShiftEnabled: false,
   *   },
   * });
   * ```
```

src/player.ts
```diff
@@ -38,6 +38,95 @@ export interface PlayerConfig {
...
+   * Default is `false`.
+   *
+   *  @example
+   * ```
+   * const player = new Player({
+   *   {
+   *     isPictureInPictureEnabled: true,
+   *   }
+   * })
+   * ```
```
pogist, line 127:
```suggestion
   * Whether the picture-in-picture mode option is enabled for iOS or not. Default is `false`.
   *
   *  @example
   * ```
   * const player = new Player({
   *   playbackConfig: {
   *     isPictureInPictureEnabled: true,
   *   },
   * });
   * ```
```

pogist (changes requested):

README.md
pogist: Please, move them back to the `Setting up a license key` section just the way it was.

jonathanm-tkf:
Hi @pogist how are you?
I caught up with your recommendations and suggestions if you have time to review them, let me know if you agree or need any updates.
Pido perdon debido a que he estado bastante distante del proyecto pero fueron semanas bastante duras en el trabajo, mil disculpas
Thanks and regards.

jonathanm-tkf:
Hi there @pogist 👋 

Some news with this integration?

Thanks for your support

pogist (approved):
Hi, @jonathanm-tkf 👋

Sorry I didn't get in touch with you in a while. I got caught up on work these last few weeks and totally forgot to leave my review the last time I checked, so I'm really sorry for it 🙏.

But I saw you applied the docs changes I asked and it seems really good to me, thanks for your effort 😁. We just need to wait for the native devs to review it now and then we'll be good to merge it. I believe we'll be able to include it on the next release already too. So again, thanks for your support and sorry about the delay.

jonathanm-tkf:
Thanks, @pogist thanks for your support I really appreciate it!!!

src/player.ts
```diff
@@ -28,6 +28,81 @@ export interface PlayerConfig extends NativeInstanceConfig {
    * ```
    */
   licenseKey?: string;
+  /**
+   * Configuration options for player's default playback behavior.
```
rolandkakonyi, line 32: Please always use the existing documentation from the native SDKs to ensure consistency.

```suggestion
   * Configures playback behaviour. A default PlaybackConfig is set initially.
```

src/player.ts
```diff
@@ -28,6 +28,81 @@ export interface PlayerConfig extends NativeInstanceConfig {
...
+   */
+  playbackConfig?: PlaybackConfig;
+}
+
+/**
+ * Configuration options for player's default playback behavior.
```
rolandkakonyi, line 38: Please use either the iOS or the Android SDKs documentation for public API documentation for this library

```suggestion
 * Configures the playback behaviour of the player.
```

src/player.ts
```diff
@@ -28,6 +28,81 @@ export interface PlayerConfig extends NativeInstanceConfig {
...
+   *     isTimeShiftEnabled: false,
+   *   },
+   * });
+   * ```
+   */
+  isTimeShiftEnabled?: boolean;
+  /**
+   * Optionally isBackgroundPlaybackEnabled: Specifies if isBackgroundPlaybackEnabled should be enabled.
+   *
+   * This param only work on iOS.
```
rolandkakonyi, line 80: Please always use the native SDKs documentation as they provide detailed explanation for given configurations which might be crucial for their usage.

```suggestion
   * Whether background playback is enabled or not.
   * Default is `false`.
   * 
   * When set to `true`, playback is not automatically paused
   * anymore when the app moves to the background.
   * When set to `true`, also make sure to properly configure your app to allow
   * background playback.
   *
   * On tvOS, background playback is only supported for audio-only content.
```

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
```diff
@@ -25,10 +26,38 @@ class JsonConverter {
...
+                config = PlayerConfig(key = json.getString("licenseKey"))
             }
-            return PlayerConfig()
+            if (json != null && json.hasKey("playbackConfig")) {
+                var playbackConfigJson = json.getMap("playbackConfig")
+                /**
+                 * Specifies whether autoplay is enabled.
+                 *
+                 * Default is `false`.
+                 */
```
matamegger, line 39: Please also apply @rolandkakonyi suggestion about the block comment for the android/kotlin part of this PR :)

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
```diff
@@ -25,10 +26,38 @@ class JsonConverter {
          */
         @JvmStatic
         fun toPlayerConfig(json: ReadableMap?): PlayerConfig {
+            var config = PlayerConfig()
             if (json != null && json.hasKey("licenseKey")) {
-                return PlayerConfig(key = json.getString("licenseKey"))
+                config = PlayerConfig(key = json.getString("licenseKey"))
             }
```
matamegger, line 32: I would suggest something like this:

```kotlin
if (json == null) return PlayerConfig()
            
val playerConfig = if (json.hasKey("licenseKey")) {
    PlayerConfig(key = json.getString("licenseKey"))
} else {
    PlayerConfig()
}
````

The suggestion also includes a more descriptive name for the "config" variable i.e. `playerConfig`

File: android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt [hunk 20]
```diff
@@ -25,10 +26,38 @@ class JsonConverter {
...
-            return PlayerConfig()
+            if (json != null && json.hasKey("playbackConfig")) {
```
matamegger, line 33: If my code suggestion is applied, the `null`-check shouldn't be needed anymore.

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
```diff
@@ -25,10 +26,38 @@ class JsonConverter {
...
+                var playbackConfigJson = json.getMap("playbackConfig")
+                /**
+                 * Specifies whether autoplay is enabled.
+                 *
+                 * Default is `false`.
+                 */
+                if(playbackConfigJson != null && playbackConfigJson.hasKey("isAutoplayEnabled")) {
```
matamegger, line 40: NIT (for all occurrences):
```suggestion
                if (playbackConfigJson?.hasKey("isAutoplayEnabled") == true) {
```

README.md
```diff
@@ -200,6 +200,43 @@ const player = new Player({
 });
 ```
 
+### Setting up a playback configurations
+
+If needed, the default player behavior can be configured through the `playbackConfig` key when initialized.
+
+
+```typescript
+// Simply pass the `playbackConfig` property to `PlaybackConfig` when instantiating a player.
```
matamegger, line 209:
```
// Simply pass the `playbackConfig` property to `PlayerConfig` when instantiating a player.
```
This should be the `PlayerConfig` or? 🤔 
(cc @pogist)

pogist: yes, that should be `PlayerConfig`

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
jonathanm-tkf: Hi @matamegger @pogist 
At this point, I understand the guidelines but it loses a bit of sense as it is making the instance of PlayerConfig but does not match it to any variable. 

```java
     @JvmStatic
     fun toPlayerConfig(json: ReadableMap?): PlayerConfig {
        var config = PlayerConfig()
        if (json != null && json.hasKey("licenseKey")) {
            config = PlayerConfig(key = json.getString("licenseKey"))
        }
        if (json != null && json.hasKey("playbackConfig")) {
            var playbackConfigJson = json.getMap("playbackConfig")
            if (playbackConfigJson?.hasKey("isAutoplayEnabled") == true) {
                config.playbackConfig.isAutoplayEnabled = playbackConfigJson.getBoolean("isAutoplayEnabled")
            }
            /**
             * Specifies if the player should start muted.
             *
             * Default is `false`.
             */
            if(playbackConfigJson != null && playbackConfigJson.hasKey("isMuted")) {
                config.playbackConfig.isMuted = playbackConfigJson.getBoolean("isMuted")
            }
            /**
             * Specifies if time shifting (during live streaming) should be enabled.
             *
             * Default is `true`.
             */
            if(playbackConfigJson != null && playbackConfigJson.hasKey("isTimeShiftEnabled")) {
                config.playbackConfig.isTimeShiftEnabled = playbackConfigJson.getBoolean("isTimeShiftEnabled")
            }
        }
        return config
    }
```

The original idea is to be able to modify the values of ```config.playbackConfig.isAutoplayEnabled``` with the value received via parameter from RN, in the case that the config variable does not exist, it would not take this value as it is not specified, is this correct? or is this something I am missing?

jonathanm-tkf (commented):
could you please check that everything is ok thank you very much @rolandkakonyi

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
matamegger: HI @jonathanm-tkf,
I don't quite understand your question.
I will try to elaborate on my suggested code snippet.

The function consists of many `if` statements at the moment. In each of the `if`s we check if `json` is `null`. In the case that `json` is indeed `null`, we basically do nothing and return `config`... which was initialized with a default `PlayerConfig`.
To avoid the many `null` checks I suggested the first line of my snippet:
```kotlin
if (json == null) return PlayerConfig()
```

Furthermore, `PlayerConfig.key` is a `val` and thus can not be changed after initializing the `PlayerConfig`. This means that we directly need to pass the `licenseKey` from the `ReadableMap` to the `PlayerConfig` constructor (if there is a `licenseKey` available in the `ReadableMap`).
The code in the PR already does this.

With my suggestion:
```kotlin
val playerConfig = if (json.hasKey("licenseKey")) {
    PlayerConfig(key = json.getString("licenseKey"))
} else {
    PlayerConfig()
}
```
We **always** assign a value for `playerConfig`. Depending on if there is a `licenseKey` in the `ReadableMap`, we either create a default `PlayerConfig` or a `PlayerConfig` with a player license key.

As mentioned in the first comment, I would like you to rename the current variable name `config` to `playerConfig`. i.e. there should be no local variable called `config` anymore :)

So if you replace:
```kotlin
var config = PlayerConfig()
if (json != null && json.hasKey("licenseKey")) {
    config = PlayerConfig(key = json.getString("licenseKey"))
}
```
with my suggestion, and of course every usage of `config` with `playerConfig`, everything should still work as intended by you. 🙂

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
```diff
@@ -25,10 +26,33 @@ class JsonConverter {
...
+            if (json != null && json.hasKey("playbackConfig")) {
+                var playbackConfigJson = json.getMap("playbackConfig")
+                if (playbackConfigJson?.hasKey("isAutoplayEnabled") == true) {
+                    config.playbackConfig.isAutoplayEnabled = playbackConfigJson.getBoolean("isAutoplayEnabled")
+                }
+                /**
+                 * Specifies if the player should start muted.
+                 *
+                 * Default is `false`.
+                 */
```
matamegger, line 42: I think @rolandkakonyi suggestion referred to all the code comment blocks in the function(s) 🙂

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
```diff
@@ -25,10 +26,33 @@ class JsonConverter {
...
+                if(playbackConfigJson != null && playbackConfigJson.hasKey("isMuted")) {
+                    config.playbackConfig.isMuted = playbackConfigJson.getBoolean("isMuted")
+                }
+                /**
+                 * Specifies if time shifting (during live streaming) should be enabled.
+                 *
+                 * Default is `true`.
+                 */
```
matamegger, line 50: Please also remove this one.

ios/RCTConvert+BitmovinPlayer.swift
```diff
@@ -15,6 +15,48 @@ extension RCTConvert {
         if let licenseKey = json["licenseKey"] as? String {
             playerConfig.key = licenseKey
         }
+        if let playbackConfig = json["playbackConfig"] as? [String: Any?] {
+            /**
+             * Specifies whether autoplay is enabled.
+             *
+             * Default is `false`.
+             */
```
matamegger, line 23: Also this comment is not needed, as well as the other block comments in this function.

matamegger (changes requested):
I tried to provide you more details on my previous feedback.
Let me know if I missunderstood your question or something is unclear 🙂

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
jonathanm-tkf: Hi @matamegger thank you for your support, now I've caught it

apologies for the misunderstanding, I have updated the PR with the resolution as above and have taken the liberty of also removing some nullity checks as you had mentioned above, let me know if any other corrections are needed.

Regards.

ios/RCTConvert+BitmovinPlayer.swift
```diff
@@ -15,6 +15,43 @@ extension RCTConvert {
...
         }
+        if let playbackConfig = json["playbackConfig"] as? [String: Any?] {
+            if let isAutoplayEnabled = playbackConfig["isAutoplayEnabled"] as? Bool {
+                playerConfig.playbackConfig.isAutoplayEnabled = isAutoplayEnabled
+            }
+            /**
+             * Specifies if the player should start muted.
+             *
+             * Default is `false`.
+             */
```
matamegger, line 26: Please remove this block comment.

ios/RCTConvert+BitmovinPlayer.swift
```diff
@@ -15,6 +15,43 @@ extension RCTConvert {
...
+            if let isMuted = playbackConfig["isMuted"] as? Bool {
+                playerConfig.playbackConfig.isMuted = isMuted
+            }
+            /**
+             * Specifies if time shifting (during live streaming) should be enabled.
+             *
+             * Default is `true`.
+             */
```
matamegger, line 34: Please remove this block comment.

ios/RCTConvert+BitmovinPlayer.swift
```diff
@@ -15,6 +15,43 @@ extension RCTConvert {
...
+            if let isTimeShiftEnabled = playbackConfig["isTimeShiftEnabled"] as? Bool {
+                playerConfig.playbackConfig.isTimeShiftEnabled = isTimeShiftEnabled
+            }
+            /**
+             * Specifies if isBackgroundPlaybackEnabled should be enabled.
+             *
+             * Default is `false`.
+             */
```
matamegger, line 42: Please remove this block comment.

ios/RCTConvert+BitmovinPlayer.swift
```diff
@@ -15,6 +15,43 @@ extension RCTConvert {
...
+            if let isBackgroundPlaybackEnabled = playbackConfig["isBackgroundPlaybackEnabled"] as? Bool {
+                playerConfig.playbackConfig.isBackgroundPlaybackEnabled = isBackgroundPlaybackEnabled
+            }
+            /**
+             * Specifies if isPictureInPictureEnabled should be enabled.
+             *
+             * Default is `false`.
+             */
```
matamegger, line 50: Please remove this block comment.

rolandkakonyi (changes requested):

matamegger (changes requested):
I think you missed something from my previous feedback.

I explicitly added comments to the places I intended you to remove before. Sorry, if that was not entirely clear. I think after this we are good to go from my side 🙂 

EDIT: Started with my review before @rolandkakonyi and thus did't saw his comment. Sorry for repeating on the same thing.

rolandkakonyi (approved):

matamegger (approved):
🎉

PR #19, opened 2022-09-13
Add full tvOS support

### LINKED ISSUE #18: Add full tvOS support
It's possible to use the current source code on an AppleTV app, but a few changes in podspec and the example app are necessary to guarantee the end user will get seamless integration with it.

- `tvos` is missing in podspec's [platforms](https://github.com/bitmovin/bitmovin-player-react-native/blob/development/RNBitmovinPlayer.podspec#L13) configuration;
  - This will cause any attempt to install the library on a tvOS target to fail.
- The library's [example app](https://github.com/bitmovin/bitmovin-player-react-native/tree/development/example) currently has no tvOS target in its Xcode project, only iOS;
- In order to get proper TV functionalities from React Native, the example app needs to replace `react-native` with [`react-native-tvos`](https://github.com/react-native-tvos/react-native-tvos).

pogist:
close #18


RNBitmovinPlayer.podspec
```diff
@@ -10,7 +10,7 @@ Pod::Spec.new do |s|
   s.license      = package["license"]
   s.authors      = package["author"]
 
-  s.platforms    = { :ios => "12.4" }
+  s.platforms    = { :ios => "12.4", :tvos => "12.4" }
```
rolandkakonyi, line 13: Just noticed now, why do we require the `12.4` versions instead of `12.0`?
We support from 12.0 from the iOS SDK side.

example/ios/BitmovinPlayerReactNativeExample-tvOS/AppDelegate.mm
```diff
@@ -0,0 +1,111 @@
...
+  _contextContainer = std::make_shared<facebook::react::ContextContainer const>();
+  _reactNativeConfig = std::make_shared<facebook::react::EmptyReactNativeConfig const>();
+  _contextContainer->insert("ReactNativeConfig", _reactNativeConfig);
+  _bridgeAdapter = [[RCTSurfacePresenterBridgeAdapter alloc] initWithBridge:bridge contextContainer:_contextContainer];
+  bridge.surfacePresenter = _bridgeAdapter.surfacePresenter;
+#endif
+
+  UIView *rootView = RCTAppSetupDefaultRootView(bridge, @"BitmovinPlayerReactNativeExample", nil);
+
+#if !TARGET_OS_TV
```
rolandkakonyi, line 46: Do we need this check as the target is tvOS only?

example/ios/Podfile
```diff
@@ -3,12 +3,14 @@ require_relative '../node_modules/@react-native-community/cli-platform-ios/nativ
...
 source 'https://github.com/bitmovin/cocoapod-specs.git'
+source 'https://github.com/react-native-tvos/react-native-tvos-podspecs.git'
 
-platform :ios, '12.4'
 install! 'cocoapods', :deterministic_uuids => false
 
-target 'BitmovinPlayerReactNativeExample' do
+def setup os
   inhibit_all_warnings!
+  platform os, '12.4'
```
rolandkakonyi, line 12: same here, why `12.4`?

rolandkakonyi (commented):
Code looks good!

During manual testing I found that the subtitle and audio buttons above the seek bar are not working.

This results in for example no way for changing subtitles.

RNBitmovinPlayer.podspec
pogist: That's because RN 0.69 [requires](https://github.com/facebook/react-native/commit/982ca30de079d7e80bd0b50365d58b9048fb628f) at least iOS `12.4` to work and the example app is using RN 0.69 on dev right now.

I set this requirement on `.podspec` by the time I worked on the example app maybe, so it was my mistake. I think we can safely set these versions to `12.0` since only applications require `12.4` to work with RN 0.69.

example/ios/Podfile
pogist: That's because the example app in [development](https://github.com/bitmovin/bitmovin-player-react-native/blob/development/example/package.json#L17) is using RN 0.69, which [requires iOS `12.4`](https://github.com/facebook/react-native/commit/982ca30de079d7e80bd0b50365d58b9048fb628f). So I just didn't change this value from what it was when I branched out.

In fact, this PR migrates the RN version to 0.68, because of [`react-native-tvos`](https://github.com/react-native-tvos/react-native-tvos/releases/tag/v0.68.2-6), so we could be using iOS `12.0` now. But it'd be nice to keep it as `12.4` because a new version of RN-tvos was released recently based on RN 0.69. And I should try to update the example app to use it later.

rolandkakonyi: I see. Is the min. deployment target for our RN lib documented on the README?

example/ios/BitmovinPlayerReactNativeExample-tvOS/AppDelegate.mm
rolandkakonyi: I see, thx for explaining.

RNBitmovinPlayer.podspec
rolandkakonyi: I see. Thanks.

rolandkakonyi:
@pogist I found this, sounds like the same issue: https://github.com/react-native-video/react-native-video/issues/2507

example/ios/Podfile
pogist: No, that's not documented on README but it'd be pretty good to have actually. I think we could setup a `Requirements` section similar to what [RN has](https://github.com/facebook/react-native#-requirements).

pogist:
@rolandkakonyi I found out this is actually a problem with `react-native-tvos` itself, so I've been in touch with the project's maintainer to fix this: https://github.com/react-native-tvos/react-native-tvos/pull/367 and the fix should be released anytime soon now.

example/ios/Podfile
rolandkakonyi: Yes. That would be great addition in a separate PR.

rolandkakonyi:
@pogist there was a release with the fix, see: https://github.com/react-native-tvos/react-native-tvos/pull/367#issuecomment-1264725169

pogist:
> @pogist there was a release with the fix, see: [react-native-tvos/react-native-tvos#367 (comment)](https://github.com/react-native-tvos/react-native-tvos/pull/367#issuecomment-1264725169)

Yeah, I saw it yesterday 🎉. I'm bumping this branch to this version and I'll push the commit in a moment. I'm just waiting for my player license to be renewed so I can test it a little better 😅

pogist:
@rolandkakonyi RN version is up to date now 👍

rolandkakonyi (approved):
@pogist Works seamlessly! Great job there figuring out the RN tvOS bug and getting it over the line! 🎉

PR #40, opened 2022-10-21
Add support for tweaks configuration

### LINKED ISSUE #38: Add support for tweaks configuration on player
### Description

The tweaks config is an incubator for experimental features. Such features are necessary to enable some advanced behaviors, such as passing custom cookies for every player request on iOS, for example.

The RN tweaks object should include a combination of iOS + Android tweaks, which are pretty different from one another in general; see the docs for the currently supported SDKs: [iOS](https://cdn.bitmovin.com/player/ios/3/docs/Classes/TweaksConfig.html) / [Android](https://cdn.bitmovin.com/player/android/3.24.2/docs/-player%20-module/com.bitmovin.player.api/-tweaks-config/index.html).

The tweaks config interface will look like this:

> ⚠️ Note that only [`disableSubtitleTracksWhileSeeking`](https://cdn.bitmovin.com/player/ios/3/docs/Classes/TweaksConfig.html#/c:@M@BitmovinPlayer@objc(cs)BMPTweaksConfig(py)disableSubtitleTracksWhileSeeking) and [`exoPlayerCache`](https://cdn.bitmovin.com/player/android/3.24.2/docs/-player%20-module/com.bitmovin.player.api/-tweaks-config/exo-player-cache.html) properties **won't be supported** because they're scheduled for deprecation.
```typescript
// Definition
interface TweaksConfig {
  // Common tweaks between iOS and Android
  timeChangedInterval?: number;

  // iOS-only tweaks (ignored on Android)
  isNativeHlsParsingEnabled?: boolean;
  isCustomHlsLoadingEnabled?: boolean;
  seekToEndThreshold?: number;
  playbackStartBehaviour?: 'relaxed' | 'aggressive';
  unstallingBehaviour?: 'relaxed' | 'aggressive';

  // Android-only tweaks (ignored on iOS)
  bandwidthEstimateWeightLimit?: number;
  devicesThatRequireSurfaceWorkaround?: {
    deviceNames?: string[]; // Mapped to an array of `DeviceName`s in Kotlin
    modelNames?: string[]; // Mapped to an array of `ModelName`s in Kotlin
  };
  languagePropertyNormalization?: boolean;
  localDynamicDashWindowUpdateInterval?: number;
  shouldApplyTtmlRegionWorkaround?: boolean;
  useDrmSessionForClearPeriods?: boolean;
  useDrmSessionForClearSources?: boolean;
  useFiletypeExtractorFallbackForHls?: boolean;
}

// Usage
const player = usePlayer({
  tweaksConfig: {
    // Select desired tweaks...
  },
});

player.load(/** ... **/);
player.play();
```

pogist:
close #38


android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
```diff
@@ -44,6 +46,50 @@ class JsonConverter {
...
+                        for (i in 0 until deviceNames.size()) {
+                            descriptions.add(DeviceName(deviceNames.getString(i)))
+                        }
+                    }
+                    if (modelNames != null) {
+                        for (i in 0 until modelNames.size()) {
+                            descriptions.add(ModelName(modelNames.getString(i)))
+                        }
+                    }
+                    playerConfig.tweaksConfig.devicesThatRequireSurfaceWorkaround = descriptions
```
matamegger, line 72: This does not look very kotlinish, but I also noticed that these collections from RN do not follow any "native" interface like `Collection` or `Iterable`...
Do you know if there is some project that extents RN with kotlin friendly extensions? 🤔 

It could look something like this
```suggestion
                    val deviceNames = devices?.getArray("deviceNames")
                        ?.toStringList()
                        ?.map { DeviceName(it) }
                        ?: emptyList()
                    val modelNames = devices?.getArray("modelNames")
                        ?.toStringList()
                        ?.map { ModelName(it) }
                        ?: emptyList()
                    playerConfig.tweaksConfig.devicesThatRequireSurfaceWorkaround = deviceNames + modelNames
```
```kotlin
private fun ReadableArray.toStringList() =  (0 until size()).map {
    getString(it)
}
```

src/tweaksConfig.ts
```diff
@@ -0,0 +1,151 @@
...
+   *
+   * @platform Android
+   */
+  devicesThatRequireSurfaceWorkaround?: {
+    /**
+     * A device name as reported by Build.DEVICE.
+     *
+     * @see Build.DEVICE: https://developer.android.com/reference/kotlin/android/os/Build.html#DEVICE--
+     */
+    deviceNames?: string[]; // Mapped to an array of `DeviceName`s in Kotlin
```
matamegger, line 92:
```suggestion
    deviceNames?: string[];
```
I am not sure if the comment is of value here 🤔 
Why do you think it is valuable? 🙂

src/tweaksConfig.ts
```diff
@@ -0,0 +1,151 @@
...
+    /**
+     * A model name as reported by Build.MODEL.
+     *
+     * @see Build.MODEL: https://developer.android.com/reference/kotlin/android/os/Build.html#MODEL--
+     */
+    modelNames?: string[]; // Mapped to an array of `ModelName`s in Kotlin
```
matamegger, line 98:
```suggestion
    modelNames?: string[];
```
Same as with the other property 🙂

matamegger (commented):
Looks good overall.
Some smaller changes and questions from my side.

src/tweaksConfig.ts
```diff
@@ -0,0 +1,151 @@
...
+   * nativeHlsParsingEnabled won’t have any effect.
+   *
+   * Default is true.
+   *
+   * @platform iOS
+   */
+  isCustomHlsLoadingEnabled?: boolean;
+  /**
+   * The threshold which will be applied when seeking to the end in seconds. This value will be used
+   * to calculate the maximum seekable time when calling player.seek(time:) orplayer.playlist.seek(source:time:),
```
rolandkakonyi, line 44:
```suggestion
   * to calculate the maximum seekable time when calling player.seek(time:) or player.playlist.seek(source:time:),
```

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
```diff
@@ -44,6 +46,50 @@ class JsonConverter {
...
+                    playerConfig.tweaksConfig.shouldApplyTtmlRegionWorkaround = tweaksConfigJson.getBoolean("shouldApplyTtmlRegionWorkaround")
+                }
+                if (tweaksConfigJson?.hasKey("useDrmSessionForClearPeriods") == true) {
+                    playerConfig.tweaksConfig.useDrmSessionForClearPeriods = tweaksConfigJson.getBoolean("useDrmSessionForClearPeriods")
+                }
+                if (tweaksConfigJson?.hasKey("useDrmSessionForClearSources") == true) {
+                    playerConfig.tweaksConfig.useDrmSessionForClearSources = tweaksConfigJson.getBoolean("useDrmSessionForClearSources")
+                }
+                if (tweaksConfigJson?.hasKey("useFiletypeExtractorFallbackForHls") == true) {
+                    playerConfig.tweaksConfig.useFiletypeExtractorFallbackForHls = tweaksConfigJson.getBoolean("useFiletypeExtractorFallbackForHls")
```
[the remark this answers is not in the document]
pogist: Sure, I'll do it.

src/tweaksConfig.ts
```diff
@@ -0,0 +1,151 @@
+/**
+ * This configuration is used as an incubator for experimental features. Tweaks are not officially
+ * supported and are not guaranteed to be stable, i.e. their naming, functionality and API can
+ * change at any time within the tweaks or when being promoted to an official feature and moved
+ * into its final configuration namespace.
+ */
+export interface TweaksConfig {
+  /**
+   * The frequency in seconds onTimeChanged is called with TimeChangedEvents.
+   * Default is 1. Minimum is 0.025.
```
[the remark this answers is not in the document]
pogist: I'll update the comment to reflect each platform's default value then. And I don't think we need to do much to handle them from the lib's perspective because the SDKs should do that already, and no default values are being defined from the library per se.

src/tweaksConfig.ts
pogist: Oh, I actually forgot to remove those comments when I finished writing the issue 😅. But I'll remove them, they add no value indeed. Thanks 🙂.

src/tweaksConfig.ts
pogist: I'll remove them. Thanks 🙂.

ios/RCTConvert+BitmovinPlayer.swift
```diff
@@ -32,6 +32,40 @@ extension RCTConvert {
                 playerConfig.playbackConfig.isPictureInPictureEnabled = isPictureInPictureEnabled
             }
         }
+        if let tweaksConfig = json["tweaksConfig"] as? [String: Any?] {
```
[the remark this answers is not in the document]
pogist: Sure, you're right. I'll do that. Thanks 🙂.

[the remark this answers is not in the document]
rolandkakonyi: Not a blocker. Could be a follow up PR.

src/tweaksConfig.ts
pogist: Regarding the default value, I think it's only a matter of making each platform's default value explicit on the docs. Because since the lib doesn't set any default values itself, this should still be handled by the SDKs natively.

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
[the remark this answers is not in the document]
pogist: Done 8178ae447fb4ca5d715bfec28482e8e86825d2bb.

src/tweaksConfig.ts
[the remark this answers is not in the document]
pogist: Done 14608f24c81d242c23ff08db9495cdde38c9d253.

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
pogist: I created a `toList` extension on `ReadableArray` following the code you shared. Just tried to generalize it a bit to support other types beyond `String` 🙂

3353908c7551bb03a0074c31e2aad60bc7ddee4e

src/tweaksConfig.ts
pogist: Done 59d57708d89c0faf4f5d5aca0a42ea2399153f98

src/tweaksConfig.ts
pogist: Done 59d57708d89c0faf4f5d5aca0a42ea2399153f98

src/tweaksConfig.ts
pogist: Also, the typo in this comment is fixed: https://github.com/bitmovin/bitmovin-player-react-native/pull/40/commits/59d57708d89c0faf4f5d5aca0a42ea2399153f98

android/src/main/java/com/bitmovin/player/reactnative/extensions/NativeCollections.kt
```diff
@@ -0,0 +1,18 @@
...
+inline fun <reified T> ReadableArray.toList(): List<T?> = (0 until size()).map { i ->
+    getDynamic(i).let {
+        when (T::class) {
+            Boolean::class -> it.asBoolean() as T
+            String::class -> it.asString() as T
+            Double::class -> it.asDouble() as T
+            Int::class -> it.asInt() as T
+            ReadableArray::class -> it.asArray() as T
+            ReadableMap::class -> it.asMap() as T
+            else -> null
```
matamegger, line 15: I would throw an exception in this case.
`IllegalArgumentException` maybe and stating that the input type is not supported. 
(I think) then we can also drop the nullability of the entries in the return list.

matamegger: Btw. By convention the file would be named after the Type we are extending i.e. `ReadableArrayExtensions`

matamegger (approved):
Looks good overall.
See my comments.

android/src/main/java/com/bitmovin/player/reactnative/extensions/NativeCollections.kt
pogist: I don't think it's a good idea to throw an exception for nullable items. The idea is that `ReadableArray` represents an array from javascript. And in javascript arrays can have multiple items of different types (including `null`). So the idea is that `ReadableArray` should represent the same structure in Java/Kotlin as it does in javascript.

What we can do in this case is to keep the nullable items as optionals (what we're already doing here) or drop the nullable items when parsing the structure from `ReadableArray` to `List`. But throwing an exception doesn't make sense to me since nullable items are common in JS arrays so they shouldn't represent an error.

pogist: I've updated the file name: 592480d714f9d4ff112239a2028f5c6ba146ed99

PR #45, opened 2022-10-27
Feat/style config

joornby-angel:
The start of the style config for the bitmovin player.  

1. I've only implemented the pieces that our company is going to use for the moment.  Would it be better for me to implement the whole spec?
1. For things that are different between operating systems, namely `userInterfaceType`, what would you like to do?  Right now I've just added the documentation that it's for iOS and tvOS only, but if there's a different way you wanted to handle it, I'd be glad to change it.

### DIFF
--- a/android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
@@ -46,6 +46,12 @@ class JsonConverter {
                     playerConfig.tweaksConfig = it
                 }
             }
+            if (json.hasKey("styleConfig")) {
+                var styleConfigJson = json.getMap("styleConfig")
+                if (styleConfigJson?.hasKey("isUiEnabled") == true) {
+                    playerConfig.styleConfig.isUiEnabled = styleConfigJson.getBoolean("isUiEnabled")
+                }
+            }
             return playerConfig
         }
 
--- a/android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
@@ -13,6 +13,7 @@ import com.bitmovin.player.api.media.subtitle.SubtitleTrack
 import com.bitmovin.player.api.source.Source
 import com.bitmovin.player.api.source.SourceConfig
 import com.bitmovin.player.api.source.SourceType
+import com.bitmovin.player.api.ui.StyleConfig
 import com.bitmovin.player.reactnative.extensions.getName
 import com.bitmovin.player.reactnative.extensions.toList
 import com.facebook.react.bridge.*
@@ -41,15 +42,14 @@ class JsonConverter {
                     playerConfig.playbackConfig = it
                 }
             }
+            if (json.hasKey("styleConfig")) {
+                toStyleConfig(json.getMap("styleConfig"))?.let {
+                    playerConfig.styleConfig = it
+                }
+            }
             if (json.hasKey("tweaksConfig")) {
                 toTweaksConfig(json.getMap("tweaksConfig"))?.let {
                     playerConfig.tweaksConfig = it
-                }
-            }
-            if (json.hasKey("styleConfig")) {
-                var styleConfigJson = json.getMap("styleConfig")
-                if (styleConfigJson?.hasKey("isUiEnabled") == true) {
-                    playerConfig.styleConfig.isUiEnabled = styleConfigJson.getBoolean("isUiEnabled")
                 }
             }
             return playerConfig
@@ -76,6 +76,23 @@ class JsonConverter {
                 playbackConfig.isTimeShiftEnabled = json.getBoolean("isTimeShiftEnabled")
             }
             return playbackConfig
+        }
+
+        /**
+         * Converts any JS object into a `StyleConfig` object.
+         * @param json JS object representing the `StyleConfig`.
+         * @return The generated `StyleConfig` if successful, `null` otherwise.
+         */
+        @JvmStatic
+        fun toStyleConfig(json: ReadableMap?): StyleConfig? {
+            if (json == null) {
+                return null
+            }
+            val styleConfig = StyleConfig()
+            if (json.hasKey("isUiEnabled")) {
+                styleConfig.isUiEnabled = json.getBoolean("isUiEnabled")
+            }
+            return styleConfig
         }
 
         /**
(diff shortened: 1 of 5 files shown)


matamegger (approved):
👍

rolandkakonyi (commented):
@joornby-angel The code looks good so far! Thank you for the contribution!

Would you mind also adding implementation for the other properties as well?

Namely:
- `playerUiCss`
- `supplementalPlayerUiCss`
- `playerUiJs`
- `scalingMode`

You can leave out the iOS-only `userInterfaceConfig` as that would be needed for advanced use cases only.

joornby-angel:
@rolandkakonyi I'll have the other properties you mentioned in today.  

As for the `userInterfaceConfig`, our company has need of that property.  We're building a react native layer UI instead of using bitmovin's JS/HTML UI, however we don't want to manage the subtitle view.  After reading through the iOS and Android native SDK documents, it seemed like the way I needed to approach the problem was to use iOS's `.subtitle` interface type, and use Android's `SubtitleView` component.  I have the bindings for the SubtitleView on another commit that I could pull in to this PR, but that is another OS dependent difference that I wasn't sure how to handle.  In my opinion, the react-native SDK has an opportunity to build a single interface to the player in a way that the native SDK's will have a hard time matching.  We could implement the `userInterfaceConfig` in the react-native component, and render a `SubtitleView` for android, and pass along the value to iOS.  That's one of Bitmovin's bigger value adds in my mind, having a single interface and similar implementations across all of the various form factors and operating systems.  However, it seem like it's above my authority to make that call.  

I could do one of the following things, or something else that y'all want:

1. Pull in the Android `SubtitleView` and write an example showing how to build a Player experience with a subtitle only Bitmovin UI.  The drawback here is to force every react-native leaf application depending on the library to duplicate the same logic.  
2. Pull out the `userInterfaceConfig` and let y'all figure out how you want to handle the advanced operating system use cases.  The drawback here is that I have to maintain a forked branch of the repository until y'all figure out what you want to do.  I would prefer to be using the Bitmovin deployed release, because that gives both our companies incentive to keep building out features.  
3. Build a react-native component layer to handle the discrepancies in the Android and iOS native Bitmovin SDK's.  I could come up with some suggestions, but it may be better if y'all figure out how you want to handle them and tell me what you want it to look like.

rolandkakonyi:
> @rolandkakonyi I'll have the other properties you mentioned in today.
> 
> As for the `userInterfaceConfig`, our company has need of that property. We're building a react native layer UI instead of using bitmovin's JS/HTML UI, however we don't want to manage the subtitle view. After reading through the iOS and Android native SDK documents, it seemed like the way I needed to approach the problem was to use iOS's `.subtitle` interface type, and use Android's `SubtitleView` component. I have the bindings for the SubtitleView on another commit that I could pull in to this PR, but that is another OS dependent difference that I wasn't sure how to handle. In my opinion, the react-native SDK has an opportunity to build a single interface to the player in a way that the native SDK's will have a hard time matching. We could implement the `userInterfaceConfig` in the react-native component, and render a `SubtitleView` for android, and pass along the value to iOS. That's one of Bitmovin's bigger value adds in my mind, having a single interface and similar implementations across all of the various form factors and operating systems. However, it seem like it's above my authority to make that call.
> 
> I could do one of the following things, or something else that y'all want:
> 
> 1. Pull in the Android `SubtitleView` and write an example showing how to build a Player experience with a subtitle only Bitmovin UI.  The drawback here is to force every react-native leaf application depending on the library to duplicate the same logic.
> 2. Pull out the `userInterfaceConfig` and let y'all figure out how you want to handle the advanced operating system use cases.  The drawback here is that I have to maintain a forked branch of the repository until y'all figure out what you want to do.  I would prefer to be using the Bitmovin deployed release, because that gives both our companies incentive to keep building out features.
> 3. Build a react-native component layer to handle the discrepancies in the Android and iOS native Bitmovin SDK's.  I could come up with some suggestions, but it may be better if y'all figure out how you want to handle them and tell me what you want it to look like.

@joornby-angel I think for now leaving out the `userInterfaceConfig` would be a reasonable solution to get this PR merged.

For supporting the subtitle only view we would need to discuss this internally. Feel free to open an issue here with your feedback and requirements as you posted here.

example/src/screens/CustomHtmlUI.tsx
```diff
@@ -0,0 +1,80 @@
...
+  console.log(header, JSON.stringify(obj, null, 2));
+}
+
+export default function BasicPlayback() {
+  useTVGestures();
+
+  const player = usePlayer({
+    styleConfig: {
+      playerUiCss:
+        'https://cdn.bitmovin.com/player/web/8/bitmovinplayer-ui.css',
```
joornby-angel, line 22: @rolandkakonyi This is basically the same as the default interface, is there an example bitmovin hosted version somewhere that does something slightly different?  The android example use a `file://` URI and commits it to the repo, but that didn't sit right with me for some reason.

rolandkakonyi: @joornby-angel We don't have anything other than the default ones hosted right now. If you could share a CSS file with some modifications that you would feel useful to demonstrate this feature, we could put it on our CDN and use it for this example application.

rolandkakonyi: @joornby-angel Thanks for the patience. You can use this URL and commit it to the repository for the example screen: `https://storage.googleapis.com/bitmovin-player-cdn-origin/player/ui/ui-customized-sample.css`

example/src/screens/CustomHtmlUI.tsx
```diff
@@ -0,0 +1,80 @@
...
+      playerUiJs: 'https://cdn.bitmovin.com/player/web/8/bitmovinplayer-ui.js',
```
rolandkakonyi, line 23: As mentioned above, you can use this URL for the example screen:
```suggestion
      playerUiCss:
        'https://cdn.bitmovin.com/player/web/8/bitmovinplayer-ui.css',
      playerUiJs: 'https://cdn.bitmovin.com/player/web/8/bitmovinplayer-ui.js',
      supplementalPlayerUiCss:
        'https://storage.googleapis.com/bitmovin-player-cdn-origin/player/ui/ui-customized-sample.css'
```

rolandkakonyi (approved):
Approving already as only the URL for the supplemental CSS needs to be added.

example/src/screens/CustomHtmlUI.tsx
rolandkakonyi: @joornby could you make this edit?

joornby: @rolandkakonyi Yes!  I was out this week, but I'll get to this first thing Monday morning.

joornby-angel:
@rolandkakonyi Back over to you.

PR #52, opened 2022-11-03
Add proper Picture in Picture support for Android/iOS

### LINKED ISSUE #37: Add docs on how to enable PiP mode for iOS and provide `AVAudioSession` config helper
### Description

Even though it's possible to set `PlaybackConfig.isPictureInPictureEnabled` to `true` on iOS, that alone isn't enough to fully enable PiP mode. It's also necessary to configure the `Background Modes` capability on the iOS app and set the global `AVAudioSession` category to `playback`.

For the `Background Modes` setup, an entry on README will be enough to guide the users on how to accomplish it since it must be done on the application side (not on the library).

And for configuring `AVAudioSession`, which must be done through native code, a helper module can be provided by the library in order to ease this process:
```typescript
import React, { useEffect } from 'react';
import { AudioSession } from 'bitmovin-player-react-native';

// Main application component
export const App = () => {
  useEffect(() => {
    // The desired audio session category can be set when mounting the app's UI
    AudioSession
      .setCategory('playback') // This method will be ignored on Android
      .catch((error) => {
        // If a native error gets thrown, the user must be capable of handling it just
        // like any other promise rejection.
        handleError(error);
      });
  });
  return /** ... **/;
};
```

pogist:
close #37

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,15 +17,16 @@ Official React Native bindings for Bitmovin's mobile Player SDKs.
     - [Setup Android Player SDK](#setup-android-player-sdk)
   - [Getting Started](#getting-started)
     - [Setting up a license key](#setting-up-a-license-key)
-      - [Configuring through code](#configuring-through-code)
-    - [Setting up a playback configurations](#setting-up-a-playback-configurations)
-      - [Configuring `Info.plist`](#configuring-infoplist)
-      - [Configuring `AndroidManifest.xml`](#configuring-androidmanifestxml)
+      - [Through code](#through-code)
+      - [Through `Info.plist`](#through-infoplist)
+      - [Through `AndroidManifest.xml`](#through-androidmanifestxml)
+    - [Setting up the playback configuration](#setting-up-the-playback-configuration)
     - [Accessing native `Player` instances](#accessing-native-player-instances)
     - [Listening to events](#listening-to-events)
     - [Enabling DRM protection](#enabling-drm-protection)
       - [Prepare hooks](#prepare-hooks)
     - [Adding external subtitle tracks](#adding-external-subtitle-tracks)
+    - [Enabling Picture In Picture mode](#enabling-picture-in-picture-mode)
   - [Contributing](#contributing)
 
 ## Platform Support
@@ -200,7 +201,7 @@ First of all, create a license key on the [Dashboard](https://bitmovin.com/dashb
 
 Then your license key can be either set from code or by configuring `Info.plist` and `AndroidManifest.xml`.
 
-#### Configuring through code
+#### Through code
 
 ```typescript
 // Simply pass the `licenseKey` property to `PlayerConfig` when instantiating a player.
@@ -219,7 +220,24 @@ const player = new Player({
 });
 ```
 
-### Setting up a playback configurations
+#### Through `Info.plist`
+
+Add the following lines to the `<dict>` section of your `ios/Info.plist`:
+
+```xml
+<key>BitmovinPlayerLicenseKey</key>
+<string>ENTER-YOUR-LICENSE-KEY</string>
+```
+
+#### Through `AndroidManifest.xml`
+
+Add the following line to the `<application>` section of your `android/app/src/main/AndroidManifest.xml`:
+
+```xml
+<meta-data android:name="BITMOVIN_PLAYER_LICENSE_KEY" android:value="ENTER-YOUR-LICENSE-KEY" />
+```
+
+### Setting up the playback configuration
 
 If needed, the default player behavior can be configured through the `playbackConfig` key when initialized.
 
@@ -239,6 +257,14 @@ const player = usePlayer({
     // Whether background playback is enabled or not. Default is false.
     // Only available for iOS.
     isBackgroundPlaybackEnabled: true,
+    // Enable the Picture In Picture mode option on the player controls.
+    //
+    // Note iOS requires the audio session category of your app to be set to `playback` otherwise
+    // PiP mode won't work.
+    //
+    // Check out `Enabling Picture In Picture mode` section of README for more information
+    // on how to properly configure your app to support PiP.
+    isPictureInPictureEnabled: true,
   },
 });
 
@@ -251,25 +277,9 @@ const player = new Player({
     isMuted: true,
     isTimeShiftEnabled: true,
     isBackgroundPlaybackEnabled: true,
+    isPictureInPictureEnabled: true,
   },
 });
-```
-
-#### Configuring `Info.plist`
-
-Add the following lines to the `<dict>` section of your `ios/Info.plist`:
-
-```xml
-<key>BitmovinPlayerLicenseKey</key>
-<string>ENTER-YOUR-LICENSE-KEY</string>
-```
-
-#### Configuring `AndroidManifest.xml`
-
-Add the following line to the `<application>` section of your `android/app/src/main/AndroidManifest.xml`:
-
-```xml
-<meta-data android:name="BITMOVIN_PLAYER_LICENSE_KEY" android:value="ENTER-YOUR-LICENSE-KEY" />
 ```
 
 ### Accessing native `Player` instances
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Official React Native bindings for Bitmovin's mobile Player SDKs.
     - [Enabling DRM protection](#enabling-drm-protection)
       - [Prepare hooks](#prepare-hooks)
     - [Adding external subtitle tracks](#adding-external-subtitle-tracks)
-    - [Enabling Picture In Picture mode](#enabling-picture-in-picture-mode)
+    - [Enabling picture-in-picture mode](#enabling-picture-in-picture-mode)
   - [Contributing](#contributing)
 
 ## Platform Support
@@ -51,14 +51,14 @@ Please note that browsers and other browser-like environments such as webOS and 
 
 Features of the native mobile Player SDKs are progressively being implemented in this React Native library. The table below summarizes the current state of the main Player SDK features.
 
-| Feature | State |
-| --- | --- |
-| Playback of DRM-protected assets | :white_check_mark:  Available since v0.2.0 |
-| Subtitles & Captions | :white_check_mark:  Available since v0.2.0 |
-| Advertising | :gear:  In progress, Q4 2022 |
-| Playlist API | :x:  Not available |
-| Offline Playback | :x:  Not available |
-| Analytics | :x:  Coming Q1 2023 |
+| Feature                          | State                                     |
+| -------------------------------- | ----------------------------------------- |
+| Playback of DRM-protected assets | :white_check_mark: Available since v0.2.0 |
+| Subtitles & Captions             | :white_check_mark: Available since v0.2.0 |
+| Advertising                      | :gear: In progress, Q4 2022               |
+| Playlist API                     | :x: Not available                         |
+| Offline Playback                 | :x: Not available                         |
+| Analytics                        | :x: Coming Q1 2023                        |
 
 ## Installation
 
@@ -276,12 +276,12 @@ const player = usePlayer({
     // Whether background playback is enabled or not. Default is false.
     // Only available for iOS.
     isBackgroundPlaybackEnabled: true,
-    // Enable the Picture In Picture mode option on the player controls.
+    // Enable the picture-in-picture mode option on the player controls.
     //
     // Note iOS requires the audio session category of your app to be set to `playback` otherwise
     // PiP mode won't work.
     //
-    // Check out `Enabling Picture In Picture mode` section of README for more information
+    // Check out `Enabling picture-in-picture mode` section of README for more information
     // on how to properly configure your app to support PiP.
     isPictureInPictureEnabled: true,
   },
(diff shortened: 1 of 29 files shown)


README.md
```diff
@@ -505,6 +515,102 @@ The supported `PlayerView` events for subtitles are:
...
+      // Handle any native error that might occur during this process.
+      handleError(error);
+    });
+  });
+  // ...
+  return /* ... */;
+};
+```
+
+This is step is required in order to properly enable background reproduction of audio content on iOS. Without it, the picture-in-picture option appears on the player UI but doesn't work.
```
rolandkakonyi, line 577:
```suggestion
This is step is required in order to properly enable background playback on iOS. Without it, the Picture in Picture option appears on the player UI but has no effect when used.
```

ios/RNPlayerView+PlayerListener.swift
```diff
@@ -1,6 +1,6 @@
 import BitmovinPlayer
 
-extension RNPlayerView: PlayerListener {
+extension RNPlayerView: PlayerListener, UserInterfaceListener {
```
rolandkakonyi, line 3: Could you either rename this file or even better extract the conformance to `UserInterfaceListener` to another `extension` in another file?

Right now this is inconsistent having both `PlayerListener` and `UserInterfaceListener` in the same file while it is named `+PlayerListener.swift`

rolandkakonyi (commented):
Manual testing looks good!

ios/RNPlayerView+PlayerListener.swift
pogist: Sure, it makes sense. I'll extract it to another file then 🙂.

README.md
```diff
@@ -23,15 +23,16 @@ Official React Native bindings for Bitmovin's mobile Player SDKs.
...
+      - [Through code](#through-code)
+      - [Through `Info.plist`](#through-infoplist)
+      - [Through `AndroidManifest.xml`](#through-androidmanifestxml)
+    - [Setting up the playback configuration](#setting-up-the-playback-configuration)
     - [Accessing native `Player` instances](#accessing-native-player-instances)
     - [Listening to events](#listening-to-events)
     - [Enabling DRM protection](#enabling-drm-protection)
       - [Prepare hooks](#prepare-hooks)
     - [Adding external subtitle tracks](#adding-external-subtitle-tracks)
+    - [Enabling picture-in-picture mode](#enabling-picture-in-picture-mode)
```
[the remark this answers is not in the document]
pogist: Done 5d572ac1a2a5b82c39a57593aff5ffd62689bdeb

README.md
pogist: Done 5d572ac1a2a5b82c39a57593aff5ffd62689bdeb

ios/RNPlayerView+PlayerListener.swift
pogist: Done 2ccbc49ef1d3f53762697396b4065fe693eab5c1

rolandkakonyi (approved):

matamegger (commented):
Cool feature 🧊 

Take care to not mix too many concerns 🙂

android/build.gradle
```diff
@@ -27,10 +27,10 @@ repositories {
 }
 
 android {
-    compileSdk 31
+    compileSdk 33
     defaultConfig {
-        minSdkVersion 21
-        targetSdkVersion 31
+        minSdkVersion 16
+        targetSdkVersion 33
```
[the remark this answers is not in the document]
pogist: Yeah, you’re right. It’d surely be better to do these version changes in another PR but I needed the target sdk to be at least `33` to access one of the new PiP parameters: `setExpandedAspectRatio`. Then I took the opportunity also to fix the min sdk that was set to `21` but should be `16`, which is currently the min version supported by the Android Player SDK and hasn't been fixed until now.

From now on, I’ll leave these version changes in different PRs since this case was pretty much an exception.
Thanks for the feedback 🙂

android/src/main/java/com/bitmovin/player/reactnative/RNBasePlayerView.kt
```diff
@@ -0,0 +1,241 @@
...
+import androidx.appcompat.app.AppCompatActivity
+import com.bitmovin.player.PlayerView
+import com.bitmovin.player.api.Player
+import com.bitmovin.player.api.ui.PictureInPictureHandler
+import com.facebook.react.bridge.LifecycleEventListener
+import com.facebook.react.bridge.ReactApplicationContext
+
+@SuppressLint("ViewConstructor")
+open class RNBasePlayerView(val context: ReactApplicationContext): LinearLayout(context),
+    LifecycleEventListener, PictureInPictureHandler, View.OnLayoutChangeListener {
```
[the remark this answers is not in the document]
pogist: Actually, I created this `RNBasePlayerView` with the intention of removing this PiP handling logic from the `RNPlayerView`, which was getting really bloated. But I wasn't so sure how I should extract such behavior from it, so I just moved it to a parent class even though this is not the best solution 😅.

I'll try to extract the PiP handling into its own component (without the whole view logic) and let `RNPlayerView` use it directly. This way we can get rid of this base class and unnecessary inheritances too.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -44,6 +44,11 @@ class RNPlayerView(context: ReactApplicationContext) : LinearLayout(context) {
         if (playerView.parent != this) {
             (playerView.parent as ViewGroup?)?.removeView(playerView)
             addView(playerView)
+            startBubblingEvents()
+        }
+        if (isPictureInPictureAvailable) {
+            enableSmoothTransitions()
+            playerView.setPictureInPictureHandler(this)
```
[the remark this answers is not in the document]
pogist: Yes, you're right, it does feel weird 😅. I'll move the PiP handling logic to a different component, just as I mentioned earlier in another comment, and hopefully get rid of this base class 😁.

android/src/main/java/com/bitmovin/player/reactnative/RNBasePlayerView.kt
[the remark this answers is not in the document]
matamegger: Thanks for the explanation, now I understand why you decided for the given design.

That sounds good 👍 The bitmovin android player library also provides a `DefaultPictureInPictureHandler` (you probably have seen it), but it is just a minimalistic variant that might fit for simple cases without any other UI components i.e. it does a bit less than what you have implemented already.
But still I think you could extract everything into your own dedicated PiP handler and just instantiate it directly in the view.
It would be cool if we could inject it from outside, so we are more flexible, but I am not sure if that's possible/feasible in this RN wrapper.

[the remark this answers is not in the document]
pogist: Yes, I checked the `DefaultPictureInPictureHandler` and even used it in my initial implementation, but I noticed it has a few glitches when animating PiP transitions. It basically lacks a few Android parameters described in [the docs](https://developer.android.com/develop/ui/views/picture-in-picture) to enable smoother transitions, especially for video content. So I decided to create a custom `PictureInPictureHandler` to include such parameters when available and provide the best possible PiP experience for the user's running version.

Also, I can try to make it injectable from the outside e.g. from other RN native modules, but I don't think we should focus on making it injectable from JavaScript code right now since it would require the lib to expose a lot of Android-specific APIs and considerably increase the complexity of this PR. If you're interested, we could schedule a kick-off meeting later just to discuss how this Android `PictureInPictureHandler` API could look like from JS and implement it in a future PR. What do you think?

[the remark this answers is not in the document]
pogist: I have moved RN pip handler to its own component and made it injectable from other native modules. See the commits: 452c4c2da3665ac877b026b32424529e8af0ab61,  e9c8533bec1fb59a3fed0836014acb88fd35b81b.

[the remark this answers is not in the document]
matamegger: > I don't think we should focus on making it injectable from JavaScript code right now

Perfect! That was also not my intention, PiP feels quite OS specific, I was not sure if it is super nice to handle this from JS, but you kind of confirmed that it is not too nice anyway :) 

Regarding a sync: What do you think would be the benefit with a JS PiP handler. What use cases would we enabled with that?

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -35,6 +44,57 @@ class RNPlayerView(context: ReactApplicationContext) : LinearLayout(context) {
...
+    /**
+     * Object that handles PiP mode changes in React Native.
+     */
+    var pictureInPictureHandler: RNPictureInPictureHandler? = null
+
+    /**
+     * Whether this view should pause video playback when activity's onPause is called.
+     * By default, `shouldPausePlayback` is set to false when entering PiP mode.
+     */
+    private var shouldPausePlayback: Boolean = true
```
matamegger, line 56: The name is very generic, a bit miss leading as we do not pause the player directly and I think it does not provide a good idea on when the player should be pause.

Something like the following might give a bit more context.

```
    private var shouldPausePlaybackOnActivityPause: Boolean = true
```
Feel free to come up with an own variable name :) 

I like that you started with `should` 👍

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -44,6 +104,73 @@ class RNPlayerView(context: ReactApplicationContext) : LinearLayout(context) {
         if (playerView.parent != this) {
             (playerView.parent as ViewGroup?)?.removeView(playerView)
             addView(playerView)
+            startBubblingEvents()
+        }
+        pictureInPictureHandler?.let {
+            if (it.isPictureInPictureAvailable) {
```
matamegger, line 110: I think this if check might be a bug.
What if PiP is disabled when `addPlayerView` is called, but later it is enabled 🤔 (and of course always supported by the system)

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
```diff
@@ -99,9 +124,16 @@ class RNPlayerViewManager(private val context: ReactApplicationContext) : Simple
...
      */
-    private fun attachPlayer(view: RNPlayerView, playerId: NativeId?) {
+    private fun attachPlayer(view: RNPlayerView, playerId: NativeId?, playerConfig: ReadableMap?) {
         Handler(Looper.getMainLooper()).post {
             val player = getPlayerModule()?.getPlayer(playerId)
+            // set picture in picture value based on playback config's option
+            playerConfig?.getMap("playbackConfig")?.getBoolean("isPictureInPictureEnabled")?.let {
+                // set PiP handler's `isPictureInPictureEnabled` value.
+                pictureInPictureHandler.isPictureInPictureEnabled = it
+                // set the view's PiP handler instance.
```
matamegger, line 134: Comment not needed

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
```diff
@@ -99,9 +124,16 @@ class RNPlayerViewManager(private val context: ReactApplicationContext) : Simple
      * @param view Target `RNPlayerView`.
      * @param playerId `Player` instance id inside `PlayerModule`'s registry.
      */
-    private fun attachPlayer(view: RNPlayerView, playerId: NativeId?) {
+    private fun attachPlayer(view: RNPlayerView, playerId: NativeId?, playerConfig: ReadableMap?) {
         Handler(Looper.getMainLooper()).post {
             val player = getPlayerModule()?.getPlayer(playerId)
+            // set picture in picture value based on playback config's option
```
matamegger, line 130: NIT: Questionable if this comment is needed. IMO it is not needed.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
```diff
@@ -19,14 +22,33 @@ class RNPlayerViewManager(private val context: ReactApplicationContext) : Simple
     /**
      * Exported module name to JS.
      */
-    override fun getName() = "NativePlayerView"
+    companion object {
+        const val name = "NativePlayerView"
+    }
```
matamegger, line 27: Does some component need to access this constant, besides the `getName` function directly below?

If not, why the change? Moving it to a (private) top level property might be a better option (?)

matamegger (commented):
Well done 👍 
Just some small remarks 🙂

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
pogist: Alright, that makes sense. I'll rename it to what you suggested, then. This name seems good enough 🙂.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
pogist: Yeah, your point makes sense 🤔. If PiP is later re-enabled, then this method won't be called again unless the React `PlayerView` is also re-rendered. In other words, `addPlayerView` is only called when the react component is mounting. Maybe exporting the PiP option as a [`@ReactProp`](https://reactnative.dev/docs/native-components-android#3-expose-view-property-setters-using-reactprop-or-reactpropgroup-annotation) can do the trick, since it's called anytime the prop changes. I'll do some testing to see if any of the assumptions are right and I'll get back with the results later. Thanks 🙂.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
pogist: Yeah, you're right. I thought they could help other people see what was my thinking process when working on this feature, but they're actually polluting the code more than helping. So good catch, I'll remove them.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
pogist: I created this constant so we could always reference the same name on both `getName` and `@ReactModule(name = ...)`, that's present at the top of the class. This last annotation is needed in order to make this module accessible from other native modules, and not necessarily by name, even though the annotation requires it.

pogist: That's the same approach we used with the `PlayerModule` for example.

https://github.com/bitmovin/bitmovin-player-react-native/blob/development/android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt#L9

android/src/main/java/com/bitmovin/player/reactnative/RNBasePlayerView.kt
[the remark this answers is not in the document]
pogist: I think the only benefit of handling PiP from JS would be to minimize the amount of native code the users have to write and maintain themselves. It’s usually better to keep as much app logic in JS as possible, at least from the user perspective, since it helps with testing and portability between RN versions. Other than that, I can’t see many benefits.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
pogist: That's why we can't make it `private` (it will cause a compilation error).

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
pogist: Done 9108c08059e126de8ff5e8da1404a4092c63bba7

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
```diff
@@ -99,9 +124,16 @@ class RNPlayerViewManager(private val context: ReactApplicationContext) : Simple
      * @param view Target `RNPlayerView`.
      * @param playerId `Player` instance id inside `PlayerModule`'s registry.
      */
-    private fun attachPlayer(view: RNPlayerView, playerId: NativeId?) {
+    private fun attachPlayer(view: RNPlayerView, playerId: NativeId?, playerConfig: ReadableMap?) {
         Handler(Looper.getMainLooper()).post {
             val player = getPlayerModule()?.getPlayer(playerId)
+            // set picture in picture value based on playback config's option
+            playerConfig?.getMap("playbackConfig")?.getBoolean("isPictureInPictureEnabled")?.let {
+                // set PiP handler's `isPictureInPictureEnabled` value.
```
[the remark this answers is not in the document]
pogist: Done 8da9939266acb2fc41cd337092888c511d6bcefb

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
pogist: Done 8da9939266acb2fc41cd337092888c511d6bcefb

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
pogist: Done 8da9939266acb2fc41cd337092888c511d6bcefb

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
pogist: So you’re right about setting `isPictureInPictureEnabled` late not changing the player’s state, but that’s not because of the `if` statement. Actually, that happens because we don’t connect the player’s config with the underlying component state on React inside `usePlayer`.

We use `useRef` to maintain the same player instance alive across renderings, but changing its state (like config, for example) won’t cause the UI to re-render. We would need `useState` or `useReducer` for this, but enabling this behavior in the current state of the library isn’t so easy from what I could see. I think we should move this work to a different PR.

Trying to change `useRef` with `useState` to keep track of the player’s config has led to some errors, and I still don’t know what’s causing it 😕.

matamegger: But we could still remove the if check?

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
matamegger:
```
private const val MODULE_NAME = "PlayerModule"

@ReactModule(name = MODULE_NAME)
class PlayerModule(private val context: ReactApplicationContext) : ReactContextBaseJavaModule(context) {
    /**
     * In-memory mapping from `nativeId`s to `Player` instances.
     */
    private val players: Registry<Player> = mutableMapOf()

    /**
     * JS exported module name.
     */
    override fun getName() = MODULE_NAME
```

For the `PlayerModule` it could look like this. I think this would be a cleaner approach than the companion object :)

matamegger: Very similar for the `NativePlayerView` :)

That's what I meant with private top level property.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
pogist: Yeah, sure 👍

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
pogist: Oh I see, that makes sense. I can update that, sure.

pogist: Done cf4a0e628007495bb8d91fd28d0962067b11d5bd

matamegger (approved):
💪

PR #41, opened 2022-10-25
Feature: programmatic track selection

joornby-angel:
Our company is migrating from one video player to bitmovin, and we have stored the user's preferred audio and subtitle track selection in our application.  We need these api's in order to keep backwards compatibility with our user's currently selected tracks.

PR #59, opened 2022-11-30
Feat/subtitle only UI

joornby-angel:
This is a PR to start a conversation.  I'm not sure about how Bitmovin would like to handle several of the API oddities here.  This should be considered as an MVP for a `Custom React Native UI with Bitmovin Subtitles` feature.  A lot of this code is a mixture of the required Bitmovin API's, and Angel Studios code that we are running in production.  This feature is also dependent on portions of https://github.com/bitmovin/bitmovin-player-react-native/pull/41 and https://github.com/bitmovin/bitmovin-player-react-native/pull/45, so that code has been duplicated here.


android/src/main/java/com/bitmovin/player/reactnative/RNSubtitleViewManager.kt
```diff
@@ -0,0 +1,136 @@
...
+    fun setApplyEmbeddedFontSizes(view: SubtitleView, applyEmbeddedFontSizes: Boolean) {
+        view.setApplyEmbeddedFontSizes(applyEmbeddedFontSizes)
+    }
+
+    /**
+     * Sets whether styling embedded within the cues should be applied.
+     * Enabled by default.
+     * Overrides any setting made with setApplyEmbeddedFontSizes.
+     */
+    @ReactProp(name = "applyEmbeddedStyles", defaultBoolean = true )
```
joornby-angel, line 51: I'm not familiar enough with the Bitmovin Android SDK, but it seems that the `applyEmbeddedStyles` is mutually exclusive with `userDefaultStyle`.  I think it may make more sense to provide an enum of a single property that sets the style, and font size appropriately.  What do y'all think?

example/metro.config.js
```diff
@@ -9,32 +10,41 @@ const modules = Object.keys({
...
-    }),
-  },
+      extraNodeModules: modules.reduce((acc, name) => {
+        acc[name] = path.join(__dirname, 'node_modules', name);
+        return acc;
+      }, {}),
+    },
+
+    transformer: {
+      babelTransformerPath: require.resolve('react-native-svg-transformer'),
```
joornby-angel, line 41: I put some SVG controls into the example so that you could control the player.  This is only in the example portion of the application, so it won't be shipped with the library.  Do y'all have a different preference for how I could show controls in the example that wouldn't require SVG assets?

example/src/screens/CustomSubtitleOnlyUI.tsx
```diff
@@ -0,0 +1,243 @@
...
+
+const Separator = () => <View style={styles.separator} />;
+
+const trackDisplayKeyExtractor = ({ identifier }: TrackDisplay) => identifier;
+
+export default function CustomSubtitleOnlyUI() {
+  useTVGestures();
+  const [playing, setPlaying] = useState<boolean>(false);
+
+  const styleConfig: StyleConfig =
```
joornby-angel, line 46: iOS has a built in subtitle only view, so we need to keep the UI enabled, but the interface type changed.

example/src/screens/CustomSubtitleOnlyUI.tsx
```diff
@@ -0,0 +1,243 @@
...
+          player={player}
+          style={styles.player}
+          onReady={onReady}
+          onEvent={onEvent}
+          onPlaying={onPlaying}
+          onPaused={onPaused}
+          onAudioChanged={onEvent}
+          onSubtitleChanged={onEvent}
+        />
+        <SubtitleView
```
joornby-angel, line 143: This only renders when the Platform.OS is android, because iOS has it's own setting.

example/src/screens/CustomSubtitleOnlyUI.tsx
```diff
@@ -0,0 +1,243 @@
+import React, { useCallback, useState } from 'react';
```
joornby-angel, line 1: This example file is effectively what Angel Studios has running in production.  I almost wonder if it would make more sense to put this logic in the PlayerView file itself.  I chose not to do that in order to avoid future painful merge conflicts when I pull in the latest release into our forked repo.  I also chose not to do it because I wasn't sure how to expose the android only subtitle configuration options, without making the API even more weird than it is currently.

rolandkakonyi:
Closing as this was already fix in #316 and released in `0.14.0`

PR #62, opened 2022-12-12
Feat/offline module

joornby-angel:


rolandkakonyi (commented):
Please add a README section to the root `README.md` explaining how to use the feature.

joornby-angel:
@rolandkakonyi Thanks for all the feedback!  I'm currently covered on a couple of things at work, but I'll get back to this early next week I hope.

android/src/main/java/com/bitmovin/player/reactnative/offline/OfflineManager.kt
```diff
@@ -0,0 +1,208 @@
...
+        contentManager.resume()
+    }
+
+    fun suspend() {
+        contentManager.suspend()
+    }
+
+    fun cancelDownload() {
+        suspend()
+        deleteAll()
```
matamegger, line 86: `deleteAll` will delete everything and probably is not what what somebody would expect when canceling downloads 🤔 

I think correctly, you will need to get the current `OfflineContentOptions` check for what is `Downloading`,`Suspended` or `Failed` and set the `action` of these options to `Delete`.

example/android/app/build.gradle
```diff
@@ -251,6 +251,9 @@ android {
 dependencies {
     implementation fileTree(dir: "libs", include: ["*.jar"])
     //noinspection GradleDynamicVersion
+
+    implementation "androidx.localbroadcastmanager:localbroadcastmanager:1.1.0"
```
matamegger, line 255: Please add a short comment on why this is needed. i.e. this is needed for the offline feature.

android/src/main/java/com/bitmovin/player/reactnative/offline/OfflineDownloadRequest.kt
```diff
@@ -0,0 +1,5 @@
+package com.bitmovin.player.reactnative.offline
+
+class OfflineDownloadRequest(val minimumBitrate: Int, val audioOptionIds: List<String?>?, val textOptionIds: List<String?>?) {
+
+}
```
matamegger, line 5:
```suggestion
data class OfflineDownloadRequest(
    val minimumBitrate: Int, 
    val audioOptionIds: List<String?>?, 
    val textOptionIds: List<String?>?
)
```

Also, what about the nullablility. I see that the lists might be null, but I don't think we should allow for null entries in the lists 🤔 
(cc @rolandkakonyi )

matamegger (commented):
Thanks for the contribution 💪 

Some minor remarks :)

android/src/main/java/com/bitmovin/player/reactnative/offline/OfflineManager.kt
```diff
@@ -0,0 +1,208 @@
...
+    /**
+     * Process the `OfflineDownloadRequest`.
+     * The `OfflineContentOptions` are stored in memory in this class because they can not be constructed to pass to the process call.
+     * We're copying the iOS interface for the `minimumBitrate` so that react-native will have a consistent interface.
+     *
+     */
+    fun process(request: OfflineDownloadRequest) {
+        if (contentOptions != null) {
+            contentOptions!!.videoOptions
+                .firstOrNull { option -> option.bitrate >= request.minimumBitrate }
```
matamegger, line 50: Not sure what is expected on iOS. The highest video quality, or the one closest to the `minimumBitrate`?

android/src/main/java/com/bitmovin/player/reactnative/offline/OfflineManager.kt
matamegger: After a talk to @rolandkakonyi I think this is fine.

android/src/main/java/com/bitmovin/player/reactnative/offline/OfflineManager.kt
matamegger: That's actually a question for @rolandkakonyi 🙂

rolandkakonyi: The next highest from the minimum. E.g. of there is a ladder of `[32k, 64k, 120k]` and minimum is set to `50k` then it should pick `64k` as that is the next one after the minimum requirement.
Disclaimer: this is completely handled by Apple on iOS, so not sure if some other factors are considered or not.

Also not sure what happens if you set minimum to `150k`, if it would select the highest or fail the download.

rolandkakonyi: Based on a super-quick test, it seems to select the highest quality in case of overshooting.

Note that `videoOptions` should be ordered for this `firstOrNull` to succeed.

matamegger: Thanks for the input.

Yes, the code should be changed to first sort by the bitrate and then find the first above the `minimumBitrate` or if there is none, the one with the highest bitrate should be chosen.

RobertArissen:
Do you have any information regarding the expected availability of this feature?

joornby-angel:
> Do you have any information regarding the expected availability of this feature?

I'm currently indisposed due to work deadlines.  If someone wants to pick this PR up and help get it across the finish line I would not be offended.  I've received tentative approval from work to come clean this up after I finish my current responsibilities, which I guess would be in the next week or two.

JoseGuerreroT:
Hello everyone, I am very interested in this feature. I would like to know what plans you have for this PR? 
Can you play DRM content offline?

bitbalint:
> Hello everyone, I am very interested in this feature. I would like to know what plans you have for this PR? Can you play DRM content offline? thanks

Hi @JoseGuerreroT - happy to share that this feature is now on our Q3 roadmap and planned to be completed in the next 1-2 months.

joornby-angel:
@bitbalint It's been a super long time, but I have finally freed up and started looking at this yesterday.  Expect some changes to this PR in the next week.  We've been running a version of this in production for a little bit, and I have some learnings I'll try to weave into the changes as well.

bitbalint:
That sounds very good @joornby-angel , thank you!

ios/OfflineModule.swift
```diff
@@ -0,0 +1,303 @@
...
+        offlineManagerHolders[nativeId]
+    }
+
+    /**
+     Creates a new `OfflineContentManager` instance inside the internal offline managers using the provided `config` object.
+     - @param config `Config` object received from JS.  Should contain a sourceConfig and location.
+     */
+    @objc func initWithConfig(_ nativeId: NativeId, config: Any?, resolver resolve: @escaping RCTPromiseResolveBlock, rejecter reject: @escaping RCTPromiseRejectBlock) {
+        bridge.uiManager.addUIBlock { [weak self] _, _ in
+            guard
```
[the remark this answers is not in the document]
joornby-angel: @rolandkakonyi I copied this pattern from the DRM config, which looks like: 

```
        bridge.uiManager.addUIBlock { [weak self] _, _ in
            guard
                self?.drmConfigs[nativeId] == nil,
                let fairplayConfig = RCTConvert.fairplayConfig(config)
            else {
                return
            }
            self?.drmConfigs[nativeId] = fairplayConfig
            self?.initConfigBlocks(nativeId, config)
        }
```

Do you have an example of how you want the self access to be simplified?  Also, do you want me to apply that to all of the UIBlock guard clauses in this file?

android/src/main/java/com/bitmovin/player/reactnative/offline/OfflineManager.kt
joornby-angel: @rolandkakonyi @matamegger Now that i'm back to this, are we still ok with deleting the offline content associated with the specific offline id that this `OfflineManager` is constructed with?  It's been a little while, but I believe I made this decision because that's what iOS seemed to do under the hood, so it felt like deleting everything for that id was a good spot to leave the user in.

example/android/app/build.gradle
joornby-angel: @matamegger Where should this comment exist?  

This is in the required setup for the android SDK found here: https://developer.bitmovin.com/playback/docs/working-with-offline-playback-android, so I _believe_ that means we should add a Feature Guide under the react native section for the offline stuff.  I don't think I can do that though.

ios/OfflineManagerHolder.swift
```diff
@@ -0,0 +1,114 @@
+#if !os(tvOS)
```
joornby-angel, line 1: @rolandkakonyi How do you want to handle the fact that many of these classes are not available on tvOS?  I've started down the road of excluded it like this, similar to how it's done elsewhere in the project, but it seems a little cumbersome to do it this way.  If this is how we want to do it, I need to do it for the whole OfflineModule class, as well as the loadOfflineSource function in the PlayerModule.

rolandkakonyi: @joornby-angel Yes, unfortunately this is the only way we have now. These APIs are exclusively available on iOS only.

android/src/main/java/com/bitmovin/player/reactnative/offline/OfflineManager.kt
rolandkakonyi: @joornby-angel I am fine with it if we need to make it behave the same for both.
If you want to pause a download, you suspend it. If you want to terminate it, you cancel it. On iOS canceling (=termination) means you delete any partially downloaded data as you can't continue after canceled. One must start a new download after cancelation.

@matamegger do you have any additional thoughts on this?

ios/OfflineModule.swift
```diff
@@ -0,0 +1,303 @@
...
+                let managerHolder = self?.offlineManagerHolders[nativeId]
+            else {
+                reject("BitmovinOfflineModule", "Could not find the offline module instance", nil)
+                return
+            }
+
+            let offlineSourceConfig = managerHolder.contentManager.createOfflineSourceConfig(restrictedToAssetCache: true)
+
+            resolve(RCTConvert.toJson(sourceConfig: offlineSourceConfig))
```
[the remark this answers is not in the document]
joornby-angel: @rolandkakonyi I'm only serializing the portions of the offline source that are serializable.  In the end it's the same as a source config once it gets to the JS layer.  I'm currently using this API not as a way to pass the source config through the JS layer, but instead to give enough information to the JS layer so the engineer can make a decision about which kind of source needs to be loaded.  

I'm fine getting rid of this API, but I need to know how to determine whether the source has been downloaded and is available for offline playback.  Android seems to give back the download state after it's been downloaded when you ask for available download options, but I can't seem to get any of the function calls on iOS for the `OfflineContentManager` to do anything once it has started the download process.  

Is there an API I should be calling to determine if the source is available for playback?

Here's the relevant JS layer call:

```
offlineContentManager
  ?.getOfflineSourceConfig?.()
  ?.then?.((sourceConfig) => {
    onEvent({
      type: 'getOfflineSourceConfig',
      sourceConfig,
    });
    if (sourceConfig != null) {
      onEvent('Loading the offline video');
      player.loadOfflineSource(offlineContentManager);
    } else {
      onEvent('Loading the standard source configuration');
      player.load(SOURCE_CONFIG);
    }
  });
```

src/offlineContentManager.ts
```diff
@@ -0,0 +1,211 @@
...
+        this.eventSubscription = new NativeEventEmitter(
+          OfflineModule
+        ).addListener(
+          'BitmovinOfflineEvent',
+          (data?: BitmovinNativeOfflineEventData) => {
+            if (this.nativeId !== data?.nativeId) {
+              return;
+            }
+
+            if (data.eventType === OfflineEventType.onCompleted) {
```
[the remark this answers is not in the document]
joornby-angel: @rolandkakonyi I'm adding the missing event types.  I originally didn't include them because they were iOS only in the case of `onCanceled` and `onLicenseExpired`.  I opted to go with the android equivalent naming for `onLicenseRenewed`, which is `onDrmLicenseUpdated`.  I'm going to mark the functions in comment as being iOS only.  Is there a specific way you'd like these callbacks named?

android/src/main/java/com/bitmovin/player/reactnative/offline/OfflineManager.kt
matamegger: Well summarized @rolandkakonyi. It should behave the same. So this is correct! 🙂

example/android/app/build.gradle
matamegger: Good catch. Indeed we should extend the getting started guide for react native 👍 

Please still add a code comment above the dependency. I think something like this explains the need for the particular dependency:
`// Only needed if the offline feature is used`

joornby-angel: @matamegger Sounds good.  I'll get the comment added.

src/offline/offlineContentManager.ts
```diff
@@ -0,0 +1,238 @@
...
+    return OfflineModule.downloadLicense(this.nativeId);
+  };
+
+  /**
+   * Releases the currently held offline license.
+   * When finished successfully data will be passed to the `OfflineContentManagerListener.onDrmLicenseUpdated`.
+   * Errors are transmitted to the `OfflineContentManagerListener.onError`.
+   */
+  releaseLicense = (): Promise<void> => {
+    return OfflineModule.releaseLicense(this.nativeId);
```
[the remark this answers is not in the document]
matamegger: Short: There is no equivalent.

I will come back to you, but this either is a no-op on iOS or we avoid the call for now, as the DRM licenses are automatically handled by the library itself per default.
@joornby-angel is there a particular need for this API call in your integration?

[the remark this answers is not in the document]
joornby-angel: @matamegger Nope, there isn't a particular need.  This function exists in Android, so I just needed to know what to do on iOS.  I'll comment that it's iOS only, and make sure that if it gets called on iOS that it won't fail. 👍

rolandkakonyi:
Hi @joornby-angel!
Thank you for your great contribution, we are taking these changes in-house to finish it up.

Please follow #196 in case you are interested.

PR #64, opened 2022-12-14
Analytics support

pogist:


android/build.gradle
```diff
@@ -50,8 +50,10 @@ android {
 
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlinVersion"
+    implementation "org.jetbrains.kotlin:kotlin-reflect:$kotlinVersion"
```
matamegger, line 53: I think it's not very nice to add this dependency. The reflect library is quite big and we should try to avoid it if possible.

matamegger (commented):
Just a thought, but wouldn't it make sense to have the analytics collector as a separate thing. So that somebody can explicitly add the dependency to their react native project.. nor not.

Would be cool if we could get rid of the reflect dependency... that's about 3MB

android/build.gradle
pogist: Yes, it would be nice if the analytics collector was separate, but it's been agreed that its initial release is meant to be part of this wrapper, just like we've done with Ads. And those can be split into different packages in the future.

And regarding the reflection API, I added it in order to simplify mapping the analytics [`customData` fields](https://github.com/bitmovin/bitmovin-analytics-collector-android/blob/main/collector/src/main/java/com/bitmovin/analytics/data/CustomData.kt) from JS <-> Kotlin.

The idea is to loop over all the properties of a JS object to dynamically populate the properties of a Kotlin `CustomData` and vice-versa. So for this, I needed a way to dynamically get/set the properties of a Kotlin object based on its name and the only way I could achieve this was by using reflection. See the [extension file](https://github.com/bitmovin/bitmovin-player-react-native/blob/a14bcca532b5d0c1e62e3b521f799ba799d72189/android/src/main/java/com/bitmovin/player/reactnative/extensions/Any.kt).

Do you think there's any way we could replicate the same behavior without `kotlin-reflect`? If so I would gladly switch the implementation.

pogist: I also found [this alternative](https://github.com/Kotlin/kotlinx.reflect.lite) to `kotlin-reflect` that states to be lighter so maybe it could be used instead. What do you think?

ios/AnalyticsModule.swift
```diff
@@ -0,0 +1,178 @@
...
+     */
+    @objc(getUserId:resolver:rejecter:)
+    func getUserId(
+        _ nativeId: NativeId,
+        resolver resolve: @escaping RCTPromiseResolveBlock,
+        rejecter reject: @escaping RCTPromiseRejectBlock
+    ) {
+        bridge.uiManager.addUIBlock { [weak self] _, _ in
+            guard let collector = self?.collectors[nativeId] else {
+                return
```
rolandkakonyi, line 173: Should we call `reject` here to not keep the promise unresolved in case collector is not found?

ios/AnalyticsModule.swift
```diff
@@ -0,0 +1,178 @@
...
+        _ nativeId: NativeId,
+        resolver resolve: @escaping RCTPromiseResolveBlock,
+        rejecter reject: @escaping RCTPromiseRejectBlock
+    ) {
+        bridge.uiManager.addUIBlock { [weak self] _, _ in
+            guard
+                let collector = self?.collectors[nativeId],
+                let customData = RCTConvert.toJson(analyticsCustomData: collector.getCustomData())
+            else {
+                return
```
rolandkakonyi, line 153: Should we call `reject` here to not keep the promise unresolved in case collector is not found?

ios/RCTConvert+BitmovinPlayer.swift
```diff
@@ -482,7 +483,94 @@ extension RCTConvert {
...
+        guard
+            let json = json as? [String: Any?],
+            let key = json["key"] as? String
+        else {
+            return nil
+        }
+        var config = BitmovinAnalyticsConfig(key: key)
+        if let playerKey = json["playerKey"] as? String {
+            config = BitmovinAnalyticsConfig(key: key, playerKey: playerKey)
+        }
```
rolandkakonyi, line 502: Let's not create this config two times in case there is `playerKey`.

```suggestion
        let config: BitmovinAnalyticsConfig
        if let playerKey = json["playerKey"] as? String {
            config = BitmovinAnalyticsConfig(key: key, playerKey: playerKey)
        } else {
            config = BitmovinAnalyticsConfig(key: key)
        }
```

README.md
```diff
@@ -700,6 +701,55 @@ The supported `PlayerView` events for ads are:
 
 You can check out a complete ads example in the [`example/`](https://github.com/bitmovin/bitmovin-player-react-native/tree/development/example) app.
 
+### Setting up analytics
```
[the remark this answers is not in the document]
bitbalint: That's a good idea.

README.md
```diff
@@ -700,6 +701,55 @@ The supported `PlayerView` events for ads are:
...
+    videoId: 'MyVideoId',
+    title: 'Art of Motion',
+    isLive: false,
+    ads: false, // Can be changed to `true` in case `advertisingConfig` is also present.
+    path: '/examples/basic_analytics',
+    customData1: 'Custom data field 1',
+    customData2: 'Custom data field 2',
+    customData3: 'Custom data field 3',
+    customData4: 'Custom data field 4',
+    customData5: 'Custom data field 5',
```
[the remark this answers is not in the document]
pogist: 5cda7a40652e31fa3ab73a72e538ad00d4f48803

README.md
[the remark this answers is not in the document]
pogist: 5cda7a40652e31fa3ab73a72e538ad00d4f48803

ios/AnalyticsModule.swift
pogist: 76321d08a16288322f832c2a23b0d6519de671ae

ios/AnalyticsModule.swift
pogist: 76321d08a16288322f832c2a23b0d6519de671ae

ios/RCTConvert+BitmovinPlayer.swift
pogist: 881aef79c3f8080852ea79b10e0279b981df2f46

android/build.gradle
pogist: I could achieve the same behavior using just java reflection: 7e0a03865a0714c85d1a0c03e6efe9d9bf47b970. What do you think about this solution? Would it have any potential problems?

My local tests worked just fine, though.

README.md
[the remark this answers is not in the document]
rolandkakonyi: Should we mention how many `customData` fields are there?
According to the source code of the iOS collector there are fields up to `customData30`
https://github.com/bitmovin/bitmovin-analytics-collector-ios/blob/main/BitmovinAnalyticsCollector/Classes/Collector/data/CustomData.swift

README.md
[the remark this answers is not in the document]
rolandkakonyi: I think this should be a markdown link, not just plain text.

README.md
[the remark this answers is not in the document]
rolandkakonyi: Fixed in 3ae99d8

rolandkakonyi (approved):

matamegger (approved):

PR #76, opened 2023-01-19
Update react native dependencies

matamegger:
At least the android part does not build (on some machines) when developing on the wrapper.

This is likely related to the changes in the distribution of react native starting with `0.71`.
Updating the dependencies of react native, and the forked project for ios TV resolved the failures during build.


strangesource (approved):
Tested & looks good. 👍

rolandkakonyi (approved):
Looks good!

PR #80, opened 2023-01-26
Fix passing of props to the BasicFullscreenHandling

matamegger:
There is a typescript error in the `BasicFullscreenHandling` screen, because of missing type information on the props.


rolandkakonyi (approved):
Thanks for fixing this!

PR #85, opened 2023-02-20
Extract event relay from PlayerView

matamegger:
There is a lot of repeating code regarding the event listener subscription. This PR extracts and compacts the even subscription into a dedicated component, which also can handle the clean up of the listeners.


android/src/main/java/com/bitmovin/player/reactnative/EventRelay.kt
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * List of events that should be relayed and their associated name.
+     */
+    forwardingEventClassesAndNameMapping: Map<KClass<out T>, String>,
+    /**
+     * Is called for every relayed event together with its associated name.
+     */
+    private val eventOutput: (String, Event) -> Unit
+) {
+    private val playerEventListenersMap = mutableMapOf<KClass<out T>, (T) -> Unit>()
```
strangesource, line 21: This component is very generic and not coupled to `player` events. Should we rename this property to something more generic as well?

As briefly discussed offline a `List` structure would suffice here.

strangesource (commented):
Great improvements! Just some NIT from my side.

android/src/main/java/com/bitmovin/player/reactnative/EventRelay.kt
matamegger: True. Missed that name in the process as it seems.

https://github.com/bitmovin/bitmovin-player-react-native/pull/85/commits/39b28c7e9444dd2e9a0298fdaf3eb9fb04ffd37a

PR #86, opened 2023-02-27
Integrate smoke test GitHub Actions workflow

rolandkakonyi:
This PR introduces a basic CI workflow that is run on pull requests and pushes to development to ensure the project can still be built.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,106 @@
+name: CI
+
+on:
+  pull_request:
+    branches: [development]
+    paths-ignore:
+      - '*.md'
+  push:
+    #    branches: [main, development]
+    paths-ignore:
+      - '*.md'
+jobs:
+  test-build-typescript:
+    name: Build Typescript
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup node and npm registry
+        uses: actions/setup-node@v3
+        with:
+          node-version: '16'
+          registry-url: 'https://registry.npmjs.org/'
+          cache: 'yarn'
+      - name: Install node_modules
+        run: yarn install --frozen-lockfile
+      - name: Install node_modules (example/)
+        run: yarn install --frozen-lockfile --cwd example
+      - name: Compile TypeScript
+        run: yarn typescript
+  test-build-android:
+    name: Build Android
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup node and npm registry
+        uses: actions/setup-node@v3
+        with:
+          node-version: '16'
+          registry-url: 'https://registry.npmjs.org/'
+          cache: 'yarn'
+      - name: Set up Java
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'zulu'
+          java-version: '11'
+      - name: Install node_modules
+        run: yarn install --frozen-lockfile
+      - name: Install node_modules (example/)
+        run: yarn install --frozen-lockfile --cwd example
+      - name: Set up Gradle cache
+        uses: gradle/gradle-build-action@v2
+        with:
+          cache-read-only: ${{ github.ref != 'refs/heads/main' }}
+      - name: Build Android example
+        run: |
+          cd example/android && ./gradlew assembleDebug --build-cache
+  test-build-ios:
+    name: Build iOS & tvOS
+    runs-on: macOS-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup node and npm registry
+        uses: actions/setup-node@v3
+        with:
+          node-version: '16'
+          registry-url: 'https://registry.npmjs.org/'
+          cache: 'yarn'
+      - uses: maxim-lobanov/setup-xcode@v1
+        with:
+          xcode-version: '14.1'
+      - name: Install node_modules
+        run: yarn install --frozen-lockfile
+      - name: Install node_modules (example/)
+        run: yarn install --frozen-lockfile --cwd example
+      - name: Install pods
+        run: yarn pods
+      - name: Build iOS example
+        run: |
+          set -o pipefail && cd example/ios && xcodebuild -workspace BitmovinPlayerReactNativeExample.xcworkspace -scheme BitmovinPlayerReactNativeExample -configuration Debug build CODE_SIGNING_ALLOWED='NO' | xcpretty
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,150 @@
+name: CI
+on:
```
matamegger, line 2: Can we please add an empty line between elements on the same level, that do not belong to each other?

```suggestion

on:
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,150 @@
...
+  pull_request:
+    branches: [development]
+    paths-ignore:
+      - '*.md'
+  push:
```
matamegger, line 7:
```suggestion

  push:
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,150 @@
...
+    branches: [development]
+    paths-ignore:
+      - '*.md'
+concurrency:
```
matamegger, line 11:
```suggestion

concurrency:
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,150 @@
...
+  group: ci-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+jobs:
+  test-build-typescript:
+    name: Build Typescript
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup node and npm registry
```
matamegger, line 21:
```suggestion

      - name: Setup node and npm registry
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,150 @@
...
+    paths-ignore:
+      - '*.md'
+  push:
+    branches: [development]
+    paths-ignore:
+      - '*.md'
+concurrency:
+  group: ci-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+jobs:
```
matamegger, line 14:
```suggestion

jobs:
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,150 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup node and npm registry
+        uses: actions/setup-node@v3
+        with:
+          node-version: '16'
+          registry-url: 'https://registry.npmjs.org/'
+          cache: 'yarn'
+      - name: Install node_modules
```
matamegger, line 27:
```suggestion

      - name: Install node_modules
```

I think with the suggestions until here, the pattern I am anticipating should be clear. Let me know if there are questions, or if you want to discuss the pattern in general (e.g. because you do not agree with it).

.github/workflows/ci.yml
```diff
@@ -0,0 +1,150 @@
...
+            .cocoapods-cache
+            example/ios/Pods
+          key: pods-${{ hashFiles('example/ios/Podfile.lock') }}
+          restore-keys: pods-
+      - name: Install pods
+        run: yarn pods
+        env:
+          CP_HOME_DIR: ${{ github.workspace }}/.cocoapods-cache
+      - name: Save Pods cache
+        if: steps.pods-cache-restore.outputs.cache-hit != 'true'
```
matamegger, line 94: For the android setup, we only update the gradle cache on `development`, would this also make sense for the Pods?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,150 @@
...
+      - name: Set up Gradle cache
+        uses: gradle/gradle-build-action@v2
+        with:
+          cache-read-only: ${{ github.ref != 'refs/heads/development' }}
+      - name: Build Android example
+        run: ./gradlew assembleDebug --build-cache
+        working-directory: example/android
+  test-build-ios:
+    name: Build iOS
+    runs-on: macOS-latest
```
matamegger, line 63: For the tvOS build, we used an specific `macOS` version. Is there a reason, why we want to use the latest here, but a specific version for tvOS?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,150 @@
...
+          path: |
+            .cocoapods-cache
+            example/ios/Pods
+          key: ${{ steps.pods-cache-restore.outputs.cache-primary-key }}
+      - name: Build iOS example
+        run: set -o pipefail && xcodebuild -workspace BitmovinPlayerReactNativeExample.xcworkspace -scheme BitmovinPlayerReactNativeExample -configuration Debug build CODE_SIGNING_ALLOWED='NO' | xcpretty
+        working-directory: example/ios
+        env:
+          NSUnbufferedIO: YES
+  test-build-tvos:
```
matamegger, line 106: There is a lot of duplicate setup for iOS and tvOS, and also some shared setup with the android and TS build. 

I think it could be beneficial if we capsule the setup in composite actions, but that can be a future PR. 🙂

matamegger (commented):
Looks good! I only have some questions.

.github/workflows/ci.yml
rolandkakonyi: @matamegger That was a mistake, fixed in 97c3d19

.github/workflows/ci.yml
rolandkakonyi: Agree! I wanted to keep this simple for initial version.

.github/workflows/ci.yml
matamegger: The idea is to have 1 cache, and that it is optimized towards the `development` build, as most of the times PRs do not add or update dependencies. (Not sure if this also holds for this project)

If you have many caches, when are the stale one purged? 🤔

.github/workflows/ci.yml
matamegger: Very fine!

.github/workflows/ci.yml
rolandkakonyi: Eviction policy is documented here: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy

I don't see a problem having up-to-date caches for branches when change happens.

matamegger: 👍

.github/workflows/ci.yml
rolandkakonyi: fixed with d07f7d6 __ALL OCCURRENCES__

matamegger: Thanks a lot!

matamegger (approved):
Very nice! Well done 💪

PR #90, opened 2023-03-17
Replace GH token for update PR creation workflow

matamegger:
Replace the used GH token, so the CI is triggered by the automatic updates.

### DIFF
--- a/.github/workflows/create-sdk-update-pr.yml
+++ b/.github/workflows/create-sdk-update-pr.yml
@@ -19,7 +19,7 @@ jobs:
   update:
     runs-on: ubuntu-latest
     env:
-      GH_TOKEN: ${{ github.token }}
+      GH_TOKEN: ${{ secrets.GH_TOKEN }}
     steps:
       - name: Checkout
         uses: actions/checkout@v3


.github/workflows/create-sdk-update-pr.yml
```diff
@@ -19,7 +19,7 @@ jobs:
   update:
     runs-on: ubuntu-latest
     env:
-      GH_TOKEN: ${{ github.token }}
+      GH_TOKEN: ${{ secrets.GH_TOKEN }}
```
rolandkakonyi, line 22: How does this trigger the CI due to the change?

rolandkakonyi (commented):
There is a typo in the branch name :)

strangesource (approved):

.github/workflows/create-sdk-update-pr.yml
matamegger: The "build in" GH token never triggers other workflows. (To avoid accidental workflow trigger loops)

The new token is a personal access token, and as such will trigger all the workflows a regular user would trigger :)

rolandkakonyi: TIL! Thanks for the context!

PR #98, opened 2023-03-22
Add missing properties to WidevineConfig

strangesource:
## Problem
The `WidevineConfig` for Android does not expose some of its properties yet, namely `httpHeaders` and `shouldKeepDrmSessionsAlive`. These properties are important for integrating with different DRM workflows.

## Changes
- feat: add `httpHeaders` and `shouldKeepDrmSessionsAlive` properties to WidevineConfig
- docs: add usage example for the new properties


src/drm/widevineConfig.ts
```diff
@@ -6,6 +6,11 @@ export interface WidevineConfig {
    * The DRM license acquisition URL.
    */
   licenseUrl: string;
+  /**
+   * A map containing the HTTP request headers, or null.
+   * Android only.
```
matamegger, line 11: The hole `WidevineConfig` is Android only I think. iOS does not support widevine AFAIK 🤔

rolandkakonyi: We actually have it as it can be used for Casting, but casting is not exposed yet.

rolandkakonyi: See: https://github.com/bitmovin-engineering/player-ios/blob/development/BitmovinPlayer/BitmovinPlayer/config/source/drm/WidevineConfig.swift

strangesource: Good point. I will move the comment to the class level in that case.

strangesource: [f9f07ab](https://github.com/bitmovin/bitmovin-player-react-native/pull/98/commits/f9f07ab8e5032594301e5ead3321998047845fef)

hawk23 (approved):

matamegger (approved):

PR #113, opened 2023-04-27
Update AGP to 7.4.2

matamegger:
Update AGP

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,7 +1,7 @@
 buildscript {
     ext {
         kotlinVersion = '1.7.21'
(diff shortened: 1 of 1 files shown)


strangesource (commented):
Do we also have to update gradle-wrapper.properties? I think the gradle version is nolonger compatible otherwise.

matamegger:
> Do we also have to update gradle-wrapper.properties? I think the gradle version is nolonger compatible otherwise.

It is actually not checked in in this repository and thus generated on your machine automatically.

strangesource (approved):

PR #114, opened 2023-04-27
Fix unsafe config property access

matamegger:
If the `isPictureInPictureEnabled` property is not defined in the config, the access fails because of the missing check on the presents of the property.

This PR:
- Introduces an extension function for save boolean property access
- Fixes the bug where `isPictureInPictureEnabled` is not access safely


strangesource (approved):
Nice catch.

PR #135, opened 2023-06-23
Expose timeshift api

zigavehovec:
## Problem
The `timeShift` API isn't exposed yet.

## Changes
- Expose `Player.timeShift()`,  `Player.getTimeShift()` and `Player.getMaxTimeShift()`
- Expose timeShift related events: `TimeShiftEvent` and `TimeShiftedEvent`

Note that `TimeShiftEvent.target` has been renamed to `TimeShiftEvent.targetPosition` in the RN world since `target` cannot be mapped to JS objects with `RCTEventEmitter` (both Android and iOS) for some reason.


android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt
```diff
@@ -106,6 +106,18 @@ class PlayerModule(private val context: ReactApplicationContext) : ReactContextB
         }
     }
 
+    /**
+     * Call `.timShift(offset:)` on `nativeId`'s player.
```
christophaigner, line 110:
```suggestion
     * Call `.timeShift(offset:)` on `nativeId`'s player.
```

ios/RNPlayerView+PlayerListener.swift
```diff
@@ -56,7 +56,12 @@ extension RNPlayerView: PlayerListener {
     func onSeeked(_ event: SeekedEvent, player: Player) {
         onSeeked?(event.toJSON())
     }
-
+    func onTimeShift(_ event: TimeShiftEvent, player: Player) {
+        onTimeShift?(event.toJSON())
+    }
+    func onTimeShifted(_ event: TimeShiftedEvent, player: Player) {
+        onTimeShifted?(event.toJSON())
+    }
```
rolandkakonyi, line 64: please add back empty lines

```suggestion

    func onTimeShift(_ event: TimeShiftEvent, player: Player) {
        onTimeShift?(event.toJSON())
    }

    func onTimeShifted(_ event: TimeShiftedEvent, player: Player) {
        onTimeShifted?(event.toJSON())
    }

```

ios/PlayerModule.swift
```diff
@@ -450,4 +462,39 @@ class PlayerModule: NSObject, RCTBridgeModule {
             resolve(self?.players[nativeId]?.isAd)
         }
     }
+    
+    /**
+     The current time shift of the live stream in seconds. This value is always 0 if the active `source` is not a
+     live stream or there is no active playback session.
```
rolandkakonyi, line 468: We don't have the concept of playback session on iOS.

```suggestion
     The current time shift of the live stream in seconds. This value is always 0 if the active `source` is not a
     live stream or there are no sources loaded.
```

src/player.ts
```diff
@@ -235,6 +235,19 @@ export class Player extends NativeInstance<PlayerConfig> {
...
   };
 
+  /**
+   * Shifts the time to the given `offset` in seconds from the live edge. The resulting offset has to be within the
+   * timeShift window as specified by `maxTimeShift` (which is a negative value) and 0. When the provided `offset` is
+   * positive, it will be interpreted as a UNIX timestamp in seconds and converted to fit into the timeShift window.
+   * When the provided `offset` is negative, but lower than `maxTimeShift`, then it will be clamped to `maxTimeShift`.
+   * Has no effect for VoD.
+   *
+   * Has no effect if there is no active playback session.
```
rolandkakonyi, line 245: There is no playback session concept on iOS so I would adapt this a bit.

```suggestion
   * Has no effect if the player is unloaded.
```

src/player.ts
```diff
@@ -399,4 +412,20 @@ export class Player extends NativeInstance<PlayerConfig> {
   isAd = (): Promise<boolean> => {
     return PlayerModule.isAd(this.nativeId);
   };
+
+  /**
+   * The current time shift of the live stream in seconds. This value is always 0 if the active `source` is not a
+   * live stream or there is no active playback session.
```
rolandkakonyi, line 418:
```suggestion
   * live stream or the player is unloaded.
```

src/player.ts
```diff
@@ -399,4 +412,20 @@ export class Player extends NativeInstance<PlayerConfig> {
...
+   */
+  getTimeShift = (): Promise<boolean> => {
+    return PlayerModule.getTimeShift(this.nativeId);
+  };
+
+  /**
+   * The limit in seconds for time shifting. This value is either negative or 0 and it is always 0 if the active
+   * `source` is not a live stream or there is no active playback session.
```
rolandkakonyi, line 426:
```suggestion
   * `source` is not a live stream or the player is unloaded.
```

rolandkakonyi (changes requested):
I marked some changes with 🚨 as those are functionally broken.

Others are documentation related only.

Note: We missed exposing `LiveConfig` but I think that could be covered in a separate PR later.

src/player.ts
```diff
@@ -399,4 +412,20 @@ export class Player extends NativeInstance<PlayerConfig> {
   isAd = (): Promise<boolean> => {
     return PlayerModule.isAd(this.nativeId);
   };
+
+  /**
+   * The current time shift of the live stream in seconds. This value is always 0 if the active `source` is not a
+   * live stream or there is no active playback session.
+   */
+  getTimeShift = (): Promise<boolean> => {
```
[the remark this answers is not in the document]
zigavehovec: I have no idea how I missed this 🙈. Regarding async, should `isAd` then also be async?

src/player.ts
zigavehovec: What about: 

```suggestion
   * Has no effect if no sources are loaded.
```

src/player.ts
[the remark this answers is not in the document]
rolandkakonyi: @zigavehovec Yes

src/player.ts
zigavehovec: I feel like it should be understandable. I'm going with the "Has no effect if no sources are loaded" version.

src/player.ts
zigavehovec:
```suggestion
   * live stream or no sources are loaded.
```

src/player.ts
zigavehovec:
```suggestion
   * `source` is not a live stream or no sources are loaded.
```

rolandkakonyi (approved):

PR #141, opened 2023-06-28
Don't force react native version

zigavehovec:
## Problem
Our project specified a `react-native` version in dependencies, making it not compatible for apps based on other react-native versions. 
Instead, a react-native version should be specified in devDependencies.

## Changes
- Remove `react-native` and `react` form dependencies
- Add `react-native` to devDependencies


rolandkakonyi (approved):
I tested it and it works nicely!

Good job!

Here is a sample project that I used for testing:
[react-native-sandbox.zip](https://github.com/bitmovin/bitmovin-player-react-native/files/11893011/react-native-sandbox.zip)

PR #142, opened 2023-06-28
fix: webUi settings items don't work when clicked

zigavehovec:


android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
```diff
@@ -198,7 +199,13 @@ class RNPlayerViewManager(private val context: ReactApplicationContext) : Simple
             if (view.playerView != null) {
                 view.player = player
             } else {
-                val playerView = PlayerView(context, player)
+                // PlayerView has to be initialized with Activity context
+                val currentActivity = context.currentActivity
+                if (currentActivity == null) {
+                    Log.e(MODULE_NAME, "Cannot create a PlayerView, because no activity is attached.")
```
zigavehovec, line 205: I was thinking of throwing here, but because of the bridge event sending arch I suppose it could be possible to not be attached to an activity anymore, so I decided to log an error instead.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
```diff
@@ -198,7 +199,13 @@ class RNPlayerViewManager(private val context: ReactApplicationContext) : Simple
...
+                    return@post
+                }
+                val playerView = PlayerView(currentActivity, player)
```
krocard, line 208: This issue doesn't seem React native specific; could `PlayerView` constructor throw (or at the very least log an error) when it receive an application context instead of an activity one?

krocard (approved):

krocard (approved):
good catch!

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
zigavehovec: Let's discuss outside this PR 😊

rolandkakonyi (approved):
Very nice catch!

Issue #150, opened 2023-07-03
Problem with Android Subtitule custom label


jonathanm-tkf:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Hello everyone I wanted to report a possible bug in the Android integration, this does not happen in iOS I attach evidence of both platforms.

iOS

![Simulator Screen Shot - iPhone 11 Pro Max - 2023-07-03 at 16 57 05](https://github.com/bitmovin/bitmovin-player-react-native/assets/67608970/671a40c5-00a1-4190-882f-5c10674ad787)

Android

![image](https://github.com/bitmovin/bitmovin-player-react-native/assets/67608970/4b41ce5f-5142-4eee-b61e-945ddb3b4c23)

### Expected Behavior

Android

![image](https://github.com/bitmovin/bitmovin-player-react-native/assets/67608970/b5dce045-b33c-46d1-844a-5352ffc42b7d)

### Steps To Reproduce

1- Install the example
2- Run Subtitles and Captions screen
3- Press on options and try to change subtitles

### What platform(s) are you experiencing the issue on?

- [X] Android
- [ ] Android TV / Fire TV
- [ ] iOS
- [ ] tvOS

### Player React Native SDK version

0.7.2

jonathanm-tkf:
Hello guys 

I create this [PR](#151 ) and tried to solve the issue, can you check if the implementation it's ok?

jonathanm-tkf:
The only thing I couldn't solve was the CammelCase, probably something in the skin that is modified by custom CSS.

custom english => Custom English

zigavehovec:
Hi @jonathanm-tkf! 

Thanks for reporting this issue!
I've looked into this and the issue actually originates in the native SDK side and not in the React Native wrapper.
I've created an internal ticket and we will look into it.

We'll post an update when the issue is fixed here.

jonathanm-tkf:
Hi @zigavehovec and team 

can you confirm if this issue has any update from the native side?

I was checking the release list but I don't see anything regarding this issue.

https://developer.bitmovin.com/playback/docs/release-notes-android

Thanks and regards

christophaigner:
Hi @jonathanm-tkf,

Thanks again for reporting.
You're right, there is no update on this yet in our release notes, as this is scheduled for later in this quarter.

We will make sure to notify you here.

Chris

AngeloFragliassoBritBox:
[...]

PR #151, opened 2023-07-03
Problem with Android Label issue 150

jonathanm-tkf:


zigavehovec:
Hi! 
Thanks for your submission 😊
As mentioned [here](https://github.com/bitmovin/bitmovin-player-react-native/issues/150), the issue originates in the native SDK and not the React Native wrapper, so this is the wrong layer to do any changes.

We'll update you in the attached issue as soon as the state changes.

Issue #152, opened 2023-07-05
Implemented SourceOptions support


jonathanm-tkf:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Hi team, I am adding the startOffset functionality which is already in the SDK natively but not integrated into the bridge.

![image](https://github.com/bitmovin/bitmovin-player-react-native/assets/67608970/2d01d5a2-6922-4789-bd1e-951ba5065a01)

### Expected Behavior

Being able to initialize the player at a specific time without the need to seek

### What platform(s) are you experiencing the issue on?

- [X] Android
- [ ] Android TV / Fire TV
- [X] iOS
- [ ] tvOS

### Player React Native SDK version

0.7.2

jonathanm-tkf:
[PR](https://github.com/bitmovin/bitmovin-player-react-native/pull/153) associated, If the team had any duds let me know, thanks

PR #153, opened 2023-07-05
Add support SourceOptions

jonathanm-tkf:


jonathanm-tkf:
Hi @rolandkakonyi how are you? thank you very much for the help, if I'm honest I saw the property but as I'm not familiar with the SDK I'm not sure how to implement it, if you could give me some guidance on this aspect I would appreciate it. I know it is an effort for both parties but it is easier if you guide me and we solve it, thank you very much.

rolandkakonyi:
Hi @jonathanm-tkf!

You could take a look at how `SourceType` is bridged between native Swift and Kotlin code and between React Native Typescript and use the same approach for type `TimelineReferencePoint` and property `startOffsetTimelineReference`.

Check these files as reference:
- Typescript type definition: https://github.com/bitmovin/bitmovin-player-react-native/blob/6eb6b049b480a40db5e69840c2e07cf54349e877/src/source.ts#L11
- Android conversion: https://github.com/bitmovin/bitmovin-player-react-native/blob/6eb6b049b480a40db5e69840c2e07cf54349e877/android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt#L283
- iOS conversion: https://github.com/bitmovin/bitmovin-player-react-native/blob/6eb6b049b480a40db5e69840c2e07cf54349e877/ios/RCTConvert%2BBitmovinPlayer.swift#L263
-

rolandkakonyi:
Hi @jonathanm-tkf! Do you need any help with the suggestions from my last message?

jonathanm-tkf:
Hello @rolandkakonyi, unfortunately, I have not had time to implement it, due to other priorities that I have, as soon as I am free I add it, if you want to implement it go ahead, thank you very much.

rolandkakonyi:
@jonathanm-tkf Now thinking about this PR it should be already good to merge!
We can add the missing property later.

rolandkakonyi (approved):

src/source.ts
```diff
@@ -45,6 +45,16 @@ export enum LoadingState {
...
 
+/**
+ * Types of SourceOptions.
+ */
+export interface SourceOptions {
+  /**
+   * Indicates the player's offset time in seconds.
+   */
+  startOffset?: number;
+}
```
matamegger, line 56: The native side also defines a reference edge (start or end of the stream) this offset is applied to. 
Additionally, per default the reference edge for VoD stream is the start, while the reference edge for live streams is the end of the stream.
Further, having a negative number for the start of the stream does not make sense of course, because we can not start before 0.. similar with the end edge of the stream, where we can not start in the future of the live stream.

While I do not see a particular need to introduce the refrence edge config, I would ask you to update the documentation of this property a bit to reflect the behaviour a bit.
I am happy to answer further questions 🙂

matamegger (changes requested):
Thanks for contributing! 
Your changes look good already. Please just have a look at the documentation of the property you introduced. I think it needs a bit more context and explanation on the behaviour, so that other users also know how it behaves and how they can use it.

src/source.ts
matamegger: VoD:  start(currentTime=0) |--------------------| end (currentTime = duration)
Live: start(timeShift=maxTimeshift) |--------------------> end/live edge (timeShift=0)

rolandkakonyi:
@jonathanm-tkf I took your changes and added the missing property and implemented our feedback in #211

jonathanm-tkf:
Hi @rolandkakonyi unfortunately I have not had time to add this functionality could you reopen the ticket? because it is something necessary that we have nowadays.

rolandkakonyi:
@jonathanm-tkf what do you mean?
Your changes will be merged as part of #211 and released in the next release.

The issue you opened is still open, see: #152
That one will be closed once my PR is merged.

jonathanm-tkf:
Sorry I had not understood because it had been closed previously, apologies for the confusion excellent work thank you very much.

PR #169, opened 2023-07-26
Fullscreen prop for the PlayerView

matamegger:
For certain use cases it might be needed to set the fullscreen state from code.

This PR introduces the `isFullscreen` prop on the `PlayerView`, which triggers the according mechanism on the native side to enter/exit fullscreen. 
The prop also relies on the `FullscreenHandler`, so one must specify a `FullscreenHandler` in order to use the new prop.

### DIFF
--- a/android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
@@ -133,9 +133,9 @@ class RNPlayerViewManager(private val context: ReactApplicationContext) : Simple
      * to call 'functions' on them.
      * @return map between names (used in js) and command ids (used in native code).
      */
-    override fun getCommandsMap(): MutableMap<String, Int> = Commands.values().associate {
+    override fun getCommandsMap(): Map<String, Int> = Commands.values().associate {
         it.command to it.ordinal
-    }.toMutableMap()
+    }
 
     /**
      * Callback triggered in response to command dispatches from the js side.
@@ -144,7 +144,9 @@ class RNPlayerViewManager(private val context: ReactApplicationContext) : Simple
      * @param args Arguments list sent from the js side.
      */
     override fun receiveCommand(view: RNPlayerView, commandId: String?, args: ReadableArray?) {
-        val command = commandId?.toInt()?.toCommand() ?: return
+        val command = commandId?.toInt()?.toCommand() ?: throw IllegalArgumentException(
+            "The received command is not supported by the Bitmovin Player View"
+        )
         when (command) {
             Commands.ATTACH_PLAYER -> attachPlayer(view, args?.getString(1), args?.getMap(2))
             Commands.ATTACH_FULLSCREEN_BRIDGE -> args?.getString(1)?.let { fullscreenBridgeId ->


example/src/screens/BasicFullscreenHandling.tsx
```diff
@@ -103,6 +103,7 @@ export default function BasicFullscreenHandling({
     <View style={styles.container}>
       <PlayerView
         player={player}
+        isFullscreen={false}
```
rolandkakonyi, line 106: nit: Do we need it here in the sample screen? With a hard-coded value I'm not sure it adds value here.

We could explain it the full screen feature guide.

ios/RNPlayerViewManager.m
```diff
@@ -56,5 +56,6 @@ @interface RCT_EXTERN_REMAP_MODULE(NativePlayerView, RNPlayerViewManager, RCTVie
 RCT_EXTERN_METHOD(attachPlayer:(nonnull NSNumber *)viewId playerId:(NSString *)playerId playerConfig:(nullable NSDictionary *)playerConfig)
 RCT_EXTERN_METHOD(attachFullscreenBridge:(nonnull NSNumber *)viewId fullscreenBridgeId:(NSString *)fullscreenBridgeId)
 RCT_EXTERN_METHOD(setCustomMessageHandlerBridgeId:(nonnull NSNumber *)viewId customMessageHandlerBridgeId:(NSString *)fullscreenBridgeId)
+ RCT_EXTERN_METHOD(setFullscreen:(nonnull NSNumber *)viewId isFullscreen:(BOOL)isFullscreen)
```
rolandkakonyi, line 59: Please remove extra space at the beginning of the line
```suggestion
RCT_EXTERN_METHOD(setFullscreen:(nonnull NSNumber *)viewId isFullscreen:(BOOL)isFullscreen)
```

ios/RNPlayerViewManager.swift
```diff
@@ -71,6 +71,28 @@ class RNPlayerViewManager: RCTViewManager {
...
+                 let view = views?[viewId] as? RNPlayerView
+             else {
+                 return
+             }
+             guard let playerView = view.playerView else {
+                 return
+             }
+             if playerView.isFullscreen == isFullscreen {
+                 return
+             }
```
rolandkakonyi, line 87: `guard` would be the idiomatic way in Swift to return early in a situation like this.

```suggestion
             guard playerView.isFullscreen != isFullscreen else {
                 return
             }
```

src/components/PlayerView/index.tsx
```diff
@@ -35,9 +35,23 @@ export interface PlayerViewProps extends BasePlayerViewProps, PlayerViewEvents {
...
+  /**
+   * The `CustomMessageHandler` that can be used to directly communicate with the embedded WebUi.
+   */
   customMessageHandler?: CustomMessageHandler;
+
+  /**
+   * Can be set to `true` to enter fullscreen mode, or `false` to exit fullscreen mode.
+   * By using this property to change the fullscreen state, we make sure that the embedded Player UI is also aware
+   * of potential fullscreen state changes.
+   * To use this property, a `FullscreenHandler` must be set.
```
rolandkakonyi, line 52: Should we mention that for state changes they should listen to `onFullscreenEnter` and `onFullscreenExit` events?

matamegger: I would say no. The fullscreen state is a state of the screen the `PlayerView` is part of. So the screen should know the state independent of the `PlayerView`.

rolandkakonyi: @matamegger What I had in mind is that this value will get out of sync once you hit the fullscreen button on the player UI.

matamegger: https://github.com/bitmovin/bitmovin-player-react-native/pull/169/commits/05b9afa1b853cc838ae2b4801dc9cb0f9929b4ce

Rethinking this, having an explicit note is probably best.

rolandkakonyi (approved):

src/components/PlayerView/index.tsx
```diff
@@ -35,9 +35,25 @@ export interface PlayerViewProps extends BasePlayerViewProps, PlayerViewEvents {
...
 
+  /**
+   * The `CustomMessageHandler` that can be used to directly communicate with the embedded WebUi.
+   */
   customMessageHandler?: CustomMessageHandler;
+
+  /**
+   * Can be set to `true` to enter fullscreen mode, or `false` to exit fullscreen mode.
+   * Should not be used to get the current fullscreen state. Use `onFullscreenEnter` and `onFullscreenExit`
+   * or the `FullscreenHandler.isFullscreenActive` property to get the current state.
```
krocard, line 51: Isn't this surprising and error prone? How about making it a method call (eg `requestFullscreen(Boolean)`) (no getter) or renaming it `isFullscreenRequested`.

krocard (approved):

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
```diff
@@ -145,19 +144,21 @@ class RNPlayerViewManager(private val context: ReactApplicationContext) : Simple
      * @param args Arguments list sent from the js side.
      */
     override fun receiveCommand(view: RNPlayerView, commandId: String?, args: ReadableArray?) {
-        super.receiveCommand(view, commandId, args)
-        commandId?.toInt()?.let {
-            when (it) {
-                Commands.ATTACH_PLAYER.ordinal -> attachPlayer(view, args?.getString(1), args?.getMap(2))
-                Commands.ATTACH_FULLSCREEN_BRIDGE.ordinal -> args?.getString(1)?.let { fullscreenBridgeId ->
-                    attachFullscreenBridge(view, fullscreenBridgeId)
+        val command = commandId?.toInt()?.toCommand() ?: return
```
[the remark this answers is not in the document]
krocard: This is not to be acted on in this PR. It is only for context and awareness.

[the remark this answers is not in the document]
matamegger: I don't think we should use `!!`. I will add an explicit exception.

src/components/PlayerView/index.tsx
matamegger: We were advised to go for properties instead of methods. Having a method is also not directly possible in this setup. We were discussing that a bit and this is what we settled on.
Yea, naming this a bit different sounds like a good idea.

matamegger: https://github.com/bitmovin/bitmovin-player-react-native/pull/169/commits/14271faf65c61ef661030b787a8b35fc7dc31538

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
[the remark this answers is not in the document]
matamegger: https://github.com/bitmovin/bitmovin-player-react-native/pull/169/commits/9bbb0277e6ea52cb1365e228745697ad7cc66382

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
```diff
@@ -132,11 +133,9 @@ class RNPlayerViewManager(private val context: ReactApplicationContext) : Simple
...
      * @return map between names (used in js) and command ids (used in native code).
      */
-    override fun getCommandsMap(): MutableMap<String, Int> = mutableMapOf(
-        "attachPlayer" to Commands.ATTACH_PLAYER.ordinal,
-        "attachFullscreenBridge" to Commands.ATTACH_FULLSCREEN_BRIDGE.ordinal,
-        "setCustomMessageHandlerBridgeId" to Commands.SET_CUSTOM_MESSAGE_HANDLER_BRIDGE_ID.ordinal,
-    )
+    override fun getCommandsMap(): MutableMap<String, Int> = Commands.values().associate {
+        it.command to it.ordinal
+    }.toMutableMap()
```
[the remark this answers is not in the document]
matamegger: https://github.com/bitmovin/bitmovin-player-react-native/pull/169/commits/9bbb0277e6ea52cb1365e228745697ad7cc66382

krocard (approved):

Issue #163, opened 2023-07-17
Black screen playback


jonathanm-tkf:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Good morning, we are having a problem with the player, on certain occasions, this goes black screen, and we can not see the video, apparently, this happens in Android 10, 9, 8, .... in versions like 11, 12, and 13 this does not occur. Do you have any idea what could be happening? Any comments on how to fix it would be appreciated.

https://github.com/bitmovin/bitmovin-player-react-native/assets/67608970/504d6229-126c-48af-a727-c72d9b08e827

I have been looking at the native code and I see that they have a reference, I attach the code extract in case it has something to do with the problem, thank you very much.

### Expected Behavior

not see black screen and reproduce video normally

### Steps To Reproduce

1 - Run basic implementation
2 - Confirm you are on Android 10, 9, 8 .... or lower

### What platform(s) are you experiencing the issue on?

- [X] Android
- [ ] Android TV / Fire TV
- [ ] iOS
- [ ] tvOS

### Player React Native SDK version

0.7.2

### Device / Environment

- Device:
- Android 9

### Additional information / Code snippets / Screenshots

RNPlayerView.kt
```java
**
     * Try to measure and update this view layout as much as possible to
     * avoid layout problems related to React or old layout values present
     * in `playerView` due to being previously attached to a different parent.
     */
    override fun requestLayout() {
        super.requestLayout()
        post {
            measure(
                MeasureSpec.makeMeasureSpec(width, MeasureSpec.EXACTLY),
                MeasureSpec.makeMeasureSpec(height, MeasureSpec.EXACTLY)
            )
            layout(left, top, right, bottom)
        }
    }
```

christophaigner:
Hi @jonathanm-tkf,

Thank you for reporting this, we appreciate it.
We were able to reproduce this issue occasionally as well and will look into it.

Chris

jonathanm-tkf:
Hi @christophaigner if I can help in any way, please do not hesitate to let us know, unfortunately, it is something that is detrimental to our production output, we look forward to your support and thank you very much.

christophaigner:
Hi @jonathanm-tkf,

Thank you for your patience. We've just released [v0.8.0](https://github.com/bitmovin/bitmovin-player-react-native/releases/tag/v0.8.0), which contains a fix/workaround for this problem (see #174 for more details).

Let us know if you still see the issue with the latest version.

Chris

[...]

PR #174, opened 2023-07-27
fix: workaround React native command not sent

krocard:
React Native's `UIManager.dispatchViewManagerCommand` sometimes (~3/4 times on emulator, very rarely on recent devices) doesn't sent native commands. They get stuck until the next UI refresh.

This issue should be fixed in the latest React Native release but our SDK needs significant work to update to this version.

Until the React Native version is updated, force UI redraws after dispatching commands to force them to be received by the native side.

Player issue: https://github.com/bitmovin/bitmovin-player-react-native/issues/163
RN issue: https://github.com/microsoft/react-native-windows/issues/7543

### DIFF
--- a/src/components/PlayerView/index.tsx
+++ b/src/components/PlayerView/index.tsx
@@ -57,7 +57,6 @@ function dispatch(command: string, node: NodeHandle, ...args: any[]) {
     Platform.OS === 'android'
       ? (UIManager as any).NativePlayerView.Commands[command].toString()
       : UIManager.getViewManagerConfig('NativePlayerView').Commands[command];
-  //console.log("dispatching command %s", command)
   UIManager.dispatchViewManagerCommand(
     node,
     commandId,
@@ -76,7 +75,6 @@ export function PlayerView({
   customMessageHandler,
   ...props
 }: PlayerViewProps) {
-  // Native view reference.
   const fakeStateUpdater = useState(1)[1];
   const workaroundViewManagerCommandNotSent = useCallback(() => {
     setTimeout(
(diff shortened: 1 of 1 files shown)


src/components/PlayerView/index.tsx
```diff
@@ -57,6 +57,7 @@ function dispatch(command: string, node: NodeHandle, ...args: any[]) {
     Platform.OS === 'android'
       ? (UIManager as any).NativePlayerView.Commands[command].toString()
       : UIManager.getViewManagerConfig('NativePlayerView').Commands[command];
+  //console.log("dispatching command %s", command)
```
matamegger, line 60:
```suggestion
```

I guess we can remove this.

src/components/PlayerView/index.tsx
```diff
@@ -76,6 +77,17 @@ export function PlayerView({
   ...props
 }: PlayerViewProps) {
   // Native view reference.
+  const fakeStateUpdater = useState(1)[1];
```
matamegger, line 80: The comment above is no longer at the right position.. that said, I think we can remove it anyways.

src/components/PlayerView/index.tsx
```diff
@@ -76,6 +77,17 @@ export function PlayerView({
   ...props
 }: PlayerViewProps) {
   // Native view reference.
```
krocard, line 79:
```suggestion
```

src/components/PlayerView/index.tsx
krocard: Done: d87a4452dc7eaccf426b4d696fb3f28889b19b29

matamegger (approved):

rolandkakonyi (approved):
I just jumped in uninvited to test this on iOS :) 
It still works, so this is good to go from my side with the comment by Matthias addressed.

PR #181, opened 2023-08-08
Update android player to 3.41.0

bitPlayerGHActions:
Automated android player version update to 3.41.0


matamegger:
The version of the android player contains the bitmovin analytics directly. Before merging the upgrade, we also have to align the analytics integration.

Issue #176, opened 2023-08-02
We need a couple of values to be added to the analytics helper


jonathanm-tkf:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Currently, we have the following methods from the documentation

https://developer.bitmovin.com/playback/docs/enabling-bitmovin-analytics

```js
// Update the current custom data config.
player.analyticsCollector?.setCustomDataOnce({
  customData2: 'Updated custom data field 2',
  customData4: 'Updated custom data field 4',
});
```

we need to add new methods such as 

### Expected Behavior

```js
player.analyticsCollector.setVideoId('123');
player.analyticsCollector.setTitle('Title of the Video');
player.analyticsCollector.setCdnProvider('Cdn Provider');
```

### What platform(s) are you experiencing the issue on?

- [X] Android
- [ ] Android TV / Fire TV
- [X] iOS
- [ ] tvOS

### Player React Native SDK version

0.8.0

rolandkakonyi:
Hi @jonathanm-tkf!

`AnalyticsConfig` already offers options to assign those values on player creation.
Could this be used for your use case?

jonathanm-tkf:
Hi @rolandkakonyi unfortunately it doesn't work for me, as we have the Next Episode functionality, which we don't destroy the player.

I'll comment in more detail:

Currently, we have the player, which has a Next Episode button, once pressed, loads the next video, for performance reasons and to not delete the current player and recreate it we use the load implementation as follows:

```js
 player.load({
        url: videoData.url,
        type: Platform.OS === 'ios' ? SourceType.HLS : SourceType.DASH,
        title: videoData.title,
        poster: videoData.posterUrl,
...
...
..
```

after this we need to do the following 

```js
      if (player.analyticsCollector) {
        player.analyticsCollector.setCustomDataOnce({
          customData2: videoData.customId,
          customData3: videoData.supplier,
        });
        player.analyticsCollector.setVideoId(videoData.id);
        player.analyticsCollector.setTitle(getTitleAnalytics(videoData.dataAnalytics));
        player.analyticsCollector.setCdnProvider(videoData.supplier);
      }
```

Do you understand what the idea is? any recommendations? or do we need to destroy the player every time we want to modify this data?

Thanks for your help

rolandkakonyi:
Hi @jonathanm-tkf!

I see, you are correct that this is not possible right now without creating a new player instance.

[...]

PR #184, opened 2023-08-09
Add missing APIs for analytics collector

rolandkakonyi:
This PR adds missing API for the analytics integration:
- `CustomData.experimentName` property
- `Player.analyticsCollector.addSourceMetadata` API and related `SourceMetadata` type

### DIFF
--- a/android/src/main/java/com/bitmovin/player/reactnative/AnalyticsModule.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/AnalyticsModule.kt
@@ -129,7 +129,7 @@ class AnalyticsModule(private val context: ReactApplicationContext) : ReactConte
     @ReactMethod
     fun addSourceMetadata(nativeId: NativeId, playerId: NativeId?, json: ReadableMap?) {
         uiManager()?.addUIBlock { _ ->
-            val playerSource = playerModule()?.getPlayer(playerId)?.source ?: return
+            val playerSource = playerModule()?.getPlayer(playerId)?.source ?: return@addUIBlock
             JsonConverter.toAnalyticsSourceMetadata(json)?.let { sourceMetadata ->
                 collectors[nativeId]?.addSourceMetadata(playerSource, sourceMetadata)
             }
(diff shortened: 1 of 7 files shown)


matamegger:
@rolandkakonyi with the recently introduced Player/Analytics bundling I guess we will change the implementation of some of the introduced functions in a next step. right?

rolandkakonyi:
> @rolandkakonyi with the recently introduced Player/Analytics bundling I guess we will change the implementation of some of the introduced functions in a next step. right?

@matamegger yes, that is the plan, but that will require bigger changes around the analytics integration in my opinion.
This is to provide the missed APIs to help the resolution of #176.

android/src/main/java/com/bitmovin/player/reactnative/AnalyticsModule.kt
```diff
@@ -126,6 +126,17 @@ class AnalyticsModule(private val context: ReactApplicationContext) : ReactConte
...
     }
 
+    @ReactMethod
+    fun addSourceMetadata(nativeId: NativeId, playerId: NativeId?, json: ReadableMap?) {
+        uiManager()?.addUIBlock { _ ->
+            playerModule()?.getPlayer(playerId)?.source?.let { playerSource ->
+                JsonConverter.toAnalyticsSourceMetadata(json)?.let { sourceMetadata ->
+                    collectors[nativeId]?.addSourceMetadata(playerSource, sourceMetadata)
+                }
+            }
```
matamegger, line 136: Let's make this a bit easier to digest by removing one layer please.

```suggestion
            val playerSource = playerModule()?.getPlayer(playerId)?.source ?: return
            JsonConverter.toAnalyticsSourceMetadata(json)?.let { sourceMetadata ->
                collectors[nativeId]?.addSourceMetadata(playerSource, sourceMetadata)
            }
```

matamegger (approved):
good addition. Seems to be inline with the existing code.

Please consider my suggestion.

android/src/main/java/com/bitmovin/player/reactnative/AnalyticsModule.kt
rolandkakonyi: Thanks for the suggestion!
Taken!

Issue #180, opened 2023-08-03
Problem with player, black screen with controls but video don't present


jonathanm-tkf:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Hello, we have a problem with the player because it loses the video on the screen and only shows the controls, the audio is heard but the image is not represented on the device.

I've created a repository with a branch called "error", and I'll add a video demonstrating the problem.

Apparently, there is a problem with the player instance because the image is not displayed but the audio persists in the app, we have also noticed that the default skin of Bitmovin-player-ui has a double-tap functionality on the screen, which allows you to exit fullscreen mode making the video resize and appear on the screen, it is also in the attached video.

https://github.com/bitmovin/bitmovin-player-react-native/assets/67608970/0d391ffc-abb1-4e82-88b1-e7e1a745af7c

This double tap functionality unfortunately we don't have in our integration which makes it impossible for the user to do any action, The only thing you can do is via the back button and then try again, this is quite frustrating to be honest.

Points to consider

- The playback starts with audio and video, after a few seconds or minutes the playback continues with no video, only audio and the player goes black.

- There is no crash, the UI is still fully interactive, but video is completely gone.

- Scrubbing to an already watched section of the playback doesn't solve the issue, video will not recover in any case.

We know the issue is more prevalent in bad connections and it's device specific and on Android 10 or lower. No prevalence in iOS at all.

In the last 4 months we've had at least 30 clients complaining of the issue explicitly in the play store, a few examples attached:

![image](https://github.com/bitmovin/bitmovin-player-react-native/assets/67608970/e594e4fd-36ec-453e-8435-1f3a48194744)

I also attach a video of when this behaviour works correctly, in some cases and we don't know what the reason is. honestly, it's like flipping a coin and having the probability that it will work or not.

https://github.com/bitmovin/bitmovin-player-react-native/assets/67608970/17f046a4-982c-4bcc-adf5-04f250e974b7

### Expected Behavior

As we see the idea and the expectation is that the user can make use of the player by placing a next episode button, the only thing it does is to load a video, for that we use the ```player.load``` functionality which is provided in the integration. 

### Steps To Reproduce

1 - Download the repository
2 - Run the app locally
3- Clic on Landscape Fullscreen Britbox option
4- We wait for the video to load, we do seek if we prefer it.
5- Click on Next Episode, this will display a react native modal.
[...]

PR #186, opened 2023-08-11
Player sometimes plays sound but no video

zigavehovec:
## Problem (PRN-26)

_Sometimes video suddenly doesn't play anymore, but audio continues to play._
Issue: https://github.com/bitmovin/bitmovin-player-react-native/issues/180

Turns out that this was influenced by a React Native issue, reported here: https://github.com/facebook/react-native/issues/17968

In our case the native `PlayerView` adds and removes `SurfaceView` under the hood. After screen rotation newly added SurfaceView's size wasn't calculated properly, which caused no video to be displayed.

## Changes
- Add a [workaround](https://github.com/facebook/react-native/issues/17968#issuecomment-549780002) that triggers a new layout measurement

### DIFF
--- a/android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
@@ -82,6 +82,16 @@ private val EVENT_CLASS_TO_REACT_NATIVE_NAME_MAPPING_UI = mapOf<KClass<out Event
 @SuppressLint("ViewConstructor")
 class RNPlayerView(val context: ReactApplicationContext) : LinearLayout(context),
     LifecycleEventListener, View.OnLayoutChangeListener, RNPictureInPictureDelegate {
+
+    init {
+        // React Native has a bug that dynamically added views sometimes aren't laid out again properly.
+        // Since we dynamically add and remove SurfaceView under the hood this caused the player
+        // to suddenly not show the video anymore because SurfaceView was not laid out properly.
+        // Bitmovin player issue: https://github.com/bitmovin/bitmovin-player-react-native/issues/180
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -82,6 +82,16 @@ private val EVENT_CLASS_TO_REACT_NATIVE_NAME_MAPPING_UI = mapOf<KClass<out Event
...
 class RNPlayerView(val context: ReactApplicationContext) : LinearLayout(context),
     LifecycleEventListener, View.OnLayoutChangeListener, RNPictureInPictureDelegate {
+
+    init {
+        // React Native has a bug that dynamically added views sometimes aren't laid out again properly.
+        // Since we dynamically add and remove SurfaceView under the hood this caused the player
+        // to suddenly not show the video anymore because SurfaceView was not laid out properly.
+        // Bitmovin player issue: https://github.com/bitmovin/bitmovin-player-react-native/issues/180
+        // React Native layout issue: https://github.com/facebook/react-native/issues/17968
+        getViewTreeObserver().addOnGlobalLayoutListener { requestLayout() }
```
rolandkakonyi, line 92: Do we need to also remove this listener when the player view gets destroyed?

rolandkakonyi:
Tested this and it works!

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
zigavehovec: Not needed, as the listener will be garbage collected when the view is destroyed. This would only be problematic if the listener held on to a component that is longer lived then the view itself 🙂

rolandkakonyi (approved):

matamegger (approved):

PR #212, opened 2023-08-30
Add support for `AdaptationConfig`

rolandkakonyi:
## Description

This PR adds support for configuring adaptation for the player via `PlayerConfig.adaptationConfig`.

## Changes

Introduced `AdaptationConfig` type and `PlayerConfig.adaptationConfig` property.
For now only the `maxSelectableBitrate` property is supported.

To test this in the Example app, replace:
```ts
const player = usePlayer();
```
with
```ts
const player = usePlayer({
  adaptationConfig: {
    maxSelectableBitrate: 700000,
  },
});
```

and see that the content is limited to lower resolution.

## Checklist
- [x] 🗒 `CHANGELOG` entry

### DIFF
--- a/src/adaptationConfig.ts
+++ b/src/adaptationConfig.ts
@@ -1,0 +1,10 @@
+/**
+ * Configures the adaptation logic.
+ */
+export interface AdaptationConfig {
+  /**
+   * The upper bitrate boundary in bits per second for network bandwidth consumption of the played source.
+   * Can be set to `undefined` for no limitation.
+   */
+  maxSelectableBitrate?: number;
+}
--- a/src/adaptationConfig.ts
+++ b/src/adaptationConfig.ts
@@ -3,7 +3,7 @@
  */
 export interface AdaptationConfig {
   /**
-   * The upper bitrate boundary in bits per second for network bandwidth consumption of the played source.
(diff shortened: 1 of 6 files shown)


src/adaptationConfig.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Configures the adaptation logic.
+ */
+export interface AdaptationConfig {
+  /**
+   * The upper bitrate boundary in bits per second for approximate network bandwidth consumption of the played source.
```
rolandkakonyi, line 6: We had a brief discussion about with @matamegger that resulted in this API docs.
We found this describes the idea behind of the flag and the behavior the best on both platforms.

Context:
- on Android: this affects only video rendition bitrate selection
- on iOS: this affects all rendition bitrate selection

zigavehovec: Thanks for the context. I think we could even add this to the docs. What do you think about that?

zigavehovec: Additionally (to avoid confusion) we could also mention that (at least in Android) it's still possible to over this limit manually.

zigavehovec (approved):
Looks good to me! 👍

PR #214, opened 2023-09-01
Add missing `setMaxSelectableBitrate` API to `Player`

rolandkakonyi:
## Description

This PR compliments #212 by adding `setMaxSelectableBitrate` API to the `Player`.

## Changes

Added `setMaxSelectableBitrate` to `Player`, with no limit value handling when `undefined` is passed.

## Checklist
- [x] 🗒 `CHANGELOG` entry

### DIFF
--- a/src/player.ts
+++ b/src/player.ts
@@ -448,4 +448,14 @@ export class Player extends NativeInstance<PlayerConfig> {
   getMaxTimeShift = async (): Promise<number> => {
     return PlayerModule.getMaxTimeShift(this.nativeId);
   };
+
+  /**
+   * Sets the upper bitrate boundary for video qualities. All qualities with a bitrate
+   * that is higher than this threshold will not be eligible for automatic quality selection.
+   *
+   * Can be set to `undefined` for no limitation.
+   */
+  setMaxSelectableBitrate = (bitrate: number | undefined) => {
+    PlayerModule.setMaxSelectableBitrate(this.nativeId, bitrate || -1);
+  };
 }
--- a/src/player.ts
+++ b/src/player.ts
@@ -455,7 +455,7 @@ export class Player extends NativeInstance<PlayerConfig> {
    *
    * Can be set to `undefined` for no limitation.
    */
-  setMaxSelectableBitrate = (bitrate: number | undefined) => {
+  setMaxSelectableBitrate = (bitrate: number | null) => {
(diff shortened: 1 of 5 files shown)


src/player.ts
```diff
@@ -448,4 +448,14 @@ export class Player extends NativeInstance<PlayerConfig> {
...
   };
+
+  /**
+   * Sets the upper bitrate boundary for video qualities. All qualities with a bitrate
+   * that is higher than this threshold will not be eligible for automatic quality selection.
+   *
+   * Can be set to `undefined` for no limitation.
+   */
+  setMaxSelectableBitrate = (bitrate: number | undefined) => {
+    PlayerModule.setMaxSelectableBitrate(this.nativeId, bitrate || -1);
```
zigavehovec, line 459: Can we pass null to the PlayerModule? Then we wouldn't need to do any special handling here and in the native parts we would just work with nulls (which would also be a lot cleaner IMO).

I see that Android's native part is already nullable, but the iOS part isn't, is there a reason for that?

zigavehovec (commented):
Some questions regarding `undefined` vs `null`, very nice otherwise! 👌

src/player.ts
rolandkakonyi: @zigavehovec actually that's a mistake from my side.
I planned having this as nullable but when making the iOS side nullable, the React Native runtime throws an error that it can not be nullable to keep it compatible with Android. Don't ask me why, maybe a Java restriction with JS interop?

Removed the nullability from Android here: b88b5de

src/player.ts
```diff
@@ -448,4 +448,14 @@ export class Player extends NativeInstance<PlayerConfig> {
...
     return PlayerModule.getMaxTimeShift(this.nativeId);
   };
+
+  /**
+   * Sets the upper bitrate boundary for video qualities. All qualities with a bitrate
+   * that is higher than this threshold will not be eligible for automatic quality selection.
+   *
+   * Can be set to `undefined` for no limitation.
+   */
+  setMaxSelectableBitrate = (bitrate: number | undefined) => {
```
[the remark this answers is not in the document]
rolandkakonyi: It felt natural that if max. selectable bitrate is `undefined`, then there is no limit.

Changed to nullable here: d9683f2

src/player.ts
zigavehovec: I see, thanks for the clarification!

zigavehovec (approved):

PR #224, opened 2023-09-06
Add missing subtitle and audio APIs

rolandkakonyi:
## Description

This PR extends the `Player` API to allow fetching the currently selected tracks.

## Changes

Introduced the following APIs:
- `Player.getAudioTrack` to allow fetching currently selected audio track
- `Player.setAudioTrack` to allow fetching currently selected subtitle track

## Checklist
- [x] 🗒 `CHANGELOG` entry

PR #225, opened 2023-09-06
Add `getThumbnail` APIs to `Player` and `Source`

rolandkakonyi:
## Description

This PR extends the `Player` and `Source` APIs to allow fetching thumbnail information.

## Changes

Introduced:
- `Player.getThumbnail()`: to fetch thumbnail information for the current source
- `Source.getThumbnail()`: to fetch thumbnail information for the given source

## Checklist
- [x] 🗒 `CHANGELOG` entry


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## [Unreleased]
+
+### Added
+
+- Support to get thumbnail images via `Player.getThumbnail` and `Source.getThumbnail`
```
zigavehovec, line 7: I would prefer the changelog style that you used in #224 

```suggestion
- `Player.getThumbnail` and `Source.getThumbnail` APIs to get thumbnail images
```

android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt
```diff
@@ -428,6 +428,18 @@ class PlayerModule(private val context: ReactApplicationContext) : ReactContextB
...
 
+    /**
+     * Returns the thumbnail image for the active `Source` at a certain time.
+     * @param nativeId Target player id.
+     * @param time Playback time for the thumbnail.
+     */
+    @ReactMethod
+    fun getThumbnail(nativeId: NativeId, time: Double, promise: Promise) {
+        uiManager()?.addUIBlock {
+            promise.resolve(JsonConverter.fromThumbnail(players[nativeId]?.source?.getThumbnail(time)))
```
zigavehovec, line 439: `Player.getThumbnail` has been deprecated in favour of `Source.getThumbnail` on Android. From my perspective I would be in favour of only wrapping `Source.getThumbnail` in this PR, but I'm not sure what the state is on iOS, just wanted to point this out.

zigavehovec (commented):
Nice work 👍

CHANGELOG.md
rolandkakonyi: Good point!
Taking the suggested change!

android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt
rolandkakonyi: @zigavehovec I know about the deprecation on Android but I went for providing both as our `Source` API is not that rich in the React Native SDK yet. Also we don't support playlist, so only one `Source` is available at a time.

zigavehovec (approved):

PR #236, opened 2023-09-12
Migrate to the player bundled Analytics API

matamegger:
## Description

The underlying Android and iOS player come bundled together with the analytics collector i.e. the player libraries directly integrate analytics, so there is not need to manually setup and attach the collector to the player.

## Changes

- ⚠️ **(Breaking change)**
As the exposed analytics API was partially deprecated, this PR migrates the API to the new version of analytics. 
- On the native sides, the code is now using the bundled analytics collector

Other highlights are as in the changelog:
**Added:**
- `DefaultMetadata` for configuration of the bundled analytics collector
- `Player.analytics` to access the `AnalyticsApi` and interact with the bundled analytics collector
- `SourceConfig.analyticsSourceMetadata` for extended configuration of the bundled analytics collector

**Changed:**
- `AnalyticsConfig` properties to match the bitmovin analytics v3 API

**Removed:**
- `AnalyticsCollector` in favor of the bundled analytics functionality
- `CdnProvider`, as the property on the `AnalyticsConfig` is now a `string`

## Checklist
- [x] 🗒 `CHANGELOG` entry


src/analytics/config.ts
```diff
@@ -1,66 +1,77 @@
...
  */
-export enum CdnProvider {
-  BITMOVIN = 'bitmovin',
-  AKAMAI = 'akamai',
-  FASTLY = 'fastly',
-  MAXCDN = 'maxcdn',
-  CLOUDFRONT = 'cloudfront',
-  CHINACACHE = 'chinacache',
-  BITGRAVITY = 'bitgravity',
+export interface AnalyticsConfig {
```
testcenter, line 4: is it intended to not support `backendUrl` and `retryPolicy` of the `AnalyticsConfig`? :thinking:

testcenter (approved):
lgtm :+1:

src/analytics/config.ts
matamegger: Yes, those are features that where not supported before.
I mainly wanted to migrate to the new API, without adding new stuff. I don't see a particular need for them at the moment and adding them should be straight forward then.

android/src/main/java/com/bitmovin/player/reactnative/AnalyticsModule.kt
```diff
@@ -11,172 +10,35 @@ private const val MODULE_NAME = "AnalyticsModule"
...
-                    "[AnalyticsModule]", "Could not find source for player ($playerId)"
-                )
-                collector == null -> Log.d(
-                    "[AnalyticsModule]", "Could not find analytics collector ($nativeId)"
-                )
-                sourceMetadata == null -> Log.d(
-                    "[AnalyticsModule]", "Could not convert source metadata, thus they are not applied to the collector ($nativeId)"
-                )
-                else -> collector.addSourceMetadata(source,  sourceMetadata)
+                playerModule()?.getPlayer(nativeId)?.analytics?.sendCustomDataEvent(it)
```
rolandkakonyi, line 28: Should this be implemented on `PlayerModule` now?
Do we still need `AnalyticsModule`?

ios/AnalyticsModule.swift
```diff
@@ -30,190 +27,42 @@ class AnalyticsModule: NSObject, RCTBridgeModule {
...
-
-    /**
-     Updates the custom data config for a `BitmovinPlayerCollector` instance.
-     - Parameter nativeId: Native Id of the collector instance.
      - Parameter json: Custom data config json.
      */
-    @objc(setCustomDataOnce:json:)
-    func setCustomDataOnce(_ nativeId: NativeId, json: Any?) {
+    @objc(sendCustomDataEvent:json:)
+    func sendCustomDataEvent(_ playerId: NativeId, json: Any?) {
```
rolandkakonyi, line 36: Same here, should we move this to `PlayerModule`?

rolandkakonyi (commented):
Manually tested and looks good!

android/src/main/java/com/bitmovin/player/reactnative/AnalyticsModule.kt
matamegger: In favour of not making the `PlayerModule` and `SourceModule` unnecessarily big I do like the separation into a separate `...Module`.

What do you think about renaming the `AnalyticsModule` to `PlayerAnalyticsModule` and introducing a `SourceAnalyticsModule` once we expose the API to react?

rolandkakonyi: That is a good point.
I really like the idea of `PlayerAnalyticsModule`!

rolandkakonyi (approved):

PR #239, opened 2023-09-18
Define public API for casting

rolandkakonyi:
## Description

This PR defines the public API regarding the player and events, related for the Casting Feature for Android and iOS.

## Changes

- Added the player APIs needed for the casting feature.
- Added casting related events.

## Checklist
- [x] 🗒 `CHANGELOG` entry

### DIFF
--- a/src/events.ts
+++ b/src/events.ts
@@ -554,3 +554,77 @@ export interface VideoPlaybackQualityChangedEvent extends Event {
    */
   oldVideoQuality: VideoQuality;
 }
+
+/**
+ * Emitted when casting to a cast-compatible device is available.
+ */
+export interface CastAvailableEvent extends Event {}
+
+/**
+ * Emitted when the playback on a cast-compatible device was paused.
+ */
+export interface CastPaused extends Event {}
+
+/**
+ * Emitted when the playback on a cast-compatible device has finished.
+ */
+export interface CastPlaybackFinished extends Event {}
+
+/**
+ * Emitted when playback on a cast-compatible device has started.
+ */
+export interface CastPlaying extends Event {}
+
+/**
+ * Emitted when the cast app is launched successfully.
+ */
+export interface CastStarted extends Event {
+  /**
+   * The name of the cast device on which the app was launched.
+   */
+  deviceName: string | null;
+}
+
+/**
+ * Emitted when casting is initiated, but the user still needs to choose which device should be used.
+ */
+export interface CastStart extends Event {}
+
+/**
+ * Emitted when casting to a cast-compatible device is stopped.
+ */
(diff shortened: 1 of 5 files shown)


src/events.ts
```diff
@@ -554,3 +554,83 @@ export interface VideoPlaybackQualityChangedEvent extends Event {
...
+ * Emitted when casting to a cast-compatible device is available.
+ */
+export interface CastAvailableEvent extends Event {}
+
+/**
+ * Emitted when the playback on a cast-compatible device was paused.
+ *
+ * On Android `PausedEvent` is also emitted while casting.
+ */
+export interface CastPaused extends Event {}
```
strangesource, line 568:
```suggestion
export interface CastPausedEvent extends Event {}
```
For consistency with existing event naming. Holds for all events below.

src/player.ts
```diff
@@ -491,4 +496,42 @@ export class Player extends NativeInstance<PlayerConfig> {
...
+   * should be sent.
+   *
+   * @platform iOS, Android
+   */
+  castVideo = () => {
+    PlayerModule.castVideo(this.nativeId);
+  };
+
+  /**
+   * Stops casting the current video. Has no effect if [isCasting] is false.
```
strangesource, line 530:
```suggestion
   * Stops casting the current video. Has no effect if `isCasting` is false.
```
I don't think `[]` linking is supported in typescript

src/remoteControlConfig.ts
```diff
@@ -0,0 +1,37 @@
...
+  /**
+   * A Map containing custom configuration values that are sent to the remote control receiver.
+   * Default value is an empty map.
+   */
+  customReceiverConfig?: Record<string, string | null>;
+  /**
+   * Whether casting is enabled.
+   * Default value is `true`.
+   *
+   * Has no effect if the BitmovinCastManager is not initialized in the app's context.
```
strangesource, line 19:
```suggestion
   * Has no effect if the `BitmovinCastManager` is not initialized in the app's context.
```
Does that make sense for iOS as well?

strangesource (commented):
Looking good in general, just some minor things/questions.

src/events.ts
rolandkakonyi: Oops, I slipped on the multi-cursor edit.
Fixed in 56ee8d3

src/player.ts
rolandkakonyi: Fixed in 47a8071

src/remoteControlConfig.ts
rolandkakonyi: Improved in 1d77a80

strangesource (approved):
Nice work!

PR #240, opened 2023-09-19
Use `when` instead of `if` in Android event serialization

strangesource:
## Description
Multiple `if` clause without `else if` do not prevent duplicate declaration of events and is not readable.

## Changes
Use a `when` instead of multiple `if` in order to improve readability and avoid duplicate event declaration.

The changes do not modify any functionality.

## Checklist
- [ ] 🗒 `CHANGELOG` entry

### DIFF
--- a/android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
@@ -442,7 +442,7 @@ class JsonConverter {
                 }
 
                 else -> {
-                    // Event is not supported yet
+                    // Event is not supported yet or does not have any additional data
                 }
             }
             return json
@@ -562,7 +562,7 @@ class JsonConverter {
                 }
 
                 else -> {
-                    // Event is not supported yet
+                    // Event is not supported yet or does not have any additional data
                 }
             }
             return json
--- a/android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
@@ -396,39 +396,54 @@ class JsonConverter {
             val json = Arguments.createMap()
             json.putString("name", event.getName())
             json.putDouble("timestamp", event.timestamp.toDouble())
-            if (event is SourceEvent.Load) {
-                json.putMap("source", fromSource(event.source))
-            }
-            if (event is SourceEvent.Loaded) {
-                json.putMap("source", fromSource(event.source))
-            }
-            if (event is SourceEvent.Error) {
-                json.putInt("code", event.code.value)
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
```diff
@@ -396,39 +396,54 @@ class JsonConverter {
...
+                    json.putMap("subtitleTrack", fromSubtitleTrack(event.subtitleTrack))
+                }
+
+                is SourceEvent.SubtitleTrackChanged -> {
+                    json.putMap("oldSubtitleTrack", fromSubtitleTrack(event.oldSubtitleTrack))
+                    json.putMap("newSubtitleTrack", fromSubtitleTrack(event.newSubtitleTrack))
+                }
+
+                else -> {
+                    // Event is not supported yet or does not have any additional data
```
matamegger, line 445: A `Log` would be nice :)

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
```diff
@@ -443,86 +458,112 @@ class JsonConverter {
...
+                    json.putDouble("timeOffset", event.timeOffset)
+                }
+
+                is PlayerEvent.VideoPlaybackQualityChanged -> {
+                    json.putMap("newVideoQuality", fromVideoQuality(event.newVideoQuality))
+                    json.putMap("oldVideoQuality", fromVideoQuality(event.oldVideoQuality))
+                }
+
+                else -> {
+                    // Event is not supported yet or does not have any additional data
```
matamegger, line 565: A `Log` would be nice :)

matamegger (approved):
Nice improvement!

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
strangesource: I don't think it's a good idea to log here as there are many events that just don't have any additional data that would suddenly fill up the logs even though they are fully supported. 
In order to have a meaningful log we would need to explicitly include those in the 'when' entry as supported, but I consider this out of scope for this quick improvement.

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
strangesource: https://github.com/bitmovin/bitmovin-player-react-native/pull/240#discussion_r1329690343

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
matamegger: Makes sense!

PR #241, opened 2023-09-19
Add Android API for casting on `PlayerModule`

strangesource:
## Description
Android specific API on the `PlayerModule` is not yet exposed.

## Changes
Expose Android specific API on the `PlayerModule` as `ReactMethod`.

## Checklist
- [x] 🗒 `CHANGELOG` entry

### DIFF
--- a/android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt
@@ -4,6 +4,7 @@ import android.util.Log
 import com.bitmovin.analytics.api.DefaultMetadata
 import com.bitmovin.player.api.Player
 import com.bitmovin.player.api.analytics.create
+import com.bitmovin.player.api.event.PlayerEvent
 import com.bitmovin.player.reactnative.converter.JsonConverter
 import com.facebook.react.bridge.*
 import com.facebook.react.module.annotations.ReactModule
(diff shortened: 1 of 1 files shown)


rolandkakonyi (approved):

android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt
```diff
@@ -496,6 +497,48 @@ class PlayerModule(private val context: ReactApplicationContext) : ReactContextB
...
+    }
+
+    /**
+     * Whether casting to a cast-compatible remote device is available. [PlayerEvent.CastAvailable] signals when
+     * casting becomes available.
+     */
+    @ReactMethod
+    fun isCastAvailable(nativeId: NativeId, promise: Promise) {
+        uiManager()?.addUIBlock {
+            promise.resolve(players[nativeId]?.isCastAvailable)
```
matamegger, line 528: Should we fallback to `false` if no player instance could be found?

android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt
```diff
@@ -496,6 +497,48 @@ class PlayerModule(private val context: ReactApplicationContext) : ReactContextB
...
+        }
+    }
+
+    /**
+     * Whether video is currently being casted to a remote device and not played locally.
+     */
+    @ReactMethod
+    fun isCasting(nativeId: NativeId, promise: Promise) {
+        uiManager()?.addUIBlock {
+            promise.resolve(players[nativeId]?.isCasting)
```
matamegger, line 538: Same as above

android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt
strangesource: Valid remark. I do think my current solution is in line with the existing patterns (e.g. `isAd` or `isLive`). In addition it does make sense semantically in my opinion. 

(In theory we could even return the correct value from the `BitmovinCastManager` even if we don't have a player but that's not in line with what we have on native.)

matamegger: Having a fallback is probably not ideal. Missing the player instance is a serious problem and may not happen as it is a broken fundamental framework.

Defining and implementing a proper solution is independent of the current changes.

matamegger (approved):

PR #242, opened 2023-09-19
Define `BitmovinCastManager` API

rolandkakonyi:
## Description

We need to expose `BitmovinCastManager` APIs to enable casting features.

## Changes

Introduced `BitmovinCastManager` and its APIs.

## Checklist
- [x] 🗒 `CHANGELOG` entry - already exists in #239

### DIFF
--- a/src/bitmovinCastManager.ts
+++ b/src/bitmovinCastManager.ts
@@ -9,12 +9,12 @@ const BitmovinCastManagerModule = NativeModules.BitmovinCastManagerModule;
 export interface BitmovinCastManagerOptions {
   /**
    * ID of receiver application.
-   * Using `nil` value will result in using the default application ID
+   * Using `null` value will result in using the default application ID
    */
   applicationId?: string | null;
   /**
    * A custom message namespace to be used for communication between sender and receiver.
-   * Using `nil` value will result in using the default message namespace
+   * Using `null` value will result in using the default message namespace
    */
   messageNamespace?: string | null;
 }
--- a/src/bitmovinCastManager.ts
+++ b/src/bitmovinCastManager.ts
@@ -68,17 +68,4 @@ export const BitmovinCastManager = {
     }
     return BitmovinCastManagerModule.sendMessage(message, messageNamespace);
   },
-
-  /**
-   * Sends the given metadata wrapped in a metadata message object to the cast receiver on the configured message namespace.
-   * The provided metadata must be JSON serializable.
-   * @param metadata The metadata to be sent
-   * @returns A promise that resolves when the metadata was sent successfully
-   */
-  sendMetadata: (metadata: Record<string, any>) => {
-    if (Platform.OS === 'ios' && Platform.isTV) {
-      return Promise.resolve();
-    }
-    return BitmovinCastManagerModule.sendMetadata(metadata);
-  },
 };
--- a/src/bitmovinCastManager.ts
+++ b/src/bitmovinCastManager.ts
@@ -1,0 +1,84 @@
+import { NativeModules, Platform } from 'react-native';
+
+const BitmovinCastManagerModule = NativeModules.BitmovinCastManagerModule;
+
+/**
+ * The options to be used for initializing `BitmovinCastManager`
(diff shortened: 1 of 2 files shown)


src/bitmovinCastManager.ts
```diff
@@ -0,0 +1,84 @@
...
+   *
+   * @param options The options to be used for initializing `BitmovinCastManager`
+   * @returns A promise that resolves when the `BitmovinCastManager` was initialized successfully
+   */
+  initialize: async (
+    options: BitmovinCastManagerOptions | null = null
+  ): Promise<void> => {
+    if (Platform.OS === 'ios' && Platform.isTV) {
+      return Promise.resolve();
+    }
```
strangesource, line 54: Should we document this on the function or even the interface?

src/bitmovinCastManager.ts
```diff
@@ -0,0 +1,84 @@
...
+   * The provided metadata must be JSON serializable.
+   * @param metadata The metadata to be sent
+   * @returns A promise that resolves when the metadata was sent successfully
+   */
+  sendMetadata: (metadata: Record<string, any>) => {
+    if (Platform.OS === 'ios' && Platform.isTV) {
+      return Promise.resolve();
+    }
+    return BitmovinCastManagerModule.sendMetadata(metadata);
+  },
```
strangesource, line 83: I'd rather not include this. The functionality is deprecated on Android as it requires serialization of user objects - which is difficult and sometimes not intuitive on Android.
While this is maybe more valid in the React Native context, the serialization can be easily done by the customer and they can use `sendMessage` then.

strangesource (commented):
Nice, just one remark and some minor NIT.

src/bitmovinCastManager.ts
```diff
@@ -0,0 +1,84 @@
...
+   * A custom message namespace to be used for communication between sender and receiver.
+   * Using `nil` value will result in using the default message namespace
+   */
+  messageNamespace?: string | null;
+}
+
+/**
+ * Singleton providing access to GoogleCast related features.
+ * The `BitmovinCastManager` needs to be initialized by calling `BitmovinCastManager.initialize`
+ * before `Player` creation to enable casting features.
```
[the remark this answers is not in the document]
matamegger: Seems fine 👍 .
 On Android there is an auto initialization, but explicitly initializing always seems like a good idea.

src/bitmovinCastManager.ts
```diff
@@ -0,0 +1,84 @@
...
+const BitmovinCastManagerModule = NativeModules.BitmovinCastManagerModule;
+
+/**
+ * The options to be used for initializing `BitmovinCastManager`
+ * @platform Android, iOS
+ */
+export interface BitmovinCastManagerOptions {
+  /**
+   * ID of receiver application.
+   * Using `nil` value will result in using the default application ID
```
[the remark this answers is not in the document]
rolandkakonyi: 🤦‍♂️, fixed in 3ee1aef

src/bitmovinCastManager.ts
```diff
@@ -0,0 +1,84 @@
...
+   */
+  applicationId?: string | null;
+  /**
+   * A custom message namespace to be used for communication between sender and receiver.
+   * Using `nil` value will result in using the default message namespace
```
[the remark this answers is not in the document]
rolandkakonyi: same, fixed in 3ee1aef

src/bitmovinCastManager.ts
rolandkakonyi: @strangesource It is documented by `@platform Android, iOS` on the type.
This is for safety to not get into weird errors on tvOS.

strangesource: Ah, missed that this is part of the API. 👍

src/bitmovinCastManager.ts
rolandkakonyi: OK, let's skip this one. Removed in: bcbaf4d

strangesource (approved):

PR #244, opened 2023-09-19
Add `RemoteControlConfig` serialization on Android

strangesource:
## Description
`RemoteControlConfig` is not serialized to be accessible in React Native yet.

## Changes
Serialize `RemoteControlConfig` to be accessible from React Native. 
`onConfigureMediaInfo` is left out intentionally as this is a rather advanced feature that is rarely used.

## Checklist
- [x] 🗒 `CHANGELOG` entry

### DIFF
--- a/android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
@@ -18,6 +18,7 @@ import com.bitmovin.player.api.advertising.AdQuartile
 import com.bitmovin.player.api.advertising.AdSource
 import com.bitmovin.player.api.advertising.AdSourceType
 import com.bitmovin.player.api.advertising.AdvertisingConfig
+import com.bitmovin.player.api.casting.RemoteControlConfig
 import com.bitmovin.player.api.drm.WidevineConfig
 import com.bitmovin.player.api.event.PlayerEvent
 import com.bitmovin.player.api.event.SourceEvent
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
```diff
@@ -96,9 +98,67 @@ class JsonConverter {
...
 
+        /**
+         * Converts an arbitrary [ReadableMap] to a [RemoteControlConfig].
+         *
+         * @param json JS object representing the [RemoteControlConfig].
+         * @return The generated [RemoteControlConfig].
+         */
+        private fun toRemoteControlConfig(json: ReadableMap?): RemoteControlConfig? {
+            if (json == null) return null
+            val defaultRemoteControlConfig = RemoteControlConfig()
```
matamegger, line 117: I was already thinking of this pattern in the past. I was a bit reluctant to actually do it, but it probably is a good option for the use case we have here.

matamegger (approved):
👍

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
strangesource: I took the inspiration from one of our discussions earlier. This is the best (only?) pattern to take the default values defined in the native code as well as be prepared for a potential immutable configuration object in the future.

matamegger: I agree.

PR #248, opened 2023-09-19
Update `RemoteControlConfig`'s `customReceiverConfig` type

rolandkakonyi:
## Description

On iOS it only supports `String` to `String` maps for `RemoteControlConfig. customReceiverConfig`, while in the React Native API we defined `String` to optional `String`.

## Changes

Changed the public API to reflect our restrictions and modified the Android JSON deserialization to take the right values.

## Checklist

### DIFF
--- a/android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
@@ -126,7 +126,6 @@ class JsonConverter {
             if (json.hasKey("customReceiverConfig")) {
                 customReceiverConfig = json.getMap("customReceiverConfig")
                     ?.toHashMap()
-                    ?.filterValues { entry -> entry is String  }
                     ?.mapValues { entry -> entry.value as String } ?: emptyMap()
             }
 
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
```diff
@@ -126,7 +126,8 @@ class JsonConverter {
             if (json.hasKey("customReceiverConfig")) {
                 customReceiverConfig = json.getMap("customReceiverConfig")
                     ?.toHashMap()
-                    ?.mapValues { entry -> entry.value as? String? } ?: emptyMap()
+                    ?.filterValues { entry -> entry is String  }
+                    ?.mapValues { entry -> entry.value as String } ?: emptyMap()
```
rolandkakonyi, line 130: Please be ruthless here, maybe there is better way to do this in Kotlin

strangesource: I don't think it's necessary to filter here as we can assume that the input on the ts side is correct. If we stop having this assumption the conversion will get super messy.
```suggestion
                    ?.mapValues { entry -> entry.value as String } ?: emptyMap()
```

rolandkakonyi: Sounds good!
Fixed by accepting your suggestion.

strangesource (approved):

PR #249, opened 2023-09-19
Introduce `BitmovinCastManagerModule` for iOS

rolandkakonyi:
## Description

`BitmovinCastManagerModule` is not implemented for iOS

## Changes

Implemented `BitmovinCastManagerModule` for iOS

## Checklist

### DIFF
--- a/src/bitmovinCastManager.ts
+++ b/src/bitmovinCastManager.ts
@@ -52,7 +52,7 @@ export const BitmovinCastManager = {
     if (Platform.OS === 'ios' && Platform.isTV) {
       return Promise.resolve();
     }
-    return BitmovinCastManagerModule.initialize(options);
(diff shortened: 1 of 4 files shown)


src/bitmovinCastManager.ts
```diff
@@ -43,6 +43,8 @@ export const BitmovinCastManager = {
    * This method needs to be called before `Player` creation to enable casting features.
    * If no options are provided, the default options will be used.
    *
+   * IMPORTANT: This should only be called when the Google Cast SDK is available in the application.
+   *
```
rolandkakonyi, line 47: @strangesource I added this as on iOS `initializeCasting` crashes the app when Cast SDK is not in the app. How is it on Android?

strangesource: Same on Android, it crashes as well.

PR #250, opened 2023-09-19
Fix non-DRM playback on iOS

rolandkakonyi:
## Description

Found that in #236 we accidentally broke non-DRM playback when Analytics is not enabled.

## Changes

Fixed the `Source` config parsing and `Source` creation.

## Checklist
- [ ] 🗒 `CHANGELOG` entry


matamegger (approved):
Sorry for the oversight!

PR #252, opened 2023-09-20
Implement events for iOS

rolandkakonyi:
## Description

Player events are defined but not forwarded to the React side.

## Changes

Hooked up to forward the events to React.

## Checklist


src/events.ts
```diff
@@ -630,7 +630,7 @@ export interface CastPayload {
  */
 export interface CastWaitingForDeviceEvent extends Event {
   /**
-   * The [CastPayload] object for the event
+   * The `CastPayload` object for the event
```
rolandkakonyi, line 633: Sneak to fix documentation

PR #254, opened 2023-09-20
Implement `BitmovinCastManagerModule` on Android

strangesource:
## Description
`BitmovinCastManagerModule` does not yet have an Android implementation. In addition, Android requires an additional `updateContext` function to pass the current context into the manager.

## Changes
Implement `BitmovinCastManagerModule` and add `updateContext` function to the public API for Android only.

## Checklist
- [x] 🗒 `CHANGELOG` entry

### DIFF
--- a/android/src/main/java/com/bitmovin/player/reactnative/BitmovinCastManagerModule.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/BitmovinCastManagerModule.kt
@@ -22,7 +22,9 @@ class BitmovinCastManagerModule(
      * Returns whether the [BitmovinCastManager] is initialized.
      */
     @ReactMethod
-    fun isInitialized() = BitmovinCastManager.isInitialized()
+    fun isInitialized(promise: Promise) = uiManager?.addUIBlock {
+        promise.resolve(BitmovinCastManager.isInitialized())
+    }
 
     /**
      * Initializes the [BitmovinCastManager] with the given options.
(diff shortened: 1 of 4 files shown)


android/src/main/java/com/bitmovin/player/reactnative/BitmovinCastManagerModule.kt
```diff
@@ -0,0 +1,74 @@
...
+class BitmovinCastManagerModule(
+    private val context: ReactApplicationContext,
+) : ReactContextBaseJavaModule(context) {
+    override fun getName() = MODULE_NAME
+
+    /**
+     * Returns whether the [BitmovinCastManager] is initialized.
+     */
+    @ReactMethod
+    fun isInitialized() = BitmovinCastManager.isInitialized()
```
rolandkakonyi, line 25: Don't we need to use a `Promise` here?
Or will this be called and returned synchronously?

matamegger (approved):
👌

android/src/main/java/com/bitmovin/player/reactnative/BitmovinCastManagerModule.kt
strangesource: good catch. [64998be](https://github.com/bitmovin/bitmovin-player-react-native/pull/254/commits/64998bef11a25e4358879f7a05c0e23ac1649b1d)

PR #257, opened 2023-09-21
Fix cast issues on Android

strangesource:
## Description
Casting was not yet properly working as 
- the `ExpandedControllerActivity` was not declared in the manifest yet
- the cast context needs to be initialized in the `onCreate` of the activity (see [here](https://github.com/react-native-google-cast/react-native-google-cast/blob/4a2a129761acf13bfcf4fb0835d37ee0d5c74ff0/docs/getting-started/setup.md#android) and [here](https://stackoverflow.com/questions/58884569/chromecast-button-only-appears-after-backgrounding-the-app)
- `updateContext` was not yet called on Android

## Changes
- feat(casting): move dependencies to a more sensible location
- feat(casting): add expanded controller activity to manifest
- feat(casting): eagerly initialize cast context
- feat(casting): call `updateContext` on Android

## Checklist
- [ ] 🗒 `CHANGELOG` entry

### DIFF
--- a/example/src/screens/Casting.tsx
+++ b/example/src/screens/Casting.tsx
@@ -1,5 +1,5 @@
 import React, { useCallback } from 'react';
-import { View, StyleSheet } from 'react-native';
+import { View, StyleSheet, Platform } from 'react-native';
 import { useFocusEffect } from '@react-navigation/native';
 import {
   Event,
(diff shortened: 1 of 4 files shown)


example/src/screens/Casting.tsx
```diff
@@ -16,6 +16,11 @@ function prettyPrint(header: string, obj: any) {
 export default function Casting() {
   BitmovinCastManager.initialize();
 
+  if (Platform.OS === 'android') {
```
matamegger, line 19: Is it safe to compare using `===` in this case?
Could it be that the strings are not the same, but only equal?

matamegger (approved):
Looks good!

example/src/screens/Casting.tsx
rolandkakonyi: I think this is the preferred way for TypeScript. Even the linter warns about it!

rolandkakonyi (approved):

PR #259, opened 2023-09-21
Enable casting different source on iOS

rolandkakonyi:
## Description

On the iOS SDK we support casting a different source config when casting, this is not yet available for the React Native SDK.

## Changes

Added support to configure a cast-specific `SourceConfig` for casting, via adding a new namespace, `remotePlayback` to `Source`.

## Checklist

### DIFF
--- a/android/src/main/java/com/bitmovin/player/reactnative/SourceModule.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/SourceModule.kt
@@ -46,7 +46,7 @@ class SourceModule(private val context: ReactApplicationContext) : ReactContextB
      * @param nativeId ID to be associated with the `Source` instance.
      * @param drmNativeId ID of the DRM config to use.
      * @param config `SourceConfig` object received from JS.
-     * @param sourceRemoteControlConfig `SourceRemoteControlConfig` object received from JS.
+     * @param sourceRemoteControlConfig `SourceRemoteControlConfig` object received from JS. Not supported on Android.
      * @param analyticsSourceMetadata `SourceMetadata` object received from JS.
      */
     @ReactMethod
@@ -71,7 +71,7 @@ class SourceModule(private val context: ReactApplicationContext) : ReactContextB
      * @param nativeId ID to be associated with the `Source` instance.
      * @param drmNativeId ID of the DRM config to use.
      * @param config `SourceConfig` object received from JS.
-     * @param sourceRemoteControlConfig `SourceRemoteControlConfig` object received from JS.
+     * @param sourceRemoteControlConfig `SourceRemoteControlConfig` object received from JS. Not supported on Android.
      */
     @ReactMethod
     fun initWithConfig(
--- a/android/src/main/java/com/bitmovin/player/reactnative/SourceModule.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/SourceModule.kt
@@ -53,7 +53,7 @@ class SourceModule(private val context: ReactApplicationContext) : ReactContextB
         nativeId: NativeId,
         drmNativeId: NativeId?,
         config: ReadableMap?,
-        remotePlaybackJson: ReadableMap?,
+        sourceRemotePlaybackConfig: ReadableMap?,
         analyticsSourceMetadata: ReadableMap?
     ) {
         uiManager()?.addUIBlock {
@@ -76,7 +76,7 @@ class SourceModule(private val context: ReactApplicationContext) : ReactContextB
         nativeId: NativeId,
         drmNativeId: NativeId?,
         config: ReadableMap?,
-        remotePlaybackJson: ReadableMap?
+        sourceRemotePlaybackConfig: ReadableMap?
     ) {
         uiManager()?.addUIBlock {
             initializeSource(nativeId, drmNativeId, config) { sourceConfig ->
--- a/android/src/main/java/com/bitmovin/player/reactnative/SourceModule.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/SourceModule.kt
@@ -46,6 +46,7 @@ class SourceModule(private val context: ReactApplicationContext) : ReactContextB
      * @param nativeId ID to be associated with the `Source` instance.
      * @param drmNativeId ID of the DRM config to use.
      * @param config `SourceConfig` object received from JS.
+     * @param sourceRemoteControlConfig `SourceRemoteControlConfig` object received from JS.
      * @param analyticsSourceMetadata `SourceMetadata` object received from JS.
      */
     @ReactMethod
@@ -53,7 +54,7 @@ class SourceModule(private val context: ReactApplicationContext) : ReactContextB
         nativeId: NativeId,
         drmNativeId: NativeId?,
         config: ReadableMap?,
-        sourceRemotePlaybackConfig: ReadableMap?,
+        sourceRemoteControlConfig: ReadableMap?,
         analyticsSourceMetadata: ReadableMap?
     ) {
         uiManager()?.addUIBlock {
@@ -70,13 +71,14 @@ class SourceModule(private val context: ReactApplicationContext) : ReactContextB
      * @param nativeId ID to be associated with the `Source` instance.
      * @param drmNativeId ID of the DRM config to use.
      * @param config `SourceConfig` object received from JS.
+     * @param sourceRemoteControlConfig `SourceRemoteControlConfig` object received from JS.
      */
     @ReactMethod
     fun initWithConfig(
         nativeId: NativeId,
         drmNativeId: NativeId?,
         config: ReadableMap?,
-        sourceRemotePlaybackConfig: ReadableMap?
+        sourceRemoteControlConfig: ReadableMap?
     ) {
         uiManager()?.addUIBlock {
             initializeSource(nativeId, drmNativeId, config) { sourceConfig ->
--- a/android/src/main/java/com/bitmovin/player/reactnative/SourceModule.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/SourceModule.kt
@@ -53,6 +53,7 @@ class SourceModule(private val context: ReactApplicationContext) : ReactContextB
         nativeId: NativeId,
         drmNativeId: NativeId?,
         config: ReadableMap?,
+        remotePlaybackJson: ReadableMap?,
         analyticsSourceMetadata: ReadableMap?
     ) {
         uiManager()?.addUIBlock {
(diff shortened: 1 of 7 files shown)


android/src/main/java/com/bitmovin/player/reactnative/SourceModule.kt
rolandkakonyi, line 1: Added these to keep the caller side unified

ios/PlayerModule.swift
```diff
@@ -45,7 +45,18 @@ class PlayerModule: NSObject, RCTBridgeModule {
...
+
+            playerConfig.remoteControlConfig.prepareSource = { [weak self] type, sourceConfig in
+                guard let sourceModule = self?.bridge[SourceModule.self],
+                      let sourceNativeId = sourceModule.nativeId(where: { $0.sourceConfig === sourceConfig }),
+                      let castSourceConfig = sourceModule.retrieveCastSourceConfig(sourceNativeId) else {
+                    return nil
+                }
+
+                return castSourceConfig
+            }
```
rolandkakonyi, line 59: We have to look up the cast-specific source config on the fly and we have to use `sourceConfig` to do that, as `player.source` returns `nil` when this is called.

ios/PlayerModule.swift
```diff
@@ -662,4 +666,16 @@ class PlayerModule: NSObject, RCTBridgeModule {
...
+
+    private func setupRemoteControlConfig(_ remoteControlConfig: RemoteControlConfig) {
+        remoteControlConfig.prepareSource = { [weak self] type, sourceConfig in
+            guard let sourceModule = self?.bridge[SourceModule.self],
+                  let sourceNativeId = sourceModule.nativeId(where: { $0.sourceConfig === sourceConfig }),
+                  let castSourceConfig = sourceModule.retrieveCastSourceConfig(sourceNativeId) else {
+                return nil
+            }
+
+            return castSourceConfig
```
rolandkakonyi, line 678: We have to look up the cast-specific source config on the fly and we have to use `sourceConfig` to do that, as `player.source` returns `nil` when this is called.

rolandkakonyi:
@strangesource you don't need to review the Swift changes, but please review the TypeScript and Android changes!

src/source.ts
```diff
@@ -136,6 +136,21 @@ export interface SourceConfig extends NativeInstanceConfig {
   analyticsSourceMetadata?: SourceMetadata;
 }
 
+/**
+ * The remote playback config for a source.
+ * @platform iOS
+ */
+export interface SourceRemotePlaybackConfig {
```
strangesource, line 143: NIT: The naming might be a bit problematic as it suggests that it has something to to with a remote `PlaybackConfig`. Is the idea that this will in the future potentially contain more than just the `castSourceConfig`?

android/src/main/java/com/bitmovin/player/reactnative/SourceModule.kt
```diff
@@ -53,6 +53,7 @@ class SourceModule(private val context: ReactApplicationContext) : ReactContextB
         nativeId: NativeId,
         drmNativeId: NativeId?,
         config: ReadableMap?,
+        sourceRemotePlaybackConfig: ReadableMap?,
```
strangesource, line 56: Let's add it to the function documentation and mention that this is unused/unsupported on Android. Holds for both functions.

strangesource (commented):
Some small remarks/questions from my side.

ios/SourceModule.m
```diff
@@ -2,8 +2,15 @@
...
-RCT_EXTERN_METHOD(initWithAnalyticsConfig:(NSString *)nativeId drmNativeId:(NSString *)drmNativeId config:(nullable id)config analyticsSourceMetadata:(nullable id)analyticsSourceMetadata)
+RCT_EXTERN_METHOD(initWithConfig:(NSString *)nativeId
+                  drmNativeId:(NSString *)drmNativeId
+                  config:(nullable id)config
+                  sourceRemotePlaybackConfig:(id)remotePlayback)
+RCT_EXTERN_METHOD(initWithAnalyticsConfig:(NSString *)nativeId
+                  drmNativeId:(NSString *)drmNativeId
+                  config:(nullable id)config
+                  sourceRemotePlaybackConfig:(id)remotePlayback
+                  analyticsSourceMetadata:(nullable id)analyticsSourceMetadata)
```
hawk23, line 13: Nit: Seems like the convention to align with `:` is not a thing in this file. However, looking at the rest of the file, it seems like the indentation is one tab size and not based on `(`.

ios/SourceModule.swift
```diff
@@ -66,8 +90,13 @@ class SourceModule: NSObject, RCTBridgeModule {
...
      - Parameter config: `SourceConfig` object received from JS.
      */
-    @objc(initWithConfig:drmNativeId:config:)
-    func initWithConfig(_ nativeId: NativeId, drmNativeId: NativeId?, config: Any?) {
+    @objc(initWithConfig:drmNativeId:config:sourceRemotePlaybackConfig:)
+    func initWithConfig(
+        _ nativeId: NativeId,
+        drmNativeId: NativeId?,
+        config: Any?,
+        sourceRemotePlaybackConfig: Any
```
hawk23, line 98: Should this maybe be `Any?` as it is optional?

ios/SourceModule.swift
```diff
@@ -32,15 +35,33 @@ class SourceModule: NSObject, RCTBridgeModule {
...
      - Parameter analyticsSourceMetadata: `SourceMetadata` object received from JS.
      */
-    @objc(initWithAnalyticsConfig:drmNativeId:config:analyticsSourceMetadata:)
-    func initWithAnalyticsConfig(_ nativeId: NativeId, drmNativeId: NativeId?, config: Any?, analyticsSourceMetadata: Any?) {
+    @objc(initWithAnalyticsConfig:drmNativeId:config:sourceRemotePlaybackConfig:analyticsSourceMetadata:)
+    func initWithAnalyticsConfig(
+        _ nativeId: NativeId,
+        drmNativeId: NativeId?,
+        config: Any?,
+        sourceRemotePlaybackConfig: Any,
```
hawk23, line 62: Should this maybe be `Any?` as it is optional?

ios/PlayerModule.swift
rolandkakonyi: @strangesource `===` is not equality check but identity check. Therefore it would need to be the same object instance (same pointer/memory address).
This would be a problem for a fully native project as you can create multiple sources with the same `SourceConfig` instance, BUT since in React Native we always parse the JSON and create a new `SourceConfig` object, the reuse of the same instance of `SourceConfig` can not happen.

src/source.ts
rolandkakonyi: I also struggled to find a name, this was like the 3rd one I used.
I actually like `SourceRemoteControlConfig` name, will change it to that.

rolandkakonyi: Renamed in 852d26f

android/src/main/java/com/bitmovin/player/reactnative/SourceModule.kt
rolandkakonyi: Fixed as part of the renaming in 852d26f

ios/PlayerModule.swift
strangesource: Thanks for the explanation. This should be correct then. It does require some deeper understanding of the domain to understand though - which is fine I guess.

src/source.ts
```diff
@@ -144,6 +159,13 @@ export class Source extends NativeInstance<SourceConfig> {
    * The native DRM config reference of this source.
    */
   drm?: Drm;
+  /**
+   * The remote playback config for this source.
+   * This is only used on iOS.
+   *
+   * @platform iOS
+   */
+  remotePlayback: SourceRemotePlaybackConfig | null = null;
```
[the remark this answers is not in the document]
rolandkakonyi: 😅 I just did that, then fixed in ddff297

ios/SourceModule.swift
rolandkakonyi: fixed in ee6f953

(Fun fact: you can store anything in `Any`, even optionals, so this would never fail. But looks better with `Any?` :) )

ios/SourceModule.swift
rolandkakonyi: fixed in ee6f953

ios/SourceModule.m
rolandkakonyi: fixed in 85c58e781a68f459db5d640ccaea2bcd48c2985c

ios/SourceModule.swift
```diff
@@ -32,15 +35,33 @@ class SourceModule: NSObject, RCTBridgeModule {
...
      - Parameter analyticsSourceMetadata: `SourceMetadata` object received from JS.
      */
-    @objc(initWithAnalyticsConfig:drmNativeId:config:analyticsSourceMetadata:)
-    func initWithAnalyticsConfig(_ nativeId: NativeId, drmNativeId: NativeId?, config: Any?, analyticsSourceMetadata: Any?) {
+    @objc(initWithAnalyticsConfig:drmNativeId:config:sourceRemoteControlConfig:analyticsSourceMetadata:)
+    func initWithAnalyticsConfig(
+        _ nativeId: NativeId,
+        drmNativeId: NativeId?,
+        config: Any?,
+        sourceRemoteControlConfig: Any?,
```
hawk23, line 62: Nit: I think we should also update the doc and include the new parameter. But also fine for me to remove the whole doc altogether if it is not really adding value (as this is not really public API). Same for down below

src/source.ts
```diff
@@ -136,6 +136,21 @@ export interface SourceConfig extends NativeInstanceConfig {
   analyticsSourceMetadata?: SourceMetadata;
 }
 
+/**
+ * The remote playback config for a source.
```
hawk23, line 140: Aiii
```suggestion
 * The remote control config for a source.
```

src/source.ts
```diff
@@ -144,6 +159,13 @@ export class Source extends NativeInstance<SourceConfig> {
    * The native DRM config reference of this source.
    */
   drm?: Drm;
+  /**
+   * The remote playback config for this source.
```
hawk23, line 163:
```suggestion
   * The remote control config for this source.
```

hawk23 (approved):
Just some minor things, so approving already

android/src/main/java/com/bitmovin/player/reactnative/SourceModule.kt
strangesource: Can you please also mention that this is unused/unsupported on Android? 😇

rolandkakonyi: missed that part, fixed in 2fe285f

ios/SourceModule.swift
rolandkakonyi: fixed in da16514

src/source.ts
rolandkakonyi: Looks like mass replace didn't work out 🤦 

Fixed in da16514

src/source.ts
rolandkakonyi: Fixed in da16514

strangesource (approved):
:+1

PR #261, opened 2023-09-21
Widevine DRM support for casting on iOS

rolandkakonyi:
## Description

We don't support DRM-protected playback for casting.

## Changes

Added `WidevineConfig` parsing to iOS and use it for remote playback.

## Checklist

### DIFF
--- a/ios/DrmModule.swift
+++ b/ios/DrmModule.swift
@@ -6,7 +6,7 @@ class DrmModule: NSObject, RCTBridgeModule {
     @objc var bridge: RCTBridge!
 
     /// In-memory mapping from `nativeId`s to `FairplayConfig` instances.
-    private var drmConfigs: Registry<DrmConfig> = [:]
+    private var drmConfigs: Registry<FairplayConfig> = [:]
 
     /// JS module name.
     static func moduleName() -> String! {
@@ -28,7 +28,7 @@ class DrmModule: NSObject, RCTBridgeModule {
      - Parameter nativeId: ID to associate with the `FairplayConfig` instance.
      - Returns: The associated `FairplayConfig` instance or `nil`.
      */
-    @objc func retrieve(_ nativeId: NativeId) -> DrmConfig? {
+    @objc func retrieve(_ nativeId: NativeId) -> FairplayConfig? {
         drmConfigs[nativeId]
     }
 
@@ -42,11 +42,11 @@ class DrmModule: NSObject, RCTBridgeModule {
         bridge.uiManager.addUIBlock { [weak self] _, _ in
             guard
                 self?.drmConfigs[nativeId] == nil,
-                let drmConfig = RCTConvert.drmConfig(config)
+                let fairplayConfig = RCTConvert.drmConfig(config).fairplay
             else {
                 return
             }
-            self?.drmConfigs[nativeId] = drmConfig
+            self?.drmConfigs[nativeId] = fairplayConfig
             self?.initConfigBlocks(nativeId, config)
         }
     }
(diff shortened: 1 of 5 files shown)


ios/DrmModule.swift
```diff
@@ -42,7 +42,7 @@ class DrmModule: NSObject, RCTBridgeModule {
         bridge.uiManager.addUIBlock { [weak self] _, _ in
             guard
                 self?.drmConfigs[nativeId] == nil,
-                let fairplayConfig = RCTConvert.fairplayConfig(config)
+                let fairplayConfig = RCTConvert.drmConfig(config).fairplay
```
rolandkakonyi, line 45: Only use FairPlay DRM config if available as no other DRM system is supported on iOS/tvOS

ios/RCTConvert+BitmovinPlayer.swift
```diff
@@ -989,12 +1018,15 @@ extension RCTConvert {
...
+        guard let json = json as? [String: Any?],
+              let castSourceConfigJson = json["castSourceConfig"] as? [String: Any?] else {
             return nil
         }
-
         return SourceRemoteControlConfig(
-            castSourceConfig: RCTConvert.sourceConfig(json["castSourceConfig"])
+            castSourceConfig: RCTConvert.sourceConfig(
+                json["castSourceConfig"],
+                drmConfig: RCTConvert.drmConfig(castSourceConfigJson["drmConfig"]).widevine
```
rolandkakonyi, line 1028: Use only Widevine DRM config for casting as no other is supported for casting right now.

rolandkakonyi:
@strangesource please only review the TypeScript changes!

ios/RCTConvert+BitmovinPlayer.swift
```diff
@@ -335,33 +335,62 @@ extension RCTConvert {
         }
     }
 
+    /**
+     Utility method to get a `DrmConfig` from a JS object.
+     - Parameter json: JS object
+     - Returns: The generated `DrmConfig` object
+     */
```
hawk23, line 342: Nit: IMO doc like this is not adding any value. No additional information besides what I see in the method signature. But fine for me too keep it as it aligns with the general practice in this file

ios/RCTConvert+BitmovinPlayer.swift
```diff
@@ -335,33 +335,62 @@ extension RCTConvert {
...
             fairplayConfig.certificateRequestHeaders = certificateRequestHeaders
         }
         return fairplayConfig
     }
 
+    /**
+     Utility method to get a `WidevineConfig` from a JS object.
+     - Parameter json: JS object
+     - Returns: The generated `WidevineConfig` object
+     */
```
hawk23, line 381: Nit: IMO doc like this is not adding any value. No additional information besides what I see in the method signature. But fine for me too keep it as it aligns with the general practice in this file

hawk23 (approved):

src/drm/widevineConfig.ts
```diff
@@ -31,19 +33,25 @@ export interface WidevineConfig {
...
   /**
    * Indicates if the DRM sessions should be kept alive after a source is unloaded.
    * This allows DRM sessions to be reused over several different source items with the same DRM configuration as well
    * as the same DRM scheme information.
    * Default: `false`
+   *
+   * @platform Android
    */
-  shouldKeepDrmSessionsAlive: boolean;
+  shouldKeepDrmSessionsAlive?: boolean;
```
strangesource, line 56: NIT: This is not ideal because it suggests that this is not mandatory on Android. But I don't think there is a better solution without duplicating a lot of code.

strangesource (approved):

ios/RCTConvert+BitmovinPlayer.swift
rolandkakonyi: Could be useful for external or new contributors.
This was started by our initial contributor and we kept it consistent when taking over.

src/drm/widevineConfig.ts
rolandkakonyi: @strangesource It is not "mandatory to set" as it already has a default value in the Android Player SDK.
Therefore to enable using default value, without knowing it, we should make this optional to avoid duplicating default values.

PR #262, opened 2023-09-22
Casting support

rolandkakonyi:
## Description

While both the Android and iOS Bitmovin Player SDK support Cast SDK integration, the React Native Bitmovin Player SDK does not.

## Changes

Added support to use Cast SDK if it is available in the application.

## Checklist
- [x] 🗒 `CHANGELOG` entry


strangesource (approved):
👍Once all checks are green.

rolandkakonyi:
> 👍Once all checks are green.

Found some issues when building for tvOS, but fixed in 07943df

PR #264, opened 2023-09-22
Enable running CI workflow on all PRs

rolandkakonyi:
## Description

When opening a PR that is targeting other than the `development` branch we don't run the CI workflow right now.
This is problematic when working on a bigger feature with a chain of PRs open.

## Changes

Enabled running the CI workflow on all PRs.

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,7 @@ on:
       - '*.md'
 
   push:
+    branches: [development]
     paths-ignore:
       - '*.md'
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -2,12 +2,10 @@ name: CI
 
 on:
   pull_request:
-    branches: [development]
     paths-ignore:
       - '*.md'
 
   push:
-    branches: [development]
```
matamegger, line 10: I think we want to keep this. I think we only want the CI to run for push events to a pull request, but not on a push to any branch without PR.

rolandkakonyi: That makes sense, reverted in: e1bcaf6

matamegger (approved):

PR #270, opened 2023-09-26
implementation scaling mode dynamic android and ios

jonathanm-tkf:
## Description
Currently, we don't have the possibility to change the Scaling Mode like on other platforms, similar to the fullscreen button.

## Changes
```js
     <PlayerView
        scalingMode={scalingMode}
      />
```


krocard (changes requested):
Thank you for your pull request. After careful consideration, we believe the proposed approach of mutating the `player`'s `sytyleConfig` is not desirable:
 - changes will not be taken into account by the player (player events will not be triggered)
 - the scaling mode is a property of the `PlayerView`, not the `Player`
 - React Native doesn't allow UI side-effects in `@ReactMethod`, `UIManager` has to be used instead.

Here follows is our recommendation to implement `PlayerView`'s `ScalingMode`. As the scaling mode property is similar to the fullscreen one, we have used it as an example for each step.
 1. Add a `ScalingMode` parameter to the `PlayerView` function and `PlayerViewProps`
       - https://github.com/bitmovin/bitmovin-player-react-native/blob/a72652377fa9bc48ae2bfa32197770a8dfbca760/src/components/PlayerView/index.tsx#L87-L93
 2. Add a `useEffect` block dispatching a new `setScalingMode` command at the end of the `PlayerView` function (after and similarly to `setFullscreen`)
     - https://github.com/bitmovin/bitmovin-player-react-native/blob/a72652377fa9bc48ae2bfa32197770a8dfbca760/src/components/PlayerView/index.tsx#L165-L170
 3. Update the Kotlin and Swift code to handle this new message in RNPlayerViewManager (similar to the `setFullscreen` method)
     - call [`view.playerView.setScalingMode`](https://cdn.bitmovin.com/player/android/3/docs/player-ui-web/com.bitmovin.player/-player-view/set-scaling-mode.html) in the `setScalingMode` implementation
     - Kotlin:
        - https://github.com/bitmovin/bitmovin-player-react-native/blob/a72652377fa9bc48ae2bfa32197770a8dfbca760/android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt#L169-L173
         - https://github.com/bitmovin/bitmovin-player-react-native/blob/a72652377fa9bc48ae2bfa32197770a8dfbca760/android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt#L185-L197
     - Swift:
        -  https://github.com/bitmovin/bitmovin-player-react-native/blob/a72652377fa9bc48ae2bfa32197770a8dfbca760/ios/RNPlayerViewManager.m#L65-L68
       - https://github.com/bitmovin/bitmovin-player-react-native/blob/a72652377fa9bc48ae2bfa32197770a8dfbca760/ios/RNPlayerViewManager.swift#L74-L94

jonathanm-tkf:
Good morning, @krocard, thanks for your support, regarding the possibility of Dynamic Scaling Mode, we have updated the PR with the implementation for both platforms, I would like the BM team to review the code and any comments are welcome, this would help us a lot if you can generate a new version to remove the temporary fix that we added in our application and not have the need to destroy the player as we have said before.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
```diff
@@ -196,6 +203,22 @@ class RNPlayerViewManager(private val context: ReactApplicationContext) : Simple
...
+            when (scalingMode) {
+                "Zoom" -> {
+                    view.playerView?.scalingMode = ScalingMode.Zoom
+                }
+                "Stretch" -> {
+                    view.playerView?.scalingMode = ScalingMode.Stretch
+                }
+                else -> {
+                    view.playerView?.scalingMode = ScalingMode.Fit
+                }
```
krocard, line 217: Consider using `Enum.valueOf`:
```suggestion
            view.playerView?.scalingMode = ScalingMode.valueOf(scalingMode)
```

krocard (commented):
Looks good! I'll let @rolandkakonyi review the Swift side.

ios/RNPlayerViewManager.swift
```diff
@@ -100,6 +100,27 @@ class RNPlayerViewManager: RCTViewManager {
...
+             guard let playerView = view.playerView else {
+                 return
+             }
+             if scalingMode == "Zoom" {
+                 playerView.scalingMode = ScalingMode.zoom
+             } else if scalingMode == "Stretch" {
+                 playerView.scalingMode = ScalingMode.stretch
+             } else {
+                 playerView.scalingMode = ScalingMode.fit
+             }
```
rolandkakonyi, line 120: You could use `switch` for this to simplify this:
```suggestion
             switch scalingMode {
             case "Zoom":
                 playerView.scalingMode = .zoom
             case "Stretch":
                 playerView.scalingMode = .stretch
             case "Fit":
                 playerView.scalingMode = .fit
             default:
                 break
             }
```

src/components/PlayerView/index.tsx
```diff
@@ -100,6 +105,7 @@ export function PlayerView({
   fullscreenHandler,
   customMessageHandler,
   isFullscreenRequested = false,
+  scalingMode = ScalingMode.Fit,
```
rolandkakonyi, line 108: Please remove the default value here.
Default values should be maintained in the native Player SDKs.

```suggestion
  scalingMode,
```

src/components/PlayerView/index.tsx
```diff
@@ -62,6 +63,10 @@ export interface PlayerViewProps extends BasePlayerViewProps, PlayerViewEvents {
    * To use this property, a `FullscreenHandler` must be set.
    */
   isFullscreenRequested?: Boolean;
+  /**
+   * The `Scaling Mode` that is used by the `PlayerView` to control the scaling mode.
```
rolandkakonyi, line 67: Please use the documentation similar to the iOS counterpart: https://cdn.bitmovin.com/player/ios/3/docs/Protocols/UserInterfaceApi.html#/c:@M@BitmovinPlayerCore@objc(pl)BMPUserInterfaceApi(py)scalingMode

```suggestion
   * A value defining how the video is displayed within the parent container's bounds.
   * Possible values are defined in `ScalingMode`.
```

rolandkakonyi (commented):
Nice job @jonathanm-tkf!

Please see my comments.

jonathanm-tkf:
Thank you very much for the recommendations @krocard @rolandkakonyi 

I have confirmed it locally and it works perfectly, again thank you very much for the help.

rolandkakonyi (approved):

krocard (approved):

PR #283, opened 2023-10-10
Fix API docs problems

rolandkakonyi:
## Description

There are warnings when generating docs using `yarn docs`.

## Changes

1. Turned on "treatWarningsAsErrors" to enforce writing documentation
2. Fixed `SideLoadedSubtitleTrack` API to provide without `MakeRequired` hack
3. Fixed various warnings

## Checklist


src/offline/offlineContentManager.ts
```diff
@@ -39,6 +39,35 @@ interface NativeOfflineModule extends NativeModule {
...
+    } else if (data.eventType === OfflineEventType.onSuspended) {
+      listener.onSuspended?.(data);
+    } else if (data.eventType === OfflineEventType.onResumed) {
+      listener.onResumed?.(data);
+    } else if (data.eventType === OfflineEventType.onCanceled) {
+      listener.onCanceled?.(data);
+    }
+  });
+};
+
```
rolandkakonyi, line 70: This was just moved to avoid exporting it.

src/subtitleTrack.ts
rolandkakonyi, line 1: Got rid of `MakeRequired` as we don't need it at all.

src/subtitleTrack.ts
```diff
@@ -48,13 +46,40 @@ export interface SubtitleTrack {
...
  *
- * Since `SubtitleTrack` has all of its properties as optionals for total compatibility with
- * values that may be sent from native code, this type serves as a reinforcer of what properties
- * should be required during the registration of an external subtitle track from JS.
  */
-export type SideLoadedSubtitleTrack = MakeRequired<
-  SubtitleTrack,
-  'url' | 'label' | 'language' | 'format'
->;
+export interface SideLoadedSubtitleTrack {
```
zigavehovec, line 52: I'm not a very big fan of documentation and definition duplication that we have here now between SubtitleTrack and SideLoadedSubtitleTrack. 
Would be good if we could achieve some other way of making some fields not optional. I'm no TS expert but I just tried to extend the SubtitleTack interface and it looks like it achieves what we want here:

image

On the calling side I now get an error if I don't specify a non optional property: 
image

What do you think about this approach?

zigavehovec (commented):
Nice improvements 👍 
I added one suggestion, let me know what you think about it 🙂

src/subtitleTrack.ts
rolandkakonyi: Thanks, this is a great idea!
Implemented in c73c55d

Note: looks like the documentation gets picked up by the doc gen tool and the IDE as well, so we don't have to duplicate it.

zigavehovec (approved):
Nice 👍

PR #284, opened 2023-10-10
Add missing API docs

rolandkakonyi:
## Description

We miss documentation for some of our public API.

## Changes

Added missing documentation.

## Checklist

### DIFF
--- a/src/advertising.ts
+++ b/src/advertising.ts
@@ -25,7 +25,7 @@ export enum AdSourceType {
    */
   IMA = 'ima',
   /**
-   * Unknown advertising source type.
+   * Unknown ad source type.
    */
   UNKNOWN = 'unknown',
   /**
--- a/src/advertising.ts
+++ b/src/advertising.ts
@@ -2,8 +2,17 @@
  * Quartiles that can be reached during an ad playback.
  */
 export enum AdQuartile {
+  /**
+   * Fist ad quartile.
+   */
   FIRST = 'first',
+  /**
+   * Mid ad quartile.
+   */
   MID_POINT = 'mid_point',
+  /**
+   * Third ad quartile.
+   */
   THIRD = 'third',
 }
 
@@ -11,8 +20,17 @@ export enum AdQuartile {
  * The possible types an `AdSource` can be.
  */
 export enum AdSourceType {
+  /**
+   * Google Interactive Media Ads.
+   */
   IMA = 'ima',
+  /**
+   * Unknown advertising source type.
+   */
   UNKNOWN = 'unknown',
+  /**
+   * Progressive ad type.
+   */
   PROGRESSIVE = 'progressive',
 }
 
--- a/src/components/PlayerView/events.ts
+++ b/src/components/PlayerView/events.ts
@@ -171,7 +171,7 @@ interface EventProps {
    */
   onDestroy: DestroyEvent;
   /**
-   * Called for each occurring player event.
+   * All events emitted by the player.
    */
   onEvent: Event;
   /**
--- a/src/components/PlayerView/events.ts
+++ b/src/components/PlayerView/events.ts
@@ -291,12 +291,12 @@ interface EventProps {
   onSourceError: SourceErrorEvent;
   /**
    * Event emitted when a new source loading has started.
-   * It doesn't mean that the loading of the new manifest has finished.
    */
   onSourceLoad: SourceLoadEvent;
   /**
    * Event emitted when a new source is loaded.
-   * It doesn't mean that the loading of the new manifest has finished.
+   * This does not mean that the source is immediately ready for playback.
+   * `ReadyEvent` indicates the player is ready for immediate playback.
    */
   onSourceLoaded: SourceLoadedEvent;
   /**
--- a/src/components/PlayerView/events.ts
+++ b/src/components/PlayerView/events.ts
@@ -67,65 +67,281 @@ import {
  * Used to generate the specific events interface for each component.
  */
 interface EventProps {
+  /**
+   * Event emitted when an ad break has finished.
+   */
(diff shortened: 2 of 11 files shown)


src/advertising.ts
```diff
@@ -2,17 +2,35 @@
...
+   */
   FIRST = 'first',
+  /**
+   * Mid ad quartile.
+   */
   MID_POINT = 'mid_point',
+  /**
+   * Third ad quartile.
+   */
   THIRD = 'third',
```
strangesource, line 16: NIT: Not sure how much value this adds.

src/advertising.ts
```diff
@@ -2,17 +2,35 @@
...
+   * Google Interactive Media Ads.
+   */
   IMA = 'ima',
+  /**
+   * Unknown advertising source type.
+   */
   UNKNOWN = 'unknown',
+  /**
+   * Progressive ad type.
+   */
```
strangesource, line 33: I would prefer a consistent naming: `ad type` vs. `advertising source type`.

src/components/PlayerView/events.ts
```diff
@@ -67,65 +67,281 @@ import {
...
+   * @platform iOS, Android
+   */
   onCastWaitingForDevice: CastWaitingForDeviceEvent;
+  /**
+   * Event emitted when the player is destroyed.
+   */
   onDestroy: DestroyEvent;
+  /**
+   * Called for each occurring player event.
+   */
```
strangesource, line 175: I would prefer a more consistent naming. The other functions are documented with 'event emitted', this one it's 'called' but functionality wise there is no major difference.

src/components/PlayerView/events.ts
```diff
@@ -67,65 +67,281 @@ import {
...
+   * Event emitted when the player ends stalling, due to enough data in the buffer.
+   */
   onStallEnded: StallEndedEvent;
+  /**
+   * Event emitted when a source error occurred.
+   */
   onSourceError: SourceErrorEvent;
+  /**
+   * Event emitted when a new source loading has started.
+   * It doesn't mean that the loading of the new manifest has finished.
```
strangesource: What's the reason for this additional comment? That's not something I would assume when looking at the event.

src/components/PlayerView/index.tsx
```diff
@@ -16,11 +16,23 @@ import { FullscreenHandlerBridge } from '../../ui/fullscreenhandlerbridge';
...
  * @see NativePlayerView
  */
 export interface BasePlayerViewProps {
+  /**
+   * The `FullscreenHandler` that is used by the `PlayerView` to control the fullscreen mode.
+   */
+  fullscreenHandler?: FullscreenHandler;
+
+  /**
+   * The `CustomMessageHandler` that can be used to directly communicate with the embedded WebUi.
```
strangesource, line 30:
```suggestion
   * The `CustomMessageHandler` that can be used to directly communicate with the embedded Bitmovin Web UI.
```

src/events.ts
```diff
@@ -141,25 +141,33 @@ export interface EventSource {
   metadata?: Record<string, any>;
 }
 
+/**
+ * Represents seeking position.
+ */
```
strangesource, line 146: I don't think this offers any additional value. If you want to keep it anyways, I would prefer the following:
```suggestion
/**
 * Represents a seeking position.
 */
```

src/events.ts
```diff
@@ -141,25 +141,33 @@ export interface EventSource {
...
+  source: EventSource;
+  /**
+   * The position within the `source` in seconds.
+   */
+  time: number;
+}
+
 /**
  * Emitted when the player is about to seek to a new position.
  * Only applies to VoD streams.
```
strangesource, line 160: NIT: we could cross reference `timeShift` here for live streams.

src/ui/custommessagehandler.ts
```diff
@@ -51,13 +51,26 @@ export class CustomMessageHandler {
...
+   */
   receivedSynchronousMessage(
     message: string,
     data: string | undefined
   ): string | undefined {
     return this.onReceivedSynchronousMessage(message, data);
   }
 
+  /**
+   * Gets called when a asynchronous message was received from the Bitmovin Web UI.
```
strangesource, line 69:
```suggestion
   * Gets called when an asynchronous message was received from the Bitmovin Web UI.
```

strangesource (commented):
Nice improvements. Just some minor comments from my side.

src/advertising.ts
rolandkakonyi: 🤦 copied from the iOS SDK 🙈 

@strangesource which one would you prefer?

src/components/PlayerView/events.ts
rolandkakonyi: Fixed in 736513c

src/events.ts
rolandkakonyi: What do you mean?

src/advertising.ts
strangesource: `advertising source type`, but no strong opinion. :)

strangesource (approved):
Approving as only nit picks are left.

src/advertising.ts
rolandkakonyi: I went for `ad source type` because we had no other occurrence of `advertising` but `ad` was used everywhere in current API documentation. So consistency requires us to use `ad`.

See: 8507fa0

src/events.ts
rolandkakonyi: That's a great suggestion. Please see 5681080.
By using `@link` it will have them linked in the description for easy jumping.
See it in action here: https://cdn.bitmovin.com/player/reactnative/0.12.0/docs/interfaces/SeekEvent.html

PR #285, opened 2023-10-10
Validate API docs on CI

rolandkakonyi:
## Description

There is no way to catch if API documentation is missing.

## Changes

Extended our CI workflow to validate our API documentation generation.

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,6 +60,29 @@ jobs:
       - name: Compile TypeScript
         run: yarn typescript
 
+  test-build-docs:
+    name: Build API docs
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup node and npm registry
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -60,6 +60,29 @@ jobs:
...
+          cache: 'yarn'
+
+      - name: Install node_modules
+        run: yarn install --frozen-lockfile
+
+      - name: Install node_modules (example/)
+        run: yarn install --frozen-lockfile --cwd example
+
+      - name: Build docs
+        run: yarn docs
```
matamegger, line 84: But the out put is not used anywhere?

What's the intention? I think I am missing something.

rolandkakonyi: This command will fail if there is public API that is not documented.

matamegger: Ah, now I get you. I was interpreting this differently when I was reading it yesterday 😅 👍

matamegger (approved):

PR #286, opened 2023-10-11
Introduce `PictureInPictureConfig`

rolandkakonyi:
## Description

`PlaybackConfig.isPictureInPictureEnabled` flag is deprecated but the replacement is not available in the React Native Player SDK.

## Changes

Introduced `PictureInPictureConfig` on `<PlayerView />` with `isEnabled` flag.

## Checklist
- [x] 🗒 `CHANGELOG` entry


123mpozzi (approved):

123mpozzi:
Works fine on my test device

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
```diff
@@ -174,6 +176,11 @@ class RNPlayerViewManager(private val context: ReactApplicationContext) : Simple
         }
     }
 
+    @ReactProp(name = "pictureInPictureConfig")
```
zigavehovec, line 179: This prop pattern is new in our project right? 
It looks like it replaces the whole dispatch() logic that we use for E.g. `setCustomMessageHandlerBridgeId` am I right?

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -107,6 +109,8 @@ class RNPlayerView(val context: ReactApplicationContext) :
         getViewTreeObserver().addOnGlobalLayoutListener { requestLayout() }
     }
 
+    var pictureInPictureConfig: PictureInPictureConfig? = null
```
zigavehovec, line 112: I think we should move this to where `pictureInPictureHandler` lives.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -98,6 +98,8 @@ class RNPlayerView(val context: ReactApplicationContext) :
     View.OnLayoutChangeListener,
     RNPictureInPictureDelegate {
 
+    data class PictureInPictureConfig(val isEnabled: Boolean)
```
zigavehovec, line 101: This should maybe be moved to the RNPictureInPictureHandler.kt file.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
```diff
@@ -43,7 +45,7 @@ class RNPlayerViewManager(private val context: ReactApplicationContext) : Simple
      * modules in case a full-custom implementation is needed. A default implementation is provided
      * out-of-the-box.
      */
-    var pictureInPictureHandler = RNPictureInPictureHandler(context)
+    private var pictureInPictureHandler = RNPictureInPictureHandler(context)
```
zigavehovec, line 48: Do you know why we have one instance of pictureInPictureHandler on the `RNPlayerViewManager` and one on the `RNPlayerView`? This feels like an oversight to me and i think this one should be removed probably.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
```diff
@@ -217,13 +224,18 @@ class RNPlayerViewManager(private val context: ReactApplicationContext) : Simple
...
+                    playerConfig
+                        ?.getMap("playbackConfig")
+                        ?.getBooleanOrNull("isPictureInPictureEnabled")
+                        ?: false,
+                )
+
+            isPictureInPictureEnabled.let {
+                pictureInPictureHandler.isPictureInPictureEnabled = it
+                view.pictureInPictureHandler = pictureInPictureHandler
+            }
```
zigavehovec, line 238: 1.) The `or()` function is discouraged not only because of style but also because both sides of `or` will always be evaluated as opposed to E.g. `somethingTrue()` || `somethingFalse()` that will only evaluate the left side if true.
2.) As mentioned above I think we can and should remove the pictureInPictureHandler in this file
3.) `isPictureInPictureEnabled.let` only makes the code confusing in this case

```suggestion
            val playbackConfig = playerConfig?.getMap("playbackConfig")
            val isPictureInPictureEnabled = view.pictureInPictureConfig?.isEnabled == true ||
                    playbackConfig?.getBooleanOrNull("isPictureInPictureEnabled") == true
            view.pictureInPictureHandler?.isPictureInPictureEnabled = isPictureInPictureEnabled
```

zigavehovec (commented):
I added some comments on the Android side, looks good otherwise 👍

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
rolandkakonyi: Yes. This is different.
We use `dispatch` to perform actions - a lot of times with talking to other modules - while `ReactProp` is for passing a raw value to the underlying native view.
In this case it feels more natural to use it this way to not bloat the `'attachPlayer'` dispatch call with an additional config.

Bubbling events are also directly mapped to react props.
We even have this documentation: https://github.com/bitmovin/bitmovin-player-react-native/blob/628229bed59647f76b20e8e0c2129135eea5fb00/android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt#L130-L131

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
rolandkakonyi: fixed in 90b13a1

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
rolandkakonyi: fixed in 90b13a1

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
rolandkakonyi: Good catch!
I removed the stored instance in `RNPlayerViewManager` and reusing the existing or creating a new one in 7872455

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
rolandkakonyi: I literally expected feedback on this one (maybe should have even explicitly ask for it) as I'm no expert of Kotlin code style.

Thank you for the excellent and simple suggestion!
Implemented in 0b2460f

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
zigavehovec: Thanks for the explanation!

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
rolandkakonyi: Thanks for catching!
It looks like my former refactoring in 7872455ce1c9b67927d992e70dda6c5765af3a7a and the suggested changes didn't play well together.

Fixed in ae24ef2

zigavehovec (approved):
Nice 👌

PR #287, opened 2023-10-12
`onPictureInPictureAvailabilityChanged` event support on iOS and tvOS

rolandkakonyi:
## Description

We don't support `onPictureInPictureAvailabilityChanged` on iOS and this way it is not possible to tell if PiP is available.

## Changes

Emitting `onPictureInPictureAvailabilityChanged` on iOS and tvOS to allow detection of PiP availability.

## Checklist
- [x] 🗒 `CHANGELOG` entry

### DIFF
--- a/ios/RNPlayerViewManager.swift
+++ b/ios/RNPlayerViewManager.swift
@@ -40,7 +40,7 @@ class RNPlayerViewManager: RCTViewManager {
             }
 #endif
 
-            var previousPictureInPictureAvailableValue = false
+            let previousPictureInPictureAvailableValue: Bool
             if let playerView = view.playerView {
                 playerView.player = player
                 previousPictureInPictureAvailableValue = playerView.isPictureInPictureAvailable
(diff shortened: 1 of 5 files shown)


ios/RNPlayerViewManager.swift
```diff
@@ -109,4 +114,18 @@ class RNPlayerViewManager: RCTViewManager {
...
+            return
+        }
+        let event: [AnyHashable: Any] = [
+            "isPictureInPictureAvailable": playerView.isPictureInPictureAvailable,
+            "name": "onPictureInPictureAvailabilityChanged",
+            "timestamp": Date().timeIntervalSince1970
+        ]
+        view.onPictureInPictureAvailabilityChanged?(event)
+    }
+
```
123mpozzi, line 130: nit: I would remove this extra unnecessary line please

123mpozzi (approved):

PR #302, opened 2023-10-19
Move `PictureInPictureConfig` to `PlayerViewConfig`

rolandkakonyi:
## Description

With the introduction of `PlayerViewConfig` it makes sense to move `PictureInPictureConfig` into it, as it is on the iOS SDK for example [here](https://cdn.bitmovin.com/player/ios/3/docs/Classes/PlayerViewConfig.html#/c:@M@BitmovinPlayerCore@objc(cs)BMPPlayerViewConfig(py)pictureInPictureConfig).

## Changes

Adapted parsing with introduction of internal types.

!!!IMPORTANT!!!
This is not an API breaking change because the `PictureInPictureConfig` was not yet released!

+1 made `PlayerViewConfig.uiConfig` optional on the TypeScript side to allow simpler configuration if one wants to only set `pictureInPictureConfig`.
+2 removing padding on Android when entering PiP to not show borders on the PiP view.

## Checklist

### DIFF
--- a/android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
@@ -99,6 +99,10 @@ class RNPlayerView(val context: ReactApplicationContext) :
     View.OnLayoutChangeListener,
     RNPictureInPictureDelegate {
 
+    data class RNPlayerViewConfig(
+        val playerViewConfig: PlayerViewConfig?,
+        val pictureInPictureConfig: RNPictureInPictureHandler.PictureInPictureConfig?,
+    )
     init {
         // React Native has a bug that dynamically added views sometimes aren't laid out again properly.
         // Since we dynamically add and remove SurfaceView under the hood this caused the player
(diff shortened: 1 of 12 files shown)


example/src/screens/BasicPictureInPicture.tsx
```diff
@@ -115,12 +117,20 @@ export default function BasicPictureInPicture({
...
-    <SafeAreaView edges={['bottom', 'left', 'right']} style={styles.container}>
+    <SafeAreaView
+      edges={['bottom', 'left', 'right']}
+      style={
+        // On Android, we need to remove the padding from the container when in PiP mode.
+        Platform.OS === 'android' && isInPictureInPicture
+          ? [styles.container, { padding: 0 }]
+          : styles.container
+      }
+    >
```
rolandkakonyi, line 128: This is a sneak to remove padding when in PiP view as it shows up on Android

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -98,6 +98,10 @@ class RNPlayerView(val context: ReactApplicationContext) :
     LifecycleEventListener,
     View.OnLayoutChangeListener,
     RNPictureInPictureDelegate {
+    data class RNPlayerViewConfig(
+        val playerViewConfig: PlayerViewConfig?,
+        val pictureInPictureConfig: RNPictureInPictureHandler.PictureInPictureConfig?,
+    )
```
strangesource, line 104: NIT: I would prefer to not have that as a nested class in the RNPlayerView. I don't think we have this particular situation yet, I would suggest o move it to a separate file adjacent to the `RNPlayerView` or to the bottom on `RNPlayerView.kt`.

Would also be nice to have an explanation around why we have this wrapper. Naming wise I would prefer `RNPlayerViewConfigWrapper`.

strangesource (commented):
Nice improvements! 
Just one minor remark around the Android native part.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
rolandkakonyi: fixed in e233e73

PR #305, opened 2023-10-20
Improve start release workflow

rolandkakonyi:
## Description

1. We don't utilize a CocoaPods cache, which makes the release process slower than needed
2. One might miss the notification about a new PR without explicitly assigning a reviewer for the release PR
3. There is no easy way to test the release workflow

## Changes

1. Added the same cache solution we use for `ci.yml` workflow
2. Assigning the user triggering the release process explicitly as reviewer on the release PR
3. Added new option for `dry_run`, to enable testing this workflow

Successful run (with `dry_run` enabled): https://github.com/bitmovin/bitmovin-player-react-native/actions/runs/6588544853

Runtime went from [6 minutes](https://github.com/bitmovin/bitmovin-player-react-native/actions/runs/6587925921) to [3.5 minutes](https://github.com/bitmovin/bitmovin-player-react-native/actions/runs/6588544853).

## Checklist


zigavehovec (approved):
Looks good!

PR #308, opened 2023-10-23
Implement `LiveConfig`: `minTimeshiftBufferDepth`

123mpozzi:
## Description

Add `LiveConfig` and its `minTimeshiftBufferDepth` property to control the minimum buffer depth of a stream needed to enable time shifting.

## Changes

- Add `LiveConfig`
- Add `minTimeshiftBufferDepth` property to `LiveConfig`

## Testing

Apply this patch, open **Basic Playback**, and look at the logs on Metro
[patch.diff.txt](https://github.com/bitmovin/bitmovin-player-react-native/files/13072168/patch.diff.txt)

Checks:
- `minTimeshiftBufferDepth` should be set to `-11.2`

## Checklist
- [x] 🗒 `CHANGELOG` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [Unreleased]
+
+### Added
+
+- `LiveConfig.minTimeshiftBufferDepth` to control the minimum buffer depth of a stream needed to enable time shifting.
+
 ## [0.13.0] (2023-10-20)
 
 ### Added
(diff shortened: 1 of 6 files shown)


rolandkakonyi (approved):
The changes look good! Nice job!

The testing patch isn't really showcasing the real feature, as you need a live stream to test the config but I also understand that we should keep the test live streams out of this repo.

matamegger (approved):
Well done!

PR #312, opened 2023-11-03
Introduce Playback speed related API and events

123mpozzi:
## Description

Implement **playback speed** related API and events.

## Changes

- Add getter and setter for `playbackSpeed` property of the Player
- Add `PlaybackSpeedChangedEvent` (iOS and tvOS only)
- Add `canPlayAtPlaybackSpeed` (iOS and tvOS only)

## Testing
Apply the following patch:
[playbackSpeed.patch](https://github.com/bitmovin/bitmovin-player-react-native/files/13297889/playbackSpeed.patch)

Check for:
- `getPlaybackSpeed`: `playbackSpeed` value starts at `1` (default value) after pressing Play, and correctly updates its value on UI triggers. Current value is logged on _onPlay_ triggers
- `setPlaybackSpeed`: _on seeking_, `playbackSpeed` should multiply by `2` on Android, by `-2` on iOS.
- iOS/tvOS: `PlaybackSpeedChangedEvent` being emitted correctly on UI triggers
- iOS/tvOS: `canPlayAtPlaybackSpeed` working correctly. Sample values are logged on _onPlay_ triggers

## Checklist
- [x] 🗒 `CHANGELOG` entry


android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt
```diff
@@ -543,6 +543,30 @@ class PlayerModule(private val context: ReactApplicationContext) : ReactContextB
...
+     */
+    @ReactMethod
+    fun getPlaybackSpeed(nativeId: NativeId, promise: Promise) {
+        uiManager()?.addUIBlock {
+            promise.resolve(players[nativeId]?.playbackSpeed)
+        }
+    }
+
+    /**
+     * Call `.setPlaybackSpeed(playbackSpeed:)` on `nativeId`'s player.
```
rolandkakonyi, line 559: I think this line is not really explaining what the method does.

What do you think about this?
```suggestion
     * Sets playback speed for the player.
```

ios/RNPlayerView.swift
```diff
@@ -22,6 +22,7 @@ public class RNPlayerView: UIView {
     @objc var onStallStarted: RCTBubblingEventBlock?
     @objc var onStallEnded: RCTBubblingEventBlock?
     @objc var onTimeChanged: RCTBubblingEventBlock?
+    @objc var onPlaybackSpeedChanged: RCTBubblingEventBlock?
```
rolandkakonyi, line 25: nit: Since we don't follow any strict order here, it would make sense to move it as the last event-related property, right before `@objc var config: [String: Any]?`

ios/RNPlayerViewManager.m
```diff
@@ -21,6 +21,7 @@ @interface RCT_EXTERN_REMAP_MODULE(NativePlayerView, RNPlayerViewManager, RCTVie
 RCT_EXPORT_VIEW_PROPERTY(onStallStarted, RCTBubblingEventBlock)
 RCT_EXPORT_VIEW_PROPERTY(onStallEnded, RCTBubblingEventBlock)
 RCT_EXPORT_VIEW_PROPERTY(onTimeChanged, RCTBubblingEventBlock)
+RCT_EXPORT_VIEW_PROPERTY(onPlaybackSpeedChanged, RCTBubblingEventBlock)
```
rolandkakonyi, line 24: same here, why this location?

src/player.ts
```diff
@@ -405,4 +405,41 @@ export class Player extends NativeInstance<PlayerConfig> {
...
+   * Negative values are ignored during Casting.
+   * @note
+   * During reverse playback the playback will continue until the beginning of the active source is
+   * reached. When reaching the beginning of the source, playback will be paused and the playback
+   * speed will be reset to its default value of `1`. No {@link PlaybackFinishedEvent} will be
+   * emitted in this case.
+   *
+   * @param playbackSpeed - The playback speed level to set.
+   */
+  setPlaybackSpeed = (playbackSpeed = 1) => {
```
rolandkakonyi, line 425: What's the idea behind using default value here?
I would remove it as it would be confusing to call this method without passing a parameter.

```suggestion
  setPlaybackSpeed = (playbackSpeed: number) => {
```

src/player.ts
```diff
@@ -405,4 +405,41 @@ export class Player extends NativeInstance<PlayerConfig> {
...
+
+  /**
+   * Checks the possibility to play the media at specified playback speed.
+   * @param playbackSpeed The playback speed to check.
+   * @returns `true` if it's possible to play the media at the specified playback speed, otherwise `false`.
+   * @platform iOS, tvOS
+   */
+  canPlayAtPlaybackSpeed = async (playbackSpeed: number): Promise<boolean> => {
+    return PlayerModule.canPlayAtPlaybackSpeed(this.nativeId, playbackSpeed);
+  };
```
rolandkakonyi, line 444: This is not implemented for Android, so it causes a runtime error.

Please add an early return when calling this on Android.

The value of the returns should be TBD. I picked `undefined` as we don't know if it is `true` or `false`.

@strangesource what do you think about this?

```suggestion
  canPlayAtPlaybackSpeed = async (
    playbackSpeed: number
  ): Promise<boolean | undefined> => {
    if (Platform.OS === 'android') {
      console.warn(
        `[Player ${this.nativeId}] Method canPlayAtPlaybackSpeed is not available for Android. Only iOS and tvOS devices.`
      );
      return undefined;
    }
    return PlayerModule.canPlayAtPlaybackSpeed(this.nativeId, playbackSpeed);
  };
```

rolandkakonyi (changes requested):

android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt
123mpozzi: Nice idea, I like it

ios/RNPlayerViewManager.m
123mpozzi: address in 1fa91bb913b00693336059784af698ad64254879

ios/RNPlayerView.swift
123mpozzi: address in 1fa91bb913b00693336059784af698ad64254879

src/player.ts
123mpozzi: address in c3a77e408de1dc56ebdeb849f826256ae43a0eb4

src/player.ts
```diff
@@ -405,4 +405,41 @@ export class Player extends NativeInstance<PlayerConfig> {
   castStop = () => {
     PlayerModule.castStop(this.nativeId);
   };
+
+  /**
+   * Sets the playback speed of the player. Fast forward, slow motion and reverse playback are supported.
+   * @note
+   * - Slow motion is used by values between `0` and `1`.
```
strangesource, line 412:
```suggestion
   * - Slow motion is indicated by values between `0` and `1`.
```

src/player.ts
```diff
@@ -405,4 +405,41 @@ export class Player extends NativeInstance<PlayerConfig> {
...
+   * - Fast forward by values greater than `2`.
+   * - Slow reverse is used by values between `0` and `-1`, and fast reverse is used by values less than `-1`. iOS and tvOS only.
+   * @note
+   * Negative values are ignored during Casting.
```
strangesource, line 416:
```suggestion
   * Negative values are ignored during Casting and on Android.
```
I assume.

src/player.ts
```diff
@@ -405,4 +405,41 @@ export class Player extends NativeInstance<PlayerConfig> {
   castStop = () => {
     PlayerModule.castStop(this.nativeId);
   };
+
+  /**
+   * Sets the playback speed of the player. Fast forward, slow motion and reverse playback are supported.
+   * @note
+   * - Slow motion is used by values between `0` and `1`.
+   * - Fast forward by values greater than `2`.
```
strangesource, line 413: What happens between 1 and 2?  on Android, all values above 1 are considered fast forward. But I am not sure if there is some misalignment as to what fast forward means. 🤔

src/player.ts
```diff
@@ -405,4 +405,41 @@ export class Player extends NativeInstance<PlayerConfig> {
...
+   * - Slow reverse is used by values between `0` and `-1`, and fast reverse is used by values less than `-1`. iOS and tvOS only.
+   * @note
+   * Negative values are ignored during Casting.
+   * @note
+   * During reverse playback the playback will continue until the beginning of the active source is
+   * reached. When reaching the beginning of the source, playback will be paused and the playback
+   * speed will be reset to its default value of `1`. No {@link PlaybackFinishedEvent} will be
+   * emitted in this case.
+   *
+   * @param playbackSpeed - The playback speed level to set.
```
strangesource, line 423:
```suggestion
   * @param playbackSpeed - The playback speed to set.
```
Not a level, is it?

src/player.ts
```diff
@@ -405,4 +405,41 @@ export class Player extends NativeInstance<PlayerConfig> {
...
+   */
+  setPlaybackSpeed = (playbackSpeed: number) => {
+    PlayerModule.setPlaybackSpeed(this.nativeId, playbackSpeed);
+  };
+
+  /**
+   * @returns The player's current playback speed.
```
strangesource, line 430: Should we link to `setPlaybackSpeed` for details?

src/player.ts
strangesource: Undefined makes sense IMO. 👍 Let's make sure we have it as part of the documentation.

123mpozzi: Address in cb36609afc66508d809b0f9c1aee30054a49b43e  
Update documentation in 0c7129df518386d0c55dc614f1986a022b3ecf83

src/player.ts
123mpozzi: That's correct, nice catch!

src/player.ts
123mpozzi: Yes, correct! Thank you!

src/player.ts
123mpozzi: Nice idea!
Address in fc50f75c65a9d3d27c6f7b935134b861538355ef

src/player.ts
123mpozzi: Yes, it make sense to change it to `1`. For some reason it was `2` on iOS, but we decided to go for `1` there too.  
Address in 16c8becb9c43d19f6943687349db24730862c79b

rolandkakonyi (approved):

PR #313, opened 2023-11-03
Android: Playback doesn't pause when app goes to background

zigavehovec:
## Description (PRN-81)

Player doesn't pause as intended when app goes to the background. This was due to `PlayerView.onStop` not being called.

Unfortunately the React Native's LifecycleEventListener doesn't have `onHostStop` and `onHostStart` lifecycle methods, which we need so I migrated to Activity lifecycle observer.

## Changes

- Replace `LifecycleEventListener` with `DefaultLifecycleObserver`
- Remove the unneeded PiP handling
- dispose function now also calls `playerView.onDestroy`

## Checklist
- [x] 🗒 `CHANGELOG` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@
 - `Player.buffer` to control buffer preferences and to query the current buffer state
 - `DownloadFinishedEvent` to signal when the download of specific content has finished
 
+### Fixed
+
+- Android: Playback doesn't pause when app goes to background
+
 ## [0.13.0] (2023-10-20)
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 ### Fixed
 
 - Android: Playback doesn't pause when app goes to background
+- Android: `PlayerView.onDestroy` not being called when the view is detached from the view hierarchy
 
 ## [0.13.0] (2023-10-20)
 
--- a/android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
@@ -3,7 +3,6 @@ package com.bitmovin.player.reactnative
 import android.annotation.SuppressLint
 import android.content.res.Configuration
 import android.graphics.Rect
-import android.util.Log
 import android.view.View
 import android.view.ViewGroup
 import android.widget.LinearLayout
--- a/android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
@@ -3,9 +3,12 @@ package com.bitmovin.player.reactnative
 import android.annotation.SuppressLint
 import android.content.res.Configuration
 import android.graphics.Rect
+import android.util.Log
 import android.view.View
 import android.view.ViewGroup
 import android.widget.LinearLayout
+import androidx.lifecycle.DefaultLifecycleObserver
+import androidx.lifecycle.LifecycleOwner
 import com.bitmovin.player.PlayerView
 import com.bitmovin.player.api.Player
 import com.bitmovin.player.api.event.Event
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,10 @@
 - `Player.buffer` to control buffer preferences and to query the current buffer state
 - `DownloadFinishedEvent` to signal when the download of specific content has finished
 
+### Fixed
+
+- Android: Playback doesn't pause when app goes to background
```
strangesource, line 13: Let's also add a changelog for the destroy function now being called so someone relying on the current behaviour (which I don't suspect but let's rather be safe than sorry) can see that there were changes.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -15,9 +17,8 @@ import com.bitmovin.player.api.ui.PlayerViewConfig
 import com.bitmovin.player.reactnative.converter.JsonConverter
 import com.bitmovin.player.reactnative.ui.RNPictureInPictureDelegate
 import com.bitmovin.player.reactnative.ui.RNPictureInPictureHandler
-import com.facebook.react.bridge.LifecycleEventListener
-import com.facebook.react.bridge.ReactApplicationContext
-import com.facebook.react.bridge.ReactContext
+import com.facebook.react.ReactActivity
+import com.facebook.react.bridge.*
```
strangesource, line 21: I am surprised or linitng doesn't flag * imports. 🤔  Maybe something to consider for the future. Fine for me for now as linter configuration is out of scope for the issue.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -94,11 +95,30 @@ private val EVENT_CLASS_TO_REACT_NATIVE_NAME_MAPPING_UI = mapOf<KClass<out Event
...
 @SuppressLint("ViewConstructor")
-class RNPlayerView(val context: ReactApplicationContext) :
-    LinearLayout(context),
-    LifecycleEventListener,
-    View.OnLayoutChangeListener,
-    RNPictureInPictureDelegate {
+class RNPlayerView(
+    private val context: ReactApplicationContext,
+) : LinearLayout(context), View.OnLayoutChangeListener, RNPictureInPictureDelegate {
+    private val activityLifecycle = (context.currentActivity as ReactActivity).lifecycle
```
strangesource, line 101: Is this safe to call? `currentActivity` is nullable and of type `Activity`.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
zigavehovec: We have a linting rule that allows wildcard imports for react bridge because in a lot of files there would be really a lot react bridge imports.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
zigavehovec: In my opinion yes.
- `currentActivity` is nullable because it's a function in application context scope and it's a legit use case to have no activity in some scenarios in that scope. But we are attaching a view here, so if there is no activity attached and we are creating a PlayerView we are probably doing something wrong.
- Regarding activity type a ReactActivity is needed to create RN root view, so this should also be safe. The actual cast that we need here is `AppCompatActivity`, but since RectActivity extends `AppCompatActivity` I thought this was somewhat cleaner.

What we could do is to handle it in a more descriptive way or even go very safe and instead of throwing we could just log errors - in that case something will likely be broken though. I think I prefer throwing, since I don't see a legit use case where the casting wouldn't work but maybe I am missing something of course.

Can you think of a use case where this casting wouldn't work?

CHANGELOG.md
zigavehovec: https://github.com/bitmovin/bitmovin-player-react-native/pull/313/commits/1c2243613d54468c0a6fbedbe2a6b5b7e034bb73

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
zigavehovec: We now throw an IllegalStateException with a clear message if the casting fails: https://github.com/bitmovin/bitmovin-player-react-native/pull/313/commits/bdcb1203bf00eb55e1e775ad8ef1402349768cdf

strangesource (approved):
Nice!

matamegger (approved):

PR #316, opened 2023-11-08
Android: Add subtitle-only UI

testcenter:
## Description
The feature to have the UI disabled, but show the subtitle is already available on iOS via the `UserInterfaceType.Subtitle` config.
This flag is ignored on the android implementation.

## Changes
- Add support for `UserInterfaceType.Subtitle` on android
- Change the `RNPlayerView` from `LinearLayout` -> `FrameLayout` to support overlaying the `SubtitleView`
- Add an example to the example app

## Checklist
- [x] 🗒 `CHANGELOG` entry


android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
```diff
@@ -264,11 +268,25 @@ class RNPlayerViewManager(private val context: ReactApplicationContext) : Simple
...
+                ) {
+                    playerView = PlayerView(currentActivity, player, nativePlayerViewConfig)
+                } else {
+                    val overwrittenConfig = nativePlayerViewConfig.copy(uiConfig = UiConfig.Disabled)
+                    playerView = PlayerView(currentActivity, player, overwrittenConfig)
+                    val subtitleView = SubtitleView(currentActivity)
+                    subtitleView.setPlayer(player)
+                    view.addSubtitleView(subtitleView)
+                }
+
```
matamegger, line 289: I can not comment on the according line, but this code is within an else branch where we chack if the `PlayerView` was created before or not.
However, the need for the `SubtitleView` is defined by the `Player` i.e.I think we have to create/destroy the `SubtitleView` in both cases of the if branch.

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
```diff
@@ -1179,19 +1180,31 @@ class JsonConverter {
...
+                else -> UserInterfaceType.Bitmovin
+            }
+
         /**
          * Converts the [json] to a `RNPlayerViewConfig` object.
          */
         fun toRNPlayerViewConfigWrapper(json: ReadableMap) = RNPlayerViewConfigWrapper(
             playerViewConfig = toPlayerViewConfig(json),
             pictureInPictureConfig = toPictureInPictureConfig(json.getMap("pictureInPictureConfig")),
+            userInterfaceType = toUserInterfaceType(json),
```
matamegger, line 1207: I think you are mixing here unrelated configurations 🤔 
The `PlayerViewConfig` is an own type in the `ts` world.

I think this would require a `RNStyleConfigWrapper` (?)

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -203,6 +204,13 @@ class RNPlayerView(
         }
     }
 
+    /**
+     * Make sure this is called after [addPlayerView], otherwise the subtitleView will be overdrawn by the player!
+     */
```
matamegger, line 209: Is that comment true? You are adding the `playerView` on index 0, so there shouldn't be any player in front of the `SubtitleView`. Do I miss something?

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -203,6 +204,13 @@ class RNPlayerView(
...
+    fun addSubtitleView(subtitleView: SubtitleView) {
+        addView(subtitleView)
+    }
```
matamegger, line 212: I think it might be worth tracking the view a bit closer and cleanup old view... transforming this from an `add` to a `set`.. similar thing would be appropriate with the `addPlayerView` as well. We hold the view instance explicitly, but still add it blindly to the parent view

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
```diff
@@ -1179,19 +1180,31 @@ class JsonConverter {
...
          */
         fun toPlayerViewConfig(json: ReadableMap) = PlayerViewConfig(
-            uiConfig = UiConfig.WebUi(
-                playbackSpeedSelectionEnabled = json.getMap("uiConfig")
-                    ?.getBooleanOrNull("playbackSpeedSelectionEnabled")
-                    ?: true,
-            ),
+            uiConfig = if (json.getMap("styleConfig")?.getBooleanOrNull("isUiEnabled") == false) {
+                UiConfig.Disabled
+            } else {
```
matamegger, line 1185: I think this is already taken care of in the native player (cc @strangesource)

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
```diff
@@ -264,11 +268,25 @@ class RNPlayerViewManager(private val context: ReactApplicationContext) : Simple
...
+                    UserInterfaceType.Bitmovin
+                ) {
+                    playerView = PlayerView(currentActivity, player, nativePlayerViewConfig)
+                } else {
+                    val overwrittenConfig = nativePlayerViewConfig.copy(uiConfig = UiConfig.Disabled)
+                    playerView = PlayerView(currentActivity, player, overwrittenConfig)
+                    val subtitleView = SubtitleView(currentActivity)
+                    subtitleView.setPlayer(player)
+                    view.addSubtitleView(subtitleView)
+                }
```
matamegger, line 288: For me this is very hard to read.
I am aware of the limitation we stated on the `PlayerViewConfig.uiConfig`
> Configuring the `uiConfig` only has an effect if the {@link StyleConfig.userInterfaceType} is set to {@link UserInterfaceType.Bitmovin}.

And that we have to take care of this now on android, as the `UserInterfaceType` is now also valid for android.

Maybe the (in another comment suggested) split of the `RNPlayerViewConfig` and the `userInterfaceType` makes this already clearer.

matamegger (commented):
Nice first contribution!
I think we should revisit the view injection logic.

example/src/screens/NoUiSubtitlesOnly.tsx
```diff
@@ -0,0 +1,110 @@
...
+    }, [player])
+  );
+
+  const onReady = useCallback((event: Event) => {
+    prettyPrint(`EVENT [${event.name}]`, event);
+  }, []);
+
+  const onEvent = useCallback((event: Event) => {
+    prettyPrint(`EVENT [${event.name}]`, event);
+  }, []);
```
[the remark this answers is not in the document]
testcenter: fc4e2147567cbfc7e82359d6cb5bc58821c6994c

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
testcenter: https://github.com/bitmovin/bitmovin-player-react-native/pull/316/commits/501f7bc8f2097f0f8970a0392e6413ae2c1c0875

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
testcenter: removed: https://github.com/bitmovin/bitmovin-player-react-native/pull/316/commits/5f185a9687dd49771a27cf81d0a3089c4f5f05ab

add different comment: https://github.com/bitmovin/bitmovin-player-react-native/pull/316/commits/054bc2786ca5588a10bee9d7981bfbbc43336d5a

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
testcenter: add cleanup (`removeView`) logic to `playerView` and `subtitleView`:
https://github.com/bitmovin/bitmovin-player-react-native/pull/316/commits/054bc2786ca5588a10bee9d7981bfbbc43336d5a

do you know if there is some additional method call necessary besides `removeView`? :thinking:

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
testcenter: I thought so too, but the `isUiEnabled` didn't work for me at all before adding this line :shrug: 

maybe @strangesource can shed some light :thinking:

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
testcenter: please take another look - there were multiple commits updating the readability of this logic

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
strangesource: As we provide both, PlayerViewConfig and StyleConfig, PlayerViewConfig overrides the StyleConfig values. See https://cdn.bitmovin.com/player/android/3/docs/player-ui-web/com.bitmovin.player/-player-view/config.html for reference.

strangesource: I'll create a ticket to completely remove the deprecated (internal) `StyleConfig` usage in RN/Android.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
```diff
@@ -255,6 +259,12 @@ class RNPlayerViewManager(private val context: ReactApplicationContext) : Simple
             val pictureInPictureHandler = view.pictureInPictureHandler ?: RNPictureInPictureHandler(context)
             view.pictureInPictureHandler = pictureInPictureHandler
             view.pictureInPictureHandler?.isPictureInPictureEnabled = isPictureInPictureEnabled
+
+            val rnPlayerConfigWrapper = view.config
+                ?: playerConfig?.let { JsonConverter.toRNPlayerViewConfigWrapper(it) }
+            val rnStyleConfigWrapper = playerConfig?.let { JsonConverter.toRNStyleConfigWrapper(it) }
+            val nativePlayerViewConfig = rnPlayerConfigWrapper?.playerViewConfig ?: PlayerViewConfig()
```
matamegger, line 266: I understand what you are doing, but I think it is a bit confusing / difficult to see.

A `PlayerConfig` can't be a `PlayerViewConfig`, but the code here suggests that we are doing this "cast" here. And it indeed works, because you added in that case in the `toRNPlayerViewConfigWrapper` function, but as said I think that's quite confusing.
I'll reach out to you in a place we can have a faster paced discussion.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -145,16 +147,21 @@ class RNPlayerView(
...
-    var playerView: PlayerView? = null
+    private var _playerView: PlayerView? = null
         set(value) {
             field = value
             viewEventRelay.eventEmitter = field
             playerEventRelay.eventEmitter = field?.player
         }
 
+    /**
+     * Associated bitmovin's `PlayerView`.
```
strangesource, line 158:
```suggestion
     * Associated Bitmovin's `PlayerView`.
```

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -303,3 +324,12 @@ data class RNPlayerViewConfigWrapper(
     val playerViewConfig: PlayerViewConfig?,
     val pictureInPictureConfig: RNPictureInPictureHandler.PictureInPictureConfig?,
 )
+
+data class RNStyleConfigWrapper(
```
strangesource, line 328: NIT: not sure if the RN prefix is needed as there is no such wrapper in the Android Player SDK which we could confuse this with.
```suggestion
data class StyleConfigWrapper(
```

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
strangesource: https://github.com/bitmovin/bitmovin-player-react-native/pull/316/files#r1388972875 still holds. What do you think about this?

strangesource (approved):
Good to go from my side, no more blockers. Some questions/NIT though.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
testcenter: I would keep it to be aligned with other wrappers (eg a few lines above `RNPlayerViewConfigWrapper`)

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
testcenter: https://github.com/bitmovin/bitmovin-player-react-native/pull/316/commits/9bfe41b916a04e44667e52e559843fdca1dc374b

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
testcenter: removed with https://github.com/bitmovin/bitmovin-player-react-native/pull/316/commits/a729aae634392043ada8ca50648c290eb297189f

example/src/App.tsx
```diff
@@ -31,6 +32,7 @@ export type RootStackParamsList = {
   BasicAds: undefined;
   BasicAnalytics: undefined;
   BasicPlayback: undefined;
+  NoUiSubtitlesOnly: undefined;
```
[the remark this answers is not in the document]
testcenter: I have no strong opinion on that. :+1: 
I added it to have the change testable.

whats your take on this @matamegger

matamegger (approved):

example/src/App.tsx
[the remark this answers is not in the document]
matamegger: I don't see a particular need, but I have no strong opinion on that.

It would align with a similar samples from the android player, if we also add a react player controls UI on top of that 😄 but that's out of scope for this PR I guess.

I think it's fine if there is not sample for this.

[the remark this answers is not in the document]
testcenter: removed the example: https://github.com/bitmovin/bitmovin-player-react-native/pull/316/commits/942d9b0e98e4a24a89ba0cc265c91fbe747f3917

PR #317, opened 2023-11-10
refactor: move helpers in base module

krocard:
## Problem (PRN-74)
The codebase had many null issues:
 - invalid null parameters were ignored instead of reporting an error
 - invalid state (module not loaded) were ignored instead of reporting an error
 - many functions were taking null parameters instead of non null, complexifying the code
 - errors were not reported

## Changes
- Refactor json serialization to use null operator `?.` rather than operating on null values.
- Add a `toJson` extension function to easily chain `?.` operations
- add a `Json.toXXX` extension function to easily chain `?.` operations
- change getModule to throw exception if module are not loaded, which should never happen 
- Wrap all `@ReactNative` methods in a catch to report any exception instead of crashing the app
- Introduce a base Bitmovin module for helper functions.
- All the above simplify the code significantly 

## Checklist
- [ ] 🗒 `CHANGELOG` entry if applicable
- [x] ⚗️ `New Test Streams` added to [collection](https://bitmovin.atlassian.net/wiki/spaces/PTS) if applicable


android/src/main/java/com/bitmovin/player/reactnative/BitmovinBaseModule.kt
```diff
@@ -0,0 +1,50 @@
...
+            runAndRejectOnException(promise) {
+                promise.resolve(block())
+            }
+        }
+    }
+
+    /**
+     * Helper function that returns the initialized [UIManagerModule] instance or throw.
+     */
+    protected fun uiManager(): UIManagerModule =
```
krocard, line 22: Those methods were duplicated in each module. Moved them here to deduplicate. They should probably be properties but didn't have time to refactor that.

krocard: As a follow up, those methods should be only accessible in the `runAndRejectOnException` as their nullable alternative should be used instead to be able to set `promise.reject`. They should also be properties.

android/src/main/java/com/bitmovin/player/reactnative/BitmovinBaseModule.kt
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * Helper function that gets the instantiated [PlayerModule] from modules registry.
+     */
+    protected fun playerModule(): PlayerModule =
+        context.getNativeModule(PlayerModule::class.java) ?: throw IllegalArgumentException("PlayerModule not found")
+}
+
+/** Run [block], forwarding the return value. If it throws, sets [Promise.reject] and return null. */
+inline fun <T> runAndRejectOnException(promise: Promise, crossinline block: () -> T): T? = try {
```
krocard, line 45: The promise should be the receiver.

android/src/main/java/com/bitmovin/player/reactnative/BufferModule.kt
```diff
@@ -32,36 +32,24 @@ class BufferModule(private val context: ReactApplicationContext) : ReactContextB
                 player.buffer.getLevel(bufferType, MediaType.Audio),
                 player.buffer.getLevel(bufferType, MediaType.Video),
             )
-            JsonConverter.fromRNBufferLevels(bufferLevels).let {
-                promise.resolve(it)
-            }
+            promise.resolve(bufferLevels.toJson())
```
krocard, line 35: promise is already resolved by addUIBlock
```suggestion
            bufferLevels.toJson()
```

android/src/main/java/com/bitmovin/player/reactnative/BufferModule.kt
```diff
@@ -2,28 +2,28 @@ package com.bitmovin.player.reactnative
...
-     * @param type The [type of buffer][JsonConverter.toBufferType] to return the level for.
+     * @param type The [type of buffer][toBufferType] to return the level for.
      * @param promise JS promise object.
      */
     @ReactMethod
     fun getLevel(nativeId: NativeId, type: String, promise: Promise) {
-        uiManager()?.addUIBlock { _ ->
-            val player = playerModule()?.getPlayer(nativeId) ?: return@addUIBlock
-            val bufferType = JsonConverter.toBufferType(type)
+        addUIBlock(promise) {
```
krocard, line 24: Rename to: (and consider making the promise the receiver)
```suggestion
        resolveOnUIThread(promise) {
```

android/src/main/java/com/bitmovin/player/reactnative/BufferModule.kt
```diff
@@ -32,36 +32,24 @@ class BufferModule(private val context: ReactApplicationContext) : ReactContextB
...
      */
     @ReactMethod
-    fun setTargetLevel(nativeId: NativeId, type: String, value: Double) {
-        uiManager()?.addUIBlock { _ ->
-            val player = playerModule()?.getPlayer(nativeId) ?: return@addUIBlock
-            val bufferType = JsonConverter.toBufferType(type) ?: return@addUIBlock
+    fun setTargetLevel(nativeId: NativeId, type: String, value: Double, promise: Promise) {
+        addUIBlock(promise) {
+            val player = playerModule().getPlayer(nativeId) ?: return@addUIBlock
+            val bufferType = type.toBufferType() ?: return@addUIBlock
```
krocard, line 49: Can we simplify the return?
```suggestion
            val bufferType = type.toBufferType() ?: return
```

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -283,10 +284,10 @@ class RNPlayerView(
...
      */
     private inline fun <reified E : Event> emitEvent(name: String, event: E) {
-        val payload = if (event is PlayerEvent) {
-            JsonConverter.fromPlayerEvent(event)
-        } else {
-            JsonConverter.fromSourceEvent(event as SourceEvent)
+        val payload = when (event) {
+            is PlayerEvent -> event.fromPlayerEvent()
+            is SourceEvent -> event.fromSourceEvent()
+            else -> throw IllegalArgumentException()
```
krocard, line 290: Is this throw safe ? Can we have a complete when instead? (is Event a sealed class?)

android/src/main/java/com/bitmovin/player/reactnative/SourceModule.kt
```diff
@@ -29,16 +31,10 @@ class SourceModule(private val context: ReactApplicationContext) : ReactContextB
...
-     * @return The associated `Source` instance or `null`.
+     * Fetches the [Source] instance associated with [nativeId] from internal sources or throws.
      */
-    fun getSource(nativeId: NativeId?): Source? {
-        if (nativeId == null) {
-            return null
-        }
-        return sources[nativeId]
-    }
+    fun getSource(nativeId: NativeId): Source = sources[nativeId]
```
krocard, line 36: Add a receiver `throwSafeContext` to make sure this is not called in a context that would crash the app?

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
krocard, line 1: Make sure `Hide whitespace` is ticked to review this file.

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
```diff
@@ -50,1235 +50,773 @@ import com.bitmovin.player.api.ui.UiConfig
...
 /**
- * Helper class to gather all conversion methods between JS -> Native objects.
+ * Converts an arbitrary `json` to `PlayerConfig`.
  */
-class JsonConverter {
-    companion object {
-        /**
-         * Converts an arbitrary `json` to `PlayerConfig`.
-         * @param json JS object representing the `PlayerConfig`.
-         * @return The generated `PlayerConfig` if successful, `null` otherwise.
```
krocard, line 76: The `@param` and `@return` documentation was only describing the type and name of the parameter, which is against our style guide. They have thus been removed.

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
```diff
@@ -50,1235 +50,773 @@ import com.bitmovin.player.api.ui.UiConfig
...
-                customReceiverConfig = json.getMap("customReceiverConfig")
-                    ?.toHashMap()
-                    ?.mapValues { entry -> entry.value as String } ?: emptyMap()
-            }
+/**
+ * Converts an arbitrary [ReadableMap] to a [RemoteControlConfig].
+ */
+private fun ReadableMap.toRemoteControlConfig(): RemoteControlConfig = RemoteControlConfig().apply {
+    withString("receiverStylesheetUrl") { receiverStylesheetUrl = it }
+    getMap("customReceiverConfig")
```
krocard, line 109:
```
withMap("customReceiverConfig") { customReceiverConfig = it.toHashMap().mapValues { entry -> entry.value as String } }
```
Do we even need `toHashMap`?

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
```diff
@@ -50,1235 +50,773 @@ import com.bitmovin.player.api.ui.UiConfig
...
-            return playbackConfig
-        }
+/**
+ * Converts any JS object into a `StyleConfig` object.
+ */
+fun ReadableMap.toStyleConfig(): StyleConfig = StyleConfig().apply {
+    withBoolean("isUiEnabled") { isUiEnabled = it }
+    getString("playerUiCss")?.takeIf { it.isNotEmpty() }?.let { playerUiCss = it }
+    getString("supplementalPlayerUiCss")?.takeIf { it.isNotEmpty() }?.let { supplementalPlayerUiCss = it }
+    getString("playerUiJs")?.takeIf { it.isNotEmpty() }?.let { playerUiJs = it }
```
krocard, line 159: is the filtering on empty needed?

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
```diff
@@ -50,1235 +50,773 @@ import com.bitmovin.player.api.ui.UiConfig
...
-            }
-            return styleConfig
-        }
+/**
+ * Converts any JS object into a `TweaksConfig` object.
+ */
+fun ReadableMap.toTweaksConfig(): TweaksConfig = TweaksConfig().apply {
+    withDouble("timeChangedInterval") { timeChangedInterval = it }
+    withInt("bandwidthEstimateWeightLimit") { bandwidthEstimateWeightLimit = it }
+    getMap("devicesThatRequireSurfaceWorkaround")?.let { devices ->
```
krocard, line 169:
```suggestion
    withMap("devicesThatRequireSurfaceWorkaround") { devices ->
```

android/src/main/java/com/bitmovin/player/reactnative/BufferModule.kt
matamegger, line 24: I think we might have a small problem with this function and block.
The `addUiBlock` function automatically resolves the `promise` with the return value of the `block`. However, here the `block` is also executing code that resolves or rejects the promise 🤔 
Further, the `return` statement one line below would trigger a `resolve(Unit)` in the `addUiBlock`, which might be not expected. (It's already a lot better than before though, as with the new solution we at least resolve the promise eventually)

I think we need to find a contract (best would be compile time save), to handle those situations. (`player` being `null` should probably also reject the promise)

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
```diff
@@ -50,1235 +50,773 @@ import com.bitmovin.player.api.ui.UiConfig
...
+    if (url == null || type == null) {
+        return null
+    }
+    return SourceConfig(url, type).apply {
+        title = getString("title")
+        description = getString("description")
+        posterSource = getString("poster")
+        if (hasKey("isPosterPersistent")) {
+            isPosterPersistent = getBoolean("isPosterPersistent")
+        }
```
krocard, line 242:
```suggestion
        withString("isPosterPersistent") { isPosterPersistent = it }
```

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
```diff
@@ -50,1235 +50,773 @@ import com.bitmovin.player.api.ui.UiConfig
...
+        is SourceEvent.VideoDownloadQualityChanged -> {
+            json.putMap("newVideoQuality", newVideoQuality?.fromVideoQuality())
+            json.putMap("oldVideoQuality", oldVideoQuality?.fromVideoQuality())
+        }
 
-                is PlayerEvent.VideoPlaybackQualityChanged -> {
-                    json.putMap("newVideoQuality", fromVideoQuality(event.newVideoQuality))
-                    json.putMap("oldVideoQuality", fromVideoQuality(event.oldVideoQuality))
-                }
+        else -> {
```
krocard, line 362: enumerate all events so tserialization of new ones are not forgotten.

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
```diff
@@ -50,1235 +50,773 @@ import com.bitmovin.player.api.ui.UiConfig
...
-            val sourceCustomData = toAnalyticsCustomData(json) ?: CustomData()
-            SourceMetadata(
-                title = it.getString("title"),
-                videoId = it.getString("videoId"),
-                cdnProvider = it.getString("cdnProvider"),
-                path = it.getString("path"),
-                isLive = it.getBoolean("isLive"),
-                customData = sourceCustomData,
-            )
+        else -> {
```
krocard, line 487: enumerate all events so tserialization of new ones are not forgotten.

android/src/main/java/com/bitmovin/player/reactnative/extensions/ReadableMapExtension.kt
matamegger, line 1: This is very nice!

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
```diff
@@ -50,1235 +50,773 @@ import com.bitmovin.player.api.ui.UiConfig
...
-            }
-        }
+/**
+ * Converts an arbitrary `json` to `WidevineConfig`.
+ */
+fun ReadableMap.toWidevineConfig(): WidevineConfig? = getMap("widevine")?.run {
+    WidevineConfig(getString("licenseUrl")).apply {
+        preferredSecurityLevel = getOrDefault("preferredSecurityLevel", null)
+        getBooleanOrNull("shouldKeepDrmSessionsAlive")?.let { shouldKeepDrmSessionsAlive = it }
+        httpHeaders = getMap("httpHeaders")
```
krocard, line 509: use `withMap`

krocard, line 509: Convert to `with*`

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
```diff
@@ -50,1235 +50,773 @@ import com.bitmovin.player.api.ui.UiConfig
...
-        fun toJson(options: OfflineContentOptions?): WritableMap? {
-            if (options == null) {
-                return null
-            }
+/**
+ * Converts an arbitrary `json` into a `SubtitleTrack`.
+ */
+fun ReadableMap.toSubtitleTrack(): SubtitleTrack? {
+    val url = this.getString("url")
+    val label = this.getString("label")
```
krocard, line 537: Use `?: return`

krocard, line 537:
```suggestion
    val url = this.getString("url") ?: return null
    val label = this.getString("label") ?: return null
```

matamegger (commented):
Very nice improvements!

I think we need some more work on the `uiManager`/`Promise` thing, but even that is already a good starting point!

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
```diff
@@ -50,1235 +50,773 @@ import com.bitmovin.player.api.ui.UiConfig
...
+}.build()
+
+/**
+ * Converts an arbitrary json object into an analytics `CustomData`.
+ */
+fun ReadableMap.toAnalyticsCustomData(): CustomData = CustomData.Builder().apply {
+    for (n in 1..30) {
+        this[n] = getString("customData$n")
+    }
+    getString("experimentName")?.let {
```
krocard, line 674: Use `withString`

android/src/main/java/com/bitmovin/player/reactnative/extensions/ReadableMapExtension.kt
```diff
@@ -6,6 +6,35 @@ fun ReadableMap.getBooleanOrNull(
     key: String,
 ): Boolean? = takeIf { hasKey(key) }?.getBoolean(key)
 
+fun ReadableMap.getDoubleOrNull(
+    key: String,
+): Double? = takeIf { hasKey(key) }?.getDouble(key)
```
krocard, line 11: remove?

android/src/main/java/com/bitmovin/player/reactnative/extensions/ReadableMapExtension.kt
```diff
@@ -6,6 +6,35 @@ fun ReadableMap.getBooleanOrNull(
...
+
+inline fun <T> ReadableMap.withDouble(
+    key: String,
+    block: (Double) -> T,
+): T? = takeIf { hasKey(key) }?.getDouble(key)?.let(block)
```
krocard, line 16: factorize:
```
fun <T> mapAndDo(key: String, get: ReadableMap.() -> T, block: (T) -> R) = takeIf{ hasKey(key) }?.get().let(block)
```

android/src/main/java/com/bitmovin/player/reactnative/BufferModule.kt
krocard: Done: 14fc20823ca233e4fbfa1f7e165700be5e2ef314

android/src/main/java/com/bitmovin/player/reactnative/BufferModule.kt
krocard: Fixed: f2ff634e7045f1399b537de620e811bb58d94dc2

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
krocard: It looks like exhaustive `when` is only possible in the same module as the base class? :thinking: Or at least I did not manage to make it compile with `is OflineEvent`.

android/src/main/java/com/bitmovin/player/reactnative/BitmovinBaseModule.kt
krocard: Done

android/src/main/java/com/bitmovin/player/reactnative/SourceModule.kt
krocard: Done: b39491e050674d0a134c35a1e9690c90872eebc2 (it wasn't very complicated.)

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
krocard: Yes, but moved it in to a helper function: 899d4f0749701ad6aba1fdd93810d1f53be2eb97

android/src/main/java/com/bitmovin/player/reactnative/DrmModule.kt
```diff
@@ -75,9 +75,9 @@ class DrmModule(private val context: ReactApplicationContext) : ReactContextBase
      */
     @ReactMethod
     fun initWithConfig(nativeId: NativeId, config: ReadableMap?) {
-        uiManager()?.addUIBlock {
-            if (!drmConfigs.containsKey(nativeId) && config != null) {
-                JsonConverter.toWidevineConfig(config)?.let {
+        context.uiManagerModule?.addUIBlock {
```
zigavehovec, line 78: Why isn't this class a BitmovinBaseModule with the same resolveOnUIThread logic?

android/src/main/java/com/bitmovin/player/reactnative/OfflineModule.kt
```diff
@@ -58,11 +59,11 @@ class OfflineModule(private val context: ReactApplicationContext) : ReactContext
      */
     @ReactMethod
     fun initWithConfig(nativeId: NativeId, config: ReadableMap?, drmNativeId: NativeId?, promise: Promise) {
-        uiManager()?.addUIBlock {
+        context.uiManagerModule?.addUIBlock {
```
zigavehovec, line 62: Why isn't this class a BitmovinBaseModule with the same resolveOnUIThread logic?

zigavehovec: Is there a case where config can actually be null?

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
```diff
@@ -248,7 +248,11 @@ class RNPlayerViewManager(private val context: ReactApplicationContext) : Simple
      */
     private fun attachPlayer(view: RNPlayerView, playerId: NativeId?, playerConfig: ReadableMap?) {
```
zigavehovec, line 249: What's the case of playerId being null here? I think we should log a no-op error if  `args?.getString(1)` in receiveCommand returns null.

android/src/main/java/com/bitmovin/player/reactnative/extensions/ReadableMapExtension.kt
```diff
@@ -1,25 +1,43 @@
...
- */
-fun ReadableMap.getOrDefault(
+inline fun <T> ReadableMap.withMap(
     key: String,
-    default: String?,
-) = if (hasKey(key)) getString(key) else default
+    block: (ReadableMap) -> T,
+): T? = takeIf { hasKey(key) }?.getMap(key)?.let(block)
+
+inline fun <T> ReadableMap.withInt(
```
zigavehovec, line 20: Nit: I think that the convention is to include the nullability in the name.

```suggestion
inline fun <T> ReadableMap.withIntOrNull(
```

zigavehovec (commented):
Great work! I added some questions that you might already be aware about 🙂

android/src/main/java/com/bitmovin/player/reactnative/extensions/ReadableMapExtension.kt
matamegger: That convention is for `getter`. I think this is a bit more complex, and actually, we only take care of the non null int... however, the no-op on null might be not clear enough.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
krocard: An error is now logged: 2cb51f81323d791aa502b8b5d3a61ad9203605b4

android/src/main/java/com/bitmovin/player/reactnative/extensions/ReadableMap.kt
```diff
@@ -2,18 +2,18 @@ package com.bitmovin.player.reactnative.extensions
...
+        put(it.key, it.value)
     }
 }
+
+fun Map<String, Boolean>.toReadableMap(): ReadableMap = toReadableMap(WritableMap::putBoolean)
+fun Map<String, String>.toReadableMap(): ReadableMap = toReadableMap(WritableMap::putString)
+fun Map<String, Double>.toReadableMap(): ReadableMap = toReadableMap(WritableMap::putDouble)
+fun Map<String, Int>.toReadableMap(): ReadableMap = toReadableMap(WritableMap::putInt)
+fun Map<String, ReadableArray>.toReadableMap(): ReadableMap = toReadableMap(WritableMap::putArray)
+fun Map<String, ReadableMap>.toReadableMap(): ReadableMap = toReadableMap(WritableMap::putMap)
```
matamegger, line 19: I think you wanted to prepare for future usage, but I think in this case it might be YAGNI. Also because of type erasure this does not compile as it ends up to be the same signature for all of them. (we can avoid this by annotating with `@JvmName( ... )`)
I think we can remove them for now and add later them if they are needed in future.
```suggestion
fun Map<String, String>.toReadableMap(): ReadableMap = toReadableMap(WritableMap::putString)
```

android/src/main/java/com/bitmovin/player/reactnative/BitmovinBaseModule.kt
```diff
@@ -0,0 +1,71 @@
...
+ */
+abstract class BitmovinBaseModule(
+    protected val context: ReactApplicationContext,
+) : ReactContextBaseJavaModule(context) {
+    /** [resolve] the [Promise] by running [block] in the UI thread with [UIManagerModule.addUIBlock].  */
+    protected inline fun <T> Promise.resolveOnUIThread(crossinline block: RejectPromiseOnExceptionBlock.() -> T) {
+        val uiManager = runAndRejectOnException { uiManager } ?: return
+        uiManager.addUIBlock {
+            resolveOnCurrentThread {
+                resolve(block())
```
matamegger, line 30: I think `resolveOnCurrentThread` already takes care of resolving.
```suggestion
                block()
```

android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt
```diff
@@ -83,360 +82,339 @@ class PlayerModule(private val context: ReactApplicationContext) : ReactContextB
...
-    fun destroy(nativeId: NativeId) {
-        uiManager()?.addUIBlock {
-            players[nativeId]?.let {
-                it.destroy()
-                players.remove(nativeId)
-            }
+    fun destroy(nativeId: NativeId, promise: Promise) {
+        promise.resolveOnUIThread {
+            getPlayer(nativeId).destroy()
+            players.remove(nativeId)
```
matamegger, line 205: This block will resolve with the removed player instance. That's not what we want.
Maybe introducing a function that resolves without a return value could reduce such accidental returns?

matamegger (changes requested):
Unfortunately, I see messages in the example app, that state we might have unhandled promise rejections. Most of them can be tackled with two of my suggestions I think.

Some screen of the sample do not work anymore:
- `Subtitle and captions` shows a player without a source (UI is visible though)
- `Programmatic Track Selection` sample does not seem to work anymore (no subtitles are listed)
- `Basic Ad`, app crashes after ad played back
- `Offline Playback`, cannot convert argument (`kotlin.Unit`)
- `Casting`, just a black screen, no interaction possible

android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt
matamegger: This is just one place I spotted, but there might be more where we accidentally return something.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
zigavehovec: What would you think about being more explicit in the `receiveCommand` function and then having playerId not nullable here?

Something like :
```
Commands.ATTACH_PLAYER -> {
    val playerId = args?.getString(1) ?: error("cannot attach a player without a playerId")
    val playerConfig = args.getMap(2)
    attachPlayer(view, playerId, playerConfig)
}
```

I would do this for all cases and make sure that parameters that cannot be null have to be provided. I would suggest a follow up ticket if you agree.

zigavehovec (commented):
Very nice improvements! 💪

As @matamegger mentioned the project currently doesn't compile, I agree with his suggestions. 
Apart from that it looks good to me, so I'll approve once the suggestions are applied and I can test again.

android/src/main/java/com/bitmovin/player/reactnative/DrmModule.kt
krocard: The DRM module was not refactor. It is now (by many patches.)

android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt
krocard: With the new type safe promise, this should not occur.

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
krocard: I left it as is as this refactor doesn't intend to introduce any behavior differences.

android/src/main/java/com/bitmovin/player/reactnative/BitmovinBaseModule.kt
matamegger: It's fine, it should create an error on the react side, and should be covered by the soon existing system tests as well.

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
krocard: Done: cb07f50a959d4c96fcd9bf5bc614760d40d596d4

android/src/main/java/com/bitmovin/player/reactnative/BufferModule.kt
krocard: This is partially fixed:
 - null player rejects the promise
 - promise is now typed, so unit can't be returned by mistake

The one thing that I could not fix (and I suspect is not fixable without affine types) is the code calling `promise.resolve` explicitly instead of using the provided helpers.
I made sure no Promise.resolve was left in the code.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
krocard: Done: 8773de7fd55d2b9cc5cb989d80ef831ad03ea60a

android/src/main/java/com/bitmovin/player/reactnative/extensions/ReadableMap.kt
krocard: Removed the unused variant :+1:

matamegger (commented):
Neat that you found a solution for the Promises. This makes future review and coding a lot easier 👍 

As discussed, there are still some errors, but apparently the fault is on the `Arguments.fromList`.
Also, I am not sure if the `nullability` is right on all the promises. I think we have to align with the type that is defined in the TS API.

android/src/main/java/com/bitmovin/player/reactnative/BitmovinBaseModule.kt
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * In general, code should not throw while resolving a [Promise]. Instead, [Promise.reject] should be used.
+ * This doesn't match Kotlin's error style, which uses exception. The helper methods in this class, provide such
+ * convenience, they can only be called in a context that will catch any Exception and reject the [Promise].
+ *
+ */
+abstract class BitmovinBaseModule(
+    protected val context: ReactApplicationContext,
+) : ReactContextBaseJavaModule(context) {
+    /** [resolve] the [Promise] by running [block] in the UI thread with [UIManagerModule.addUIBlock].  */
```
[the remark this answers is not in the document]
krocard: Done: 6231fd9d4f46a9de51fe2f7d1610d231f06c15a7

android/src/main/java/com/bitmovin/player/reactnative/PlayerAnalyticsModule.kt
```diff
@@ -37,22 +34,8 @@ class PlayerAnalyticsModule(private val context: ReactApplicationContext) : Reac
      */
     @ReactMethod
     fun getUserId(playerId: NativeId, promise: Promise) {
-        uiManager()?.addUIBlock { _ ->
-            playerModule()?.getPlayer(playerId)?.analytics?.let {
-                promise.resolve(it.userId)
-            }
+        promise.string.nullable.resolveOnUiThread {
```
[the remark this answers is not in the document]
krocard: This should never happen as this function is not called if there is no analytics.
The code now throws if there is no analytics.

android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt
```diff
@@ -83,360 +82,337 @@ class PlayerModule(private val context: ReactApplicationContext) : ReactContextB
...
     fun getAvailableSubtitles(nativeId: NativeId, promise: Promise) {
-        uiManager()?.addUIBlock {
-            val subtitleTracks = Arguments.createArray()
-            players[nativeId]?.source?.availableSubtitleTracks?.let { tracks ->
-                tracks.forEach {
-                    subtitleTracks.pushMap(JsonConverter.fromSubtitleTrack(it))
-                }
-            }
-            promise.resolve(subtitleTracks)
+        promise.array.nullable.resolveOnUiThreadWithPlayer(nativeId) {
```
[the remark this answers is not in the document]
krocard: Good catch. The expectation is to return an empty array if it's null. Done: 08ab485

android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt
```diff
@@ -538,80 +514,57 @@ class PlayerModule(private val context: ReactApplicationContext) : ReactContextB
...
-        context.getNativeModule(OfflineModule::class.java)
+    private inline fun <T> TPromise<T>.resolveOnUiThreadWithPlayer(
+        nativeId: NativeId,
+        crossinline block: Player.() -> T,
+    ) = resolveOnUiThread { getPlayer(nativeId, this@PlayerModule).block() }
 }
+
+private inline fun <T> List<T>.mapToReactArray(
+    transform: (T) -> WritableMap,
+): WritableArray = Arguments.fromList(map(transform))
```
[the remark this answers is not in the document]
krocard: Done: 2c0d8877db300287a4c1217c26acba70f7f76779

src/components/PlayerView/index.tsx
```diff
@@ -49,7 +50,7 @@ export function PlayerView({
   fullscreenHandler,
   customMessageHandler,
   isFullscreenRequested = false,
-  scalingMode,
+  scalingMode = ScalingMode.Fit,
```
krocard, line 53: The alternative is to default in the Kotlin code on null.
Chose to default in the TS code aligned with `isFullscreenRequest`.

android/src/main/java/com/bitmovin/player/reactnative/BitmovinBaseModule.kt
```diff
@@ -0,0 +1,106 @@
...
+value class TPromise<T>(val promise: Promise) {
+    // Promise only support built-in types. Functions that return [Unit] must resolve to `null`.
+    fun resolve(value: T): Unit = promise.resolve(value.takeUnless { it is Unit })
+    fun reject(throwable: Throwable) {
+        Log.e(MODULE_NAME, "Failed to execute Bitmovin method", throwable)
+        promise.reject(throwable)
+    }
+}
+
+inline val Promise.int get() = TPromise<Int>(this)
```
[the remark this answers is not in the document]
matamegger: I think the idea behind the convenience properties is to also show which types are possible. (the supported types can not be nicely generalized)

+1 for the function instead of the property, but this should not block the PR.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
```diff
@@ -301,10 +285,13 @@ class RNPlayerViewManager(private val context: ReactApplicationContext) : Simple
         }
     }
 
-    /**
-     * Helper function that gets the instantiated `PlayerModule` from modules registry.
-     */
-    private fun getPlayerModule(): PlayerModule? = context.getModule()
+    private inline fun runInMainLooperAndLogException(crossinline block: () -> Unit) {
```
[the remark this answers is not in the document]
krocard: Done: 49945624a1fbf70083d290e001cd8ee836f40e5

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
```diff
@@ -250,28 +236,26 @@ class RNPlayerViewManager(private val context: ReactApplicationContext) : Simple
      * @param view Target `RNPlayerView`.
      * @param playerId `Player` instance id inside `PlayerModule`'s registry.
      */
-    private fun attachPlayer(view: RNPlayerView, playerId: NativeId?, playerConfig: ReadableMap?) {
-        Handler(Looper.getMainLooper()).post {
-            val player = getPlayerModule()?.getPlayer(playerId)
+    private fun attachPlayer(view: RNPlayerView, playerId: NativeId, playerConfig: ReadableMap?) {
+        runInMainLooperAndLogException {
+            val player = playerId.let { context.playerModule?.getPlayerOrNull(it) }
+                ?: throw InvalidParameterException("Cannot create a PlayerView, invalid playerId was passed.")
```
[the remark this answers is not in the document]
krocard: Done: d36c14d

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
```diff
@@ -154,44 +157,27 @@ class RNPlayerViewManager(private val context: ReactApplicationContext) : Simple
...
-            }
+            Commands.ATTACH_PLAYER -> attachPlayer(view, args?.getString(1).require(), args?.getMap(2))
+            Commands.ATTACH_FULLSCREEN_BRIDGE -> attachFullscreenBridge(view, args?.getString(1).require())
+            Commands.SET_CUSTOM_MESSAGE_HANDLER_BRIDGE_ID -> setCustomMessageHandlerBridgeId(
+                view,
+                args?.getString(1).require(),
+            )
+            Commands.SET_FULLSCREEN -> setFullscreen(view, args?.getBoolean(1).require())
+            Commands.SET_SCALING_MODE -> setScalingMode(view, args?.getString(1).require())
+            Commands.SET_PICTURE_IN_PICTURE -> setPictureInPicture(view, args?.getBoolean(1).require())
```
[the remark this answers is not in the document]
krocard: For the scaling mode, please refer to the [relevant thread](https://github.com/bitmovin/bitmovin-player-react-native/pull/317/files#r1414554252).
For fullscreen and picture in picture, they are set to false if null in TS:
```
export function PlayerView({
// ...
  isFullscreenRequested = false,
  scalingMode,
  isPictureInPictureRequested = false,
```

android/src/main/java/com/bitmovin/player/reactnative/BitmovinBaseModule.kt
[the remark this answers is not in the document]
krocard: > `promise.resolveOnUiThreadAs<Int>`

This was my first version, but rejected due to the following drawback:
 - the type can be invalid (only primitive types, array and map are supported)
 - the type is autodeduced, so the explicit type can be omitted
 - the goal of this is to make it explicit to the reader which type is returned
 - it's not possible to constraint the types at compile time with this method
 - more verbose (eg: `<ReadableMap?>`)
 - constraining the type is more composable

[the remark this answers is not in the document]
krocard: > `fun Promise.asIntPromise()`

I kept the naming short because it's used in many places and it explicit and type safe.
I'm not a fan of repeating `promise` twice.
Should we bring it to the JFX? (naming is hard and I want this pr merged ASAP)

matamegger (approved):
Nothing to add to @zigavehovec s comments.
After those topics are resolved this is fine to be merged.

android/src/main/java/com/bitmovin/player/reactnative/BitmovinBaseModule.kt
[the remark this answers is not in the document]
matamegger: I think the `inline fun Promise.asIntPromise()` would be more dogmatic, and still on the same level of type safety as the property.
I would though prefer to merge this PR rather sooner than later. Before this starts to be a discussion, let's get this PR merged and iterate on the system in follow up PRs.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
```diff
@@ -301,10 +285,13 @@ class RNPlayerViewManager(private val context: ReactApplicationContext) : Simple
         }
     }
 
-    /**
-     * Helper function that gets the instantiated `PlayerModule` from modules registry.
-     */
-    private fun getPlayerModule(): PlayerModule? = context.getModule()
+    private inline fun runInMainLooperAndLogException(crossinline block: () -> Unit) {
+        Handler(Looper.getMainLooper()).post {
```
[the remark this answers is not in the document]
krocard: Done: d27223627ce611989c64c880d124941e0799817b

android/src/main/java/com/bitmovin/player/reactnative/BitmovinBaseModule.kt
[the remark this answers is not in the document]
zigavehovec: I'm fine keeping this as is 👍

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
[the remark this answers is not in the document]
zigavehovec: I see, thanks for the explanation!

zigavehovec (approved):
Nice! Let's get this merged 💪

PR #193, opened 2023-08-11
React native update

krocard:
Pain, pain, pain... The update is not finished:
 - Kotlin code doesn't yet compile
 - swift update is not yet started

### React native library template
Generate a new RN library using the [official tool](https://reactnative.dev/docs/native-modules-setup):
``` bash
$ npx create-react-native-library@latest --slug "bitmovin-player-react-native" --description "Official React Native bindings for Bitmovin's mobile Player SDKs." --author-name "Bitmovin GmbH" --author-email "react@bitmovin.com" --author-url "https://github.com/bitmovin/bitmovin-player-react-native" --repo-url "https://github.com/bitmovin/bitmovin-player-react-native" native-module
> native-module
> kotlin & swift
# Fix paths
$ find -name 'bitmovinplayer*'| sed 's#bitmovinplayerreactnativeexample#bitmovin/player/reactnative#'| xargs mkdir -p
$ find -name 'bitmovinplayer*'| sed -E 's#(.*)/bitmovinplayerreactnativeexample#\0 \1/bitmovin/player/reactnative/example#'|xargs -L1 mv
$ grep -r "com.bitmovinplayerreactnative" --files-with-match native-module/* | xargs sed -i -e 's/com.bitmovinplayerreactnativeexample/com.bitmovin.player.reactnative.example/' -e 's/com.bitmovinplayerreactnative/com.bitmovin.player.reactnative/'
$ meld native-module ~/dev/bitmovin/player-react-native
```
I then went over all files that changed and applied the differences.

### Official tool upgrade tool (Fail)
Doc: https://reactnative.dev/docs/upgrading 
```
npx react-native upgrade 0.71
```

Web app:[ https://react-native-community.github.io/upgrade-helper/?from=0.69.7&to=0.71.12](https://react-native-community.github.io/upgrade-helper/?from=0.69.7&to=0.71.12)
React template app: https://github.com/facebook/react-native/blob/v0.72.1/packages/react-native/template
Tool doesn’t work, No file was changed.
*The tool expects files matching the react native app template. We have a library, not an app.*
I tried to still apply the diff to our example app, without success (doesn’t build).
### Align-deps (Fail)
``` bash
npx @rnx-kit/align-deps --requirements react-native@0.71 --write
```

PR #318, opened 2023-11-13
Update React Native

zigavehovec:
## Problem (PRN-66)
Our project is not using the latest React Native version

As @krocard mentioned here: https://github.com/bitmovin/bitmovin-player-react-native/pull/193, the official upgrade tool didn't work, so the following method was used: 
- Generated a new RN library using the [official tool](https://reactnative.dev/docs/native-modules-setup)
- Compared changes in https://react-native-community.github.io/upgrade-helper/?from=0.69.7&to=0.72.6
Using the 2 diffs I created changes in this PR.


example/package.json
```diff
@@ -13,15 +13,24 @@
...
     "react-native-modal": "13.0.1",
     "react-native-orientation-locker": "1.5.0",
-    "react-native-safe-area-context": "4.3.1",
+    "react-native-safe-area-context": "4.7.4",
     "react-native-screens": "3.15.0",
     "react-native-system-navigation-bar": "^2.6.1"
   },
   "devDependencies": {
-    "pod-install": "^0.1.39"
+    "@babel/core": "^7.20.0",
```
zigavehovec, line 25: These might not be needed

example/package.json
```diff
@@ -13,15 +13,24 @@
...
+    "@babel/preset-env": "^7.20.0",
+    "@babel/runtime": "^7.20.0",
+    "@react-native/metro-config": "^0.72.11",
+    "babel-plugin-module-resolver": "^5.0.0",
+    "metro-react-native-babel-preset": "0.76.8",
```
zigavehovec, line 30: These might not be needed.

example/react-native.config.js
```diff
@@ -0,0 +1,10 @@
+const path = require('path');
+const pak = require('../package.json');
+
+module.exports = {
```
zigavehovec, line 4: This code enabled auto-linking of the native modules.

example/src/hooks/useTVGestures.ts
```diff
@@ -1,4 +1,4 @@
-import 'react-native/tvos-types.d';
+//import 'react-native/tvos-types.d'; TODO: this is not working since updating RN version. Probably iOS linking issue
```
zigavehovec, line 1: I'm not sure if we need this line, I get no errors when removing it

example/package.json
rolandkakonyi: Still used in `example/babel.config.js`

zigavehovec: The library generated with the [official tool](https://reactnative.dev/docs/native-modules-setup) only contains babel dependencies in the example project, but after removing them from the main package.json, I observed this warnings: 

```
warning "react-native > @react-native/codegen@0.72.7" has unmet peer dependency "@babel/preset-env@^7.1.6".
warning "react-native > @react-native/codegen > jscodeshift@0.14.0" has unmet peer dependency "@babel/preset-env@^7.1.6".
warning "react-native > @react-native-community/cli > @react-native-community/cli-plugin-metro > metro-react-native-babel-transformer@0.76.8" has unmet peer dependency "@babel/core@*".
```

I think the best way forward would be to just have them in both packahe.json files. What do you think about that?

rolandkakonyi: Let's keep them to keep warning count low.

Brewfile.lock.json
rolandkakonyi, line 1: I added a small change to prevent this file from changing when running `yarn bootstrap` in 944088c64e6364c33f7643e26f1cfd9783c4317a

android/gradle.properties
```diff
@@ -1 +1,8 @@
 android.useAndroidX=true
+BitmovinPlayerReactNative_kotlinVersion=1.7.0
+BitmovinPlayerReactNative_minSdkVersion=21
+BitmovinPlayerReactNative_targetSdkVersion=31
+BitmovinPlayerReactNative_compileSdkVersion=31
```
rolandkakonyi, line 5: Did we change this from 33?

android/src/main/AndroidManifestNew.xml
rolandkakonyi, line 1: What is the purpose of this file?

example/android/build.gradle
```diff
@@ -2,51 +2,28 @@
 
 buildscript {
     ext {
-        buildToolsVersion = "31.0.0"
+        buildToolsVersion = "33.0.0"
         minSdkVersion = 21
         compileSdkVersion = 33
         targetSdkVersion = 33
```
rolandkakonyi, line 8: See my other comment, there we use 31 in `android/gradle.properties`

example/android/settings.gradle
```diff
@@ -1,11 +1,4 @@
 rootProject.name = 'BitmovinPlayerReactNativeExample'
 apply from: file("../node_modules/@react-native-community/cli-platform-android/native_modules.gradle"); applyNativeModulesSettingsGradle(settings)
 include ':app'
-
-include ':bitmovin-player-react-native'
```
rolandkakonyi, line 5: We don't need this anymore?

rolandkakonyi (commented):
Code review looks good!

I am still doing manual testing.

rolandkakonyi:
Manual testing results:
- iOS device: ✅
- iOS simulator: ✅ 
- tvOS device: ✅ 
- tvOS simulator: ✅
- Android emulator: ✅ 
- Android device: 🤷‍♂️ I don't have one

example/android/app/src/main/java/com/bitmovin/player/reactnative/example/MainActivity.java
```diff
@@ -30,25 +31,16 @@ protected String getMainComponentName() {
...
-      ReactRootView reactRootView = new ReactRootView(getContext());
-      // If you opted-in for the New Architecture, we enable the Fabric Renderer.
-      reactRootView.setIsFabric(BuildConfig.IS_NEW_ARCHITECTURE_ENABLED);
-      return reactRootView;
-    }
+    return new DefaultReactActivityDelegate(
+            this,
+            getMainComponentName(),
+            // If you opted-in for the New Architecture, we enable the Fabric Renderer.
+            DefaultNewArchitectureEntryPoint.getFabricEnabled());
```
krocard, line 44: Double checking, there should not be a `if NEW_ARCHITECTURE` here, right?

example/package.json
krocard: Which of our dependency triggers those warning? Aka which of our dependency has babel peer dependency? :thinking:
I don't see why we would need babel in our library. Only the app should need it.

example/src/screens/BasicPictureInPicture.tsx
```diff
@@ -118,7 +118,7 @@ export default function BasicPictureInPicture({
 
   return (
     <SafeAreaView
-      edges={['bottom', 'left', 'right']}
+      edges={Platform.isTV ? [] : ['bottom', 'left', 'right']}
```
krocard, line 121: Nice to see they are building official TV support :)

package.json
```diff
@@ -32,7 +38,7 @@
     "pods-install": "yarn example pod-install",
     "pods-update": "cd example/ios && pod update --silent",
     "bootstrap": "yarn && yarn example && yarn pods && yarn brew",
```
krocard, line 40: Did you avoid migrating those commands? I had a lot more changes when I tried a couple months ago:
https://github.com/bitmovin/bitmovin-player-react-native/pull/193/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R35

krocard (approved):
Very nice update! Congratulation on porting all the changes.

example/package.json
rolandkakonyi: looks like: `react-native/codegen` and `react-native-community/cli`

android/gradle.properties
zigavehovec: Yes, good eye I didn't notice this change. 
https://github.com/bitmovin/bitmovin-player-react-native/pull/318/commits/b1bb85e32a11a5e73996b56f6b3d06c1a02b9dbd

android/src/main/AndroidManifestNew.xml
zigavehovec: Since `AGP 7.3.0 ` namespace should be defined in the the `build.gradle` file instead of AndoridManifest.xml file. 
To avoid a warning the [official tool](https://reactnative.dev/docs/native-modules-setup) generates a "new" manifest file that is used instead of the default one. You can se the usage here: https://github.com/bitmovin/bitmovin-player-react-native/pull/318/files/2f2d0df3d8bbb462ffaacfc81b20df441f78ca9a#diff-197b190e4a3512994d2cebed8aff5479ff88e136b8cc7a4b148ec9c3945bd65aR56

example/android/settings.gradle
zigavehovec: Nope. This now auto-magically happens with the auto-linking support. Finding out about this was the biggest paint point during this migration, since [this](https://github.com/bitmovin/bitmovin-player-react-native/pull/318#discussion_r1395444381) is needed for gradle auto linking to work. 

You can read a bit about how this works here: https://github.com/react-native-community/cli/blob/main/docs/autolinking.md#platform-android

example/src/screens/BasicPictureInPicture.tsx
rolandkakonyi: Actually this came from me 😅. It actually doesn't work as intended on tvOS simulator, so I opened #329 for proper fix.

rolandkakonyi (approved):

example/android/app/src/main/java/com/bitmovin/player/reactnative/example/MainActivity.java
zigavehovec: Correct, logging `DefaultNewArchitectureEntryPoint.getFabricEnabled()` returns false. I couldn't verify if changing `newArchEnabled` to true influences this (I assume it does), because to enable the new arhitecture we also have to update other dependencies.

package.json
zigavehovec: Yes I avoided those. My strategy was to do minimal changes step by step and see what is needed and what not. Since everything seems to be working I guess we are fine, but I'll double check this before merging.

rolandkakonyi (changes requested):
There is one thing I didn't think before: CHANGELOG entry!

Please add a descriptive entry to the CHANGELOG describing any breaking changes (if any).

Do customers need any migration? It would be great to verify.

matamegger (approved):
Very nice that you managed to migrate!

Agree with the points @rolandkakonyi brought up i.e. a changelog entry would be very nice, and also we should verify if there are any migration steps needed for integrators.

rolandkakonyi:
@zigavehovec I fixed conflicts as they were mainly due to my changes.

rolandkakonyi:
We increased the min. supported version to `0.65.0` in #352 to keep the widest range of supported React Native versions available after this update.

rolandkakonyi (approved):

PR #323, opened 2023-11-15
Decouple player view and player lifecycle

matamegger:
## Description

The `RNPlayerView` takes charge of the `Player` life cycle. As the `Player` is created independently of the `PlayerView` (and then the player is explicitly attached to the view), it is unexpected that the `PlayerView` destroys the `Player` once itself is destroyed.

## Changes

Detach the `Player` from the `RNPlayerView`/`PlayerView` internally, when the `RNPlayerView` is disposed and before the `PlayerView` is destroyed.

## Checklist
- [x] 🗒 `CHANGELOG` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixed
 
-- Android: `PlayerView` destroys attached `Player` instance on destroy. `Player` life cycle must be handled on the creation side
+- Android: `PlayerView` destroys attached `Player` instance on destroy. `Player` lifecycle must be handled on the creation side
 
 ## [0.14.0] (2023-11-14)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [Unreleased]
+
+### Fixed
+
+- Android: `PlayerView` destroys attached `Player` instance on destroy. `Player` life cycle must be handled on the creation side
+
 ## [0.14.0] (2023-11-14)
 
 ### Added
--- a/android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
@@ -149,6 +149,7 @@ class RNPlayerView(
 
     private var _playerView: PlayerView? = null
         set(value) {
+            field?.removeOnLayoutChangeListener(this)
             field = value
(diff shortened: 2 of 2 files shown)


android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -186,10 +186,15 @@ class RNPlayerView(
      * Cleans up the resources and listeners produced by this view.
      */
     fun dispose() {
-        viewEventRelay.eventEmitter = null
-        playerEventRelay.eventEmitter = null
-        playerView?.removeOnLayoutChangeListener(this)
-        playerView?.onDestroy()
+        _playerView?.run {
```
zigavehovec, line 189: - I would prefer using a scope with a named parameter, avoiding the need to use `@RNPlayerView`. 
- Also do we not need the eventEmitter handling anymore?

matamegger: I don't think that other scope functions are actually applicable. Best matching would be `with` actually.. but then I have to explicitly take care of the nullability and we still have to specify which `this` we are using.
(`let` and `also` do not make sense from a wording perspective)

`eventEmitter` is handled by detatching (setting to `null`) the `_playerView`

https://github.com/bitmovin/bitmovin-player-react-native/pull/323/commits/ebd580ae7af786bdb422ae054258ef0d33b0aeec
https://github.com/bitmovin/bitmovin-player-react-native/pull/323/commits/beb7519a6a67603c35b986bffd8d2b04fa93e66e

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## [Unreleased]
+
+### Fixed
+
+- Android: `PlayerView` destroys attached `Player` instance on destroy. `Player` life cycle must be handled on the creation side
```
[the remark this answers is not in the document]
matamegger: https://github.com/bitmovin/bitmovin-player-react-native/pull/323/commits/2c29de123fd56a04c5cb85c54d3fc226b23eee79

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
zigavehovec: Very nice!

zigavehovec (approved):
Nice 👌

PR #327, opened 2023-11-23
Fix `onEvent` never triggered on `PlayerView` on Android

rolandkakonyi:
## Checklist
- [x] 🗒 `CHANGELOG` entry

### DIFF
--- a/android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
@@ -25,7 +25,6 @@ import com.facebook.react.uimanager.events.RCTEventEmitter
 import kotlin.reflect.KClass
 
 private val EVENT_CLASS_TO_REACT_NATIVE_NAME_MAPPING = mapOf(
-    PlayerEvent::class to "event",
     PlayerEvent.Error::class to "playerError",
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -25,7 +25,6 @@ import com.facebook.react.uimanager.events.RCTEventEmitter
 import kotlin.reflect.KClass
 
 private val EVENT_CLASS_TO_REACT_NATIVE_NAME_MAPPING = mapOf(
-    PlayerEvent::class to "event",
```
rolandkakonyi, line 28: This did nothing, so I removed.

strangesource: This is a bit of a flaw in our event emitter. From a typing perspective one would expect that this works - but it doesn't.

strangesource (approved):
Nice catch and clean solution. 💪

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
rolandkakonyi: Thanks for the clarification! It confused me too!

PR #328, opened 2023-11-23
Fix incomplete `onEvent` payload on iOS

rolandkakonyi:
## Description

`onEvent` listener receives only `name` and `timestamp` properties as its payload, regardless of the actual event.

## Changes

The problem is that functions declared in extensions in Swift don't support runtime polymorphism, therefore when calling `event.toJSON()` when `event` has type `Event`, it will invoke the one from `extension Event`, not the actual type's implementation.

To address this I made the JSON serialization protocol-based, thus enabling runtime dynamic dispatch for method calls, resulting in runtime polymorphism.

At I implemented the same approach as we have for Android, with `JSONConverter`, but that way we would lose the ability to detect missing event serialization during compilation when new events are introduced.

## Checklist
- [x] 🗒 `CHANGELOG` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixed
 
-- iOS: `onEvent` on iOS has partial payload information
+- iOS: `onEvent` on iOS has incomplete payload information
 
 ## [0.14.1] (2023-11-16)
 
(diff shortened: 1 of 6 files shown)


hawk23 (approved):
LGTM 🎉 🌮

PR #329, opened 2023-11-24
Fix Picture in Picture sample screen layout on tvOS

rolandkakonyi:
## Description

Found that on tvOS playback can not be controlled when using the Picture in Picture sample screen.
This is due to the player not taking the whole screen. This is a feature from Apple to not have controls for inline players shown.

## Changes

Replaced `SafeAreaView` for tvOS with `View` works now.

Note: I tried changing `edges` property to `[]` but that didn't work as expected on tvOS simulator.

|Before|After|
|---|---|
|Simulator Screenshot - Apple TV 17 0 - 2023-11-24 at 11 03 05|Simulator Screenshot - Apple TV 17 0 - 2023-11-24 at 11 01 08|

## Checklist
- [x] 🗒 `CHANGELOG` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixed
 
 - Android: `PlayerView` destroys attached `Player` instance on destroy. `Player` lifecycle must be handled on the creation side
+- tvOS: Picture in Picture sample screen has unwanted padding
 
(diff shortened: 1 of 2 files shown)


123mpozzi (approved):
Seems good!

PR #330, opened 2023-11-24
Player testing: host app setup

rolandkakonyi:
## Description

To enable running integration tests, we need an application to run on our supported platforms.

## Changes

Created a new application using the react native command line tools and manually adapted it to have the same extra changes as we have for the example app.
No other changes were made.
Also updated ignore files accordingly.

I went for creating a new application and keeping the example application untouched as I wanted to have the example application's changes and test host app changes separated.

Note: further changes to enable player testing come in upcoming PRs.

## Checklist
- [x] 🗒 `CHANGELOG` entry - not applicable, we should add a single entry once everything gets merged for integration testing


integration_test/src/App.tsx
rolandkakonyi, line 1: This is the default contents of the template application.
This is changed in #332

integration_test/README.md
matamegger, line 1: I guess this will then be filled with instructions on the e2e tests?

matamegger (approved):
Interestingly that we need another "app", but kinda makes sense given the setup RN is following.

integration_test/README.md
rolandkakonyi: Yes, I am working on that!

rolandkakonyi:
> Interestingly that we need another "app", but kinda makes sense given the setup RN is following.

@matamegger we don't "need", BUT I chose to, in order to avoid additional changes to the example app for the sake of testing. Especially as we reference it in our guides, it could confuse readers.

matamegger:
> > Interestingly that we need another "app", but kinda makes sense given the setup RN is following.
> 
> @matamegger we don't "need", BUT I chose to, in order to avoid additional changes to the example app for the sake of testing. Especially as we reference it in our guides, it could confuse readers.

I very much approve the separation of example and test setup here (well done 👍 ).
My interest was towards the need for an actual app with UI, instead of something more low profile like it can be done in the native android ecosystem for instance. But as said, it makes sense in the context of RN.

rolandkakonyi:
> > > Interestingly that we need another "app", but kinda makes sense given the setup RN is following.
> > 
> > 
> > @matamegger we don't "need", BUT I chose to, in order to avoid additional changes to the example app for the sake of testing. Especially as we reference it in our guides, it could confuse readers.
> 
> I very much approve the separation of example and test setup here (well done 👍 ). My interest was towards the need for an actual app with UI, instead of something more low profile like it can be done in the native android ecosystem for instance. But as said, it makes sense in the context of RN.

To be complete, I add more details here:
For Google IMA we need a UI stack, as the iOS Google IMA SDK needs a view to host the ad contents and UI. Additionally, DRM playback also requires a view stack where it renders, as we have seen failures in the past with DRM playback and no UI.

PR #331, opened 2023-11-24
Player testing: Cavy framework setup

rolandkakonyi:
## Description

In order to enable writing integration tests, we need a framework that can run tests, that runs them on the native platform (unlike Jest), and has access to code-level components, such as `Player`.
After an initial investigation, I found https://github.com/pixielabs/cavy.

Note: I noticed there were no commits in the last ~2 years, BUT no other solution I found had the characteristics we needed. At the same time, I familiarized myself with their codebase, which isn't too complex, nor too big, therefore forking and fixing issues, or implementing features shouldn't be too hard.
(I already opened a [PR](https://github.com/pixielabs/cavy/pull/350) for fixing a race condition)

## Changes

Integrated Cavy into the integration test application and set it up with a basic setup that can be extended to have player integration tests.

## Checklist
- [x] 🗒 `CHANGELOG` entry - not applicable, we should add a single entry once everything gets merged for integration testing


example/package.json
```diff
@@ -6,7 +6,10 @@
   "scripts": {
     "android": "react-native run-android",
     "ios": "react-native run-ios",
-    "start": "react-native start"
+    "start": "react-native start",
+    "pods": "yarn pods-install || yarn pods-update",
+    "pods-install": "yarn pod-install",
+    "pods-update": "pod update --silent"
```
rolandkakonyi, line 12: These were moved here from the root `package.json` to separate concerns

integration_test/babel.config.js
rolandkakonyi, line 1: Changes here mirror the one from the example app

integration_test/src/TestableApp.tsx
rolandkakonyi, line 1: This has a very basic setup which gets modified for actual player testing in #332

integration_test/tests/exampleSpec.ts
rolandkakonyi, line 1: This is a placeholder test to prove the framework works

.github/workflows/ci.yml
rolandkakonyi, line 1: We have to install yarn dependencies for the integration tests as well to allow typescript related validation.

matamegger (approved):
Nice! Works well.
I just happen to have metro already running (for the example though), but the error already told me what went wrong.

PR #337, opened 2023-11-28
Extend documentation for integration tests

rolandkakonyi:
## Description

There is not much documentation about player testing in the repo right now.

## Changes

- Updated `integration_test/README.md`
- Updated `CONTRIBUTING.md` with instructions for adding new tests
- Added API documentation for player testing functions.

## Checklist
- [x] 🗒 `CHANGELOG` entry

### DIFF
--- a/integration_test/playertesting/PlayerTesting.ts
+++ b/integration_test/playertesting/PlayerTesting.ts
@@ -53,8 +53,7 @@ export const callPlayer = async <T>(
 };
 
 /**
- * Calls the given function with the player instance and expects the given event to occur.
- * @param fn The function to call.
+ * Expects the given event to occur.
  * @param expectationConvertible The event to expect.
  * @param timeoutSeconds The number of seconds to wait for the event to occur.
  * @returns A promise that resolves when the function is finished.
@@ -79,8 +78,7 @@ export const expectEvent = async <T extends Event>(
 };
 
 /**
- * Calls the given function with the player instance and expects the given events to occur.
- * @param fn The function to call.
+ * Expects the given events to occur.
  * @param expectationsConvertible The events to expect.
  * @param timeoutSeconds The number of seconds to wait for the events to occur.
  * @returns A promise that resolves when the function is finished.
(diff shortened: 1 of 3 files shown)


integration_test/playertesting/PlayerTesting.ts
```diff
@@ -13,19 +13,61 @@ import {
...
+ * @see {@link Player}
+ */
 export const callPlayer = async <T>(
   fn: (player: Player) => Promise<T>
 ): Promise<T> => {
   return await PlayerWorld.shared.callPlayer(fn);
 };
 
+/**
+ * Calls the given function with the player instance and expects the given event to occur.
```
matamegger, line 56: ? which given function

> Listens for the specified `Event` to be emitted and waits until the event is received

integration_test/playertesting/PlayerTesting.ts
```diff
@@ -13,19 +13,61 @@ import {
...
+ * @param fn The function to call.
```
matamegger, line 57:
```suggestion
```

integration_test/playertesting/PlayerTesting.ts
```diff
@@ -36,6 +78,28 @@ export const expectEvent = async <T extends Event>(
   );
 };
 
+/**
+ * Calls the given function with the player instance and expects the given events to occur.
+ * @param fn The function to call.
```
matamegger, line 83: Same as with the single event expectation

integration_test/playertesting/PlayerTesting.ts
```diff
@@ -46,6 +110,22 @@ export const expectEvents = async (
   );
 };
 
+/**
+ * Calls the given function with the player instance and expects the given event to occur.
```
matamegger, line 114: > Starts listening for the specified `SingleEventExpectationz before executing the passed `fn` function.
This is the race-condition-safe version of calling `callPlayer` and `expectEvent` after that. Useful when events are directly tied to calls in the `fn` function block and therefore synchronously emitted.

Maybe more in that direction.

integration_test/playertesting/PlayerTesting.ts
rolandkakonyi: oopsie, fixed in a4169682

integration_test/playertesting/PlayerTesting.ts
rolandkakonyi: fixed in a4169682

integration_test/playertesting/PlayerTesting.ts
rolandkakonyi: fixed in a4169682

integration_test/playertesting/PlayerTesting.ts
rolandkakonyi: fixed in a4169682

matamegger (approved):

PR #338, opened 2023-11-29
Fix silent iOS playback when device is muted while running the example application

rolandkakonyi:
## Description

Some developers ran into the issue of having no audio during playback, due to their device being silenced.

## Changes

Setting audio session category on app level fixes this.

Documentation was updated on the following pages:
- https://developer.bitmovin.com/playback/docs/implementing-a-basic-player
- https://developer.bitmovin.com/playback/docs/react-native-how-to-let-audio-play-when-the-ios-device-is-in-silent-mode

## Checklist
- [x] 🗒 `CHANGELOG` entry - not applicable as no SDK changes were made

### DIFF
--- a/example/src/App.tsx
+++ b/example/src/App.tsx
@@ -1,9 +1,9 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import { Platform, Button } from 'react-native';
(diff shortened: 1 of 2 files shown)


123mpozzi (approved):
Working fine 👍 !!

PR #340, opened 2023-11-30
Add initial set of player tests

rolandkakonyi:
## Description

We have no meaningful tests for player behavior.

## Changes

Added 11 tests for basic player behavior testing.

## Checklist
- [x] 🗒 `CHANGELOG` entry - not applicable


integration_test/tests/exampleSpec.ts
rolandkakonyi, line 1: Removed this as we have actual tests now

integration_test/tests/playbackTest.ts
```diff
@@ -0,0 +1,61 @@
...
+    spec.it('emits a Paused event after calling pause API', async () => {
+      await startPlayerTest({}, async () => {
+        await loadSourceConfig(Sources.artOfMotionHls);
+        await playFor(1);
+        await callPlayerAndExpectEvent((player) => {
+          player.pause();
+        }, EventType.Paused);
+      });
+    });
+  });
```
matamegger, line 37: This should be two separate blocks I think.
```suggestion
  spec.describe('calling play when a source is loaded', () => {
    spec.it(
      'emits a Play and Playing events',
      async () => {
        await startPlayerTest({}, async () => {
          await loadSourceConfig(Sources.artOfMotionHls);
          await callPlayerAndExpectEvents((player) => {
            player.play();
          }, EventSequence(EventType.Play, EventType.Playing));
        });
      }
    );
  });
  spec.describe('calling pause when a source is loaded', () => {
    spec.it('emits a Paused event', async () => {
      await startPlayerTest({}, async () => {
        await loadSourceConfig(Sources.artOfMotionHls);
        await playFor(1);
        await callPlayerAndExpectEvent((player) => {
          player.pause();
        }, EventType.Paused);
      });
    });
  });
```

matamegger (approved):
Looks good. One small structural suggestion.

integration_test/tests/playbackTest.ts
rolandkakonyi: Applied feedback in 118b30c

PR #349, opened 2023-12-07
Add advertising related player tests

rolandkakonyi:
## Description

These changes add player integration tests to verify advertising-related events are emitted.

## Changes

Added advertising-related player integration tests.

## Checklist
- [x] 🗒 `CHANGELOG` entry - internal changes only, no changelog needed


integration_test/tests/advertisingTest.ts
```diff
@@ -0,0 +1,181 @@
...
+          schedule: adItems,
+        } as AdvertisingConfig;
+        await startPlayerTest({ advertisingConfig }, async () => {
+          await callPlayer(async (player) => {
+            player.load(Sources.artOfMotionHls);
+            player.play();
+          });
+          if (type !== 'progressive') {
+            await expectEvent(EventType.AdManifestLoaded, 20);
+          }
```
rolandkakonyi, line 62: I experienced longer waiting times with the Android emulator, therefore added this extra expectation to stabilize the test. APPLIES TO THE OTHER OCCURRENCES TOO

integration_test/tests/advertisingTest.ts
```diff
@@ -0,0 +1,181 @@
...
+            EventSequence(
+              EventType.AdBreakStarted,
+              EventType.AdStarted,
+              EventType.AdQuartile,
+              EventType.AdQuartile,
+              EventType.AdQuartile,
+              EventType.AdFinished,
+              EventType.AdBreakFinished
+            ),
+            18
```
zigavehovec, line 73: Is there a reason for this quite specific 18? Just a though, what if we increase the default timeout to account for Android emulators being a bit slower and set it to 20?

zigavehovec (approved):
LGTM!

integration_test/tests/advertisingTest.ts
rolandkakonyi: I would not touch the default timeout as usually 10 seconds is pretty long already.
This is specific for ads on the Android emulator so belongs to the specific test.

PR #353, opened 2023-12-13
Always run android integration tests on emulator

rolandkakonyi:
## Description

Currently when executing `yarn integration-test test:android` it runs on the first available device or emulator.
The integration tests can only run on emulators right now.

## Changes

Added a lookup and auto-start of emulator, to enable passing device ID to set target for running.
Starting the emulator is needed as otherwise we don't know the device ID for it.

## Checklist
- [x] 🗒 `CHANGELOG` entry - internal change, no entry needed


integration_test/package.json
```diff
@@ -6,9 +6,9 @@
...
-    "stop-test:ios": "xcrun simctl list devices available -e -j | jq --raw-output '.devices.[].[] | select( .state == \"Booted\" and (.deviceTypeIdentifier | contains(\"iPhone\"))).udid' | xargs -I UDID xcrun simctl terminate UDID com.bitmovin.player.reactnative.IntegrationTest 2> /dev/null; exit 0",
-    "start-test:android": "yarn cavy run-android",
+    "stop-test:android": "adb shell am force-stop com.bitmovin.player.reactnative.integrationtest; true",
+    "stop-test:ios": "xcrun simctl list devices available -e -j | jq --raw-output '.devices.[].[] | select( .state == \"Booted\" and (.deviceTypeIdentifier | contains(\"iPhone\"))).udid' | xargs -I UDID xcrun simctl terminate UDID com.bitmovin.player.reactnative.IntegrationTest 2> /dev/null; true",
+    "start-test:android": "yarn cavy run-android --deviceId \"$([ -z \"$(adb devices | grep -v List | grep device | grep -m 1 emulator | cut -f 1)\" ] && (emulator -list-avds | head -n 1 | xargs -I NAME emulator -avd NAME 1> /dev/null 2> /dev/null & adb wait-for-device shell 'while [[ -z $(getprop sys.boot_completed) ]]; do sleep 1; done;'); (adb devices | grep -v List | grep device | grep -m 1 emulator | cut -f 1))\"",
```
rolandkakonyi, line 11: This checks if there is an emulator running, if not starts one, then runs on the running emulator.

integration_test/package.json
```diff
@@ -6,9 +6,9 @@
     "android": "react-native run-android",
     "ios": "react-native run-ios",
     "start": "react-native start",
-    "stop-test:android": "adb shell am force-stop com.bitmovin.player.reactnative.integrationtest",
-    "stop-test:ios": "xcrun simctl list devices available -e -j | jq --raw-output '.devices.[].[] | select( .state == \"Booted\" and (.deviceTypeIdentifier | contains(\"iPhone\"))).udid' | xargs -I UDID xcrun simctl terminate UDID com.bitmovin.player.reactnative.IntegrationTest 2> /dev/null; exit 0",
-    "start-test:android": "yarn cavy run-android",
+    "stop-test:android": "adb shell am force-stop com.bitmovin.player.reactnative.integrationtest; true",
```
zigavehovec, line 9: What does the "true" do?

integration_test/package.json
zigavehovec: Quite a script, would it be possible to extract it into a command with a descriptive name by chance? Something like: "run-on-active-emulator"?

rolandkakonyi: I wanted to keep the CLI interface minimal, I don't want to have `start-test:ios` and `start-test:android-run-on-active-emulator`.
Wanted to keep it consistent.
Once the limitation is lifted we can remove this or extract IMO

integration_test/package.json
rolandkakonyi: It returns with success exit code even if the previous command failed. E.g. if no emulator was running.

integration_test/package.json
zigavehovec: Makes sense!

zigavehovec (approved):
LGTM!

CONTRIBUTING.md
```diff
@@ -205,7 +205,7 @@ The `package.json` file contains various scripts for common tasks:
 - `yarn example android`: run the example app on Android.
 - `yarn example pods`: install pods only.
 - `yarn integration-test start`: start the Metro server for the integration tests.
-- `yarn integration-test test:android`: run the player tests on Android.
-- `yarn integration-test test:ios`: run the player tests on iOS.
+- `yarn integration-test test:android`: run the player tests on Android emulator.
+- `yarn integration-test test:ios`: run the player tests on iOS simulator.
```
rolandkakonyi, line 209: @zigavehovec I updated the documentation here!

integration_test/README.md
```diff
@@ -24,4 +24,4 @@ yarn integration-test test:android # Run tests on Android emulator
 
 To set the license key to be used for the tests, you can set the key `"licenseKey"` in `integration_test/app.json`.
 
-Note: The tests are currently only supported on iOS simulators and Android emulators. Running them on real devices is not supported at the moment.
+Note: The tests are currently only supported on iOS simulators and Android emulators. Running them on real devices is not supported at the moment because the testing framework has hard-coded `localhost` as the server address.
```
rolandkakonyi, line 27: @zigavehovec I updated the documentation here!

PR #356, opened 2023-12-14
Showcase ambient mode / screen timeout handling

zigavehovec:
## Description

Default samples go into ambient mode / screen timeout due to screen inactivity. Since the handling should be implemented on the app to app basis, this PR just adds an example and reference on how this can be handled.

## Changes

- Add `getWindow().addFlags(FLAG_KEEP_SCREEN_ON);` to `MainActivity.java`

## Checklist
- [x] 🗒 `CHANGELOG` entry - No changelog as only example code was updated


strangesource (approved):
Nice! Do you think this is worth a changelog entry? I am unsure.

zigavehovec:
> Nice! Do you think this is worth a changelog entry? I am unsure.

I thought about it but I don't think since it's not something that's going to be released.

PR #368, opened 2023-12-19
Simplify iOS development team setup

rolandkakonyi:
## Description

Currently when trying to run the iOS or tvOS example application on a physical device, it ends up with an error requiring an Apple Development Team to be set in the project file.
This is a change that should not be committed into the repository as the development team can be unique to each contributor.
This causes unnecessary overhead with change management.
This PR changes that to simplify the handling of this via an optional `Developer.xcconfig` file.

## Changes

1. Introduced specific build configuration files (`.xcconfig`) for iOS and tvOS per config (`Debug`/`Release`) and configured to optionally import the `Developer.xcconfig` file.
[Same way as in our Flutter Player SDK](https://github.com/bitmovin/bitmovin-player-flutter/blob/main/CONTRIBUTING.md#for-ios-development).
2. Updated `CONTRIBUTING.md` with the instructions for setup.
3. Applied workaround for Resource Bundle targets from the Pods project to avoid the need of a development team there.

NOTE: if you already have the project checked out, you need to run `yarn example pods` for the Pods project workaround to take effect.

## Checklist
- [x] 🗒 `CHANGELOG` entry - not applicable, only example project and documentation changes


example/ios/Podfile
```diff
@@ -105,3 +106,14 @@ def fix_deployment_target(installer)
     end
   end
 end
+
+def disable_resource_bundle_signing(installer)
+  installer.pods_project.targets.each do |target|
+    target_is_resource_bundle = target.respond_to?(:product_type) && target.product_type == 'com.apple.product-type.bundle'
+    target.build_configurations.each do |config|
+      if target_is_resource_bundle
+        config.build_settings['CODE_SIGNING_ALLOWED'] = 'NO'
```
rolandkakonyi, line 115: This is needed to workaround the required code-signing for resource bundle targets.

Initially I wanted to set the team properly but it is not trivial to get the value out of the `Developer.xcconfig` file.
This is no problem as long as we don't want to submit the example application to the App Store (it will never happen :) )

zigavehovec: Thanks for the comment. Might be worth add it to the code also 🙂

zigavehovec (approved):
LGTM!

example/ios/Podfile
rolandkakonyi: That's a great idea, added comment for all workaround here: ed4dfa5

PR #372, opened 2024-01-09
boost install workaround

dweinber:
## Description
Since 6th January 2024, installing dependencies failed due to `boost`, which is used by `react-native`.
See issues https://github.com/boostorg/boost/issues/843 and https://github.com/facebook/react-native/issues/42180 for details, the second issue contains the official react-native workaround proposal.

Note: A fix should be available in `react-native` version `0.72.9` which should be an easy upgrade (only the RN package version changed according to the [RN Upgrade Helper](https://react-native-community.github.io/upgrade-helper/?from=0.72.6&to=0.72.9) and wouldn't require the workaround. However, we're using also `react-native-tvos` which hasn't released a fix yet so we anyway need the workaround for now.

## Changes
After installing the dependencies from npm, the new patch files will be applied to the installed `react-native` npm package. The patch files contain updated download locations for `boost`.

## Checklist
- [ ] 🗒 `CHANGELOG` entry


rolandkakonyi (approved):
All green! Nice job! Thanks for the effort!

PR #388, opened 2024-01-26
Add SRT (SubRip) subtitle support

rolandkakonyi:
## Description

This PR adds support for SubRip (SRT) subtitle support for side-loaded subtitles.
In #387 the build broke due to the iOS Player update changes with new enum values.

## Changes

Added support for SRT subtitles by mapping the relevant enum cases and mime types accordingly.

## Checklist
- [x] 🗒 `CHANGELOG` entry

### DIFF
--- a/android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
@@ -530,7 +530,12 @@ fun ReadableMap.toSubtitleTrack(): SubtitleTrack? {
 /**
  * Converts any subtitle format name in its mime type representation.
  */
-private fun String.toSubtitleMimeType(): String = "text/$this"
+private fun String.toSubtitleMimeType(): String = when (this) {
+    "vtt" -> "text/vtt"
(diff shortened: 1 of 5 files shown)


android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
```diff
@@ -530,7 +530,12 @@ fun ReadableMap.toSubtitleTrack(): SubtitleTrack? {
 /**
  * Converts any subtitle format name in its mime type representation.
  */
-private fun String.toSubtitleMimeType(): String = "text/$this"
+private fun String.toSubtitleMimeType(): String = when (this) {
+    "vtt" -> "text/vtt"
```
strangesource, line 534: not needed, same as the else branch, right?

strangesource (approved):
Nice, just some nit.

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
rolandkakonyi: True, fixed in cc2198e

PR #404, opened 2024-03-01
Update react native to 0.73.4

matamegger:
## Description

To unblock player updates we need to upgrade the react native version. The react native version update also comes with required upgrades of the tooling e.g. jdk, node updates.

## Changes

- Update GH workflows to set up the according jdk, node and xcode version
- Update to react native `0.73.4` as well as to `0.73.4-0` of the tvos fork
- Remove flipper integration (not needed by us, and there has been a change in the framework setup)
- Refactored java app entry points to kotlin (applied from the official template)
- Updated the gradle setup as well as the build, compile, target and ndk version
- Removed boost "jfrog" workaround
- example: Updated react native navigation, orientation locker, safe are context, native screens, system navigation bar
- example: fixed (now facing) type issues
- Adjusted the ios project setup in alignment with the changes in react native upgrade helper tool

## Checklist
- [x] 🗒 `CHANGELOG` entry


matamegger:
@zigavehovec @rolandkakonyi feel free to redistribute the review. I thought you would be the best match for this kind of PR :)

krocard (approved):
Very nice :+1:

rolandkakonyi:
I will keep updating this comment as this will take some time 😅

Integration tests: passed ✅ 
Manual testing: in progress ⏳
Code review: not started yet

.github/workflows/ci-ios-tvos.yml
rolandkakonyi, line 1: I can't remember exactly if the Xcode version change was the fix we needed, or something else.
Do you remember if this is needed?
I would not have the Xcode version updated unless needed to keep it in sync with our min. supported Xcode version as on the iOS Player SDK.

matamegger: Good point! I will double check.

zigavehovec (approved):
The changes look good 👍
Manual testing for Android looks good apart from the offline feature, where  RuntimeException is thrown when trying to start the download: 
```
Caused by: java.lang.SecurityException: Starting FGS with type dataSync callerApp=ProcessRecord{f884966 14500:com.bitmovin.player.reactnative.example/u0a326} targetSDK=34 requires permissions: all of the permissions allOf=true [android.permission.FOREGROUND_SERVICE_DATA_SYNC]
```
This is unrelated to the changes in this PR but should be very easy to fix (just adding the FOREGROUND_SERVICE_DATA_SYNC permission in the example manifest). Could you take care of this?

matamegger:
> 2. Landscape Fullscreen Handling on Android and iOS has the wrong initial state. It goes full screen but portrait mode. On iOS then it gets stuck as landscape, even staying like that for the list screen.

This seems to be the case before this PR already. Not going to fix this within this PR.

PR #406, opened 2024-03-01
Support `CueEnter` and `CueExit` events

123mpozzi:
## Description

Implement two **subtitle cue events** and their relative listeners:
- `CueEnter` - Fired when a subtitle entry transitions into the active status
- `CueExit` - Fired when an active subtitle entry transitions into the inactive status

They currently support the following properties:
- `start` - The playback time in seconds when the subtitle should be rendered
- `end` - The playback time in seconds when the subtitle should be hidden
- `text` (_Optional_) - The textual content of the subtitle

## Changes

- Add support for `CueEnterEvent` and `CueExitEvent`
- Add related listeners: `onCueEnter`, and `onCueExit`
- Update the demo page _Subtitle and captions_ to listen to the newly added events

## Testing

- Cover the events being fired in the correct order in integration tests
- Add Sintel to Sources

### How to Manual test
- Run the page `Examples > Subtitle and captions`
- Sets subtitles to "_Custom English (WebVTT)_" as its cues also feature the `end` property
  > But often the tracks do not have the `end` property (maybe the native player is not communicating it to us?). In all cases, it should not be a concern of this PR as the parameters serialization seems to be working correctly
- Seek to somewhere around `2:10` to find some dialogue
- Look at the logs on Metro: it should listen successfully to the added events

## Checklist
- [x] 🗒 `CHANGELOG` entry


android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
```diff
@@ -470,6 +470,18 @@ fun PlayerEvent.toJson(): WritableMap {
...
+            json.putDouble("start", start)
+            json.putDouble("end", end)
+            json.putString("text", text)
+        }
+
+        is PlayerEvent.CueExit -> {
+            json.putDouble("start", start)
+            json.putDouble("end", end)
+            json.putString("text", text)
+        }
```
strangesource, line 483: Does it make sense to relay all cue events even if they don't make sense with the supported content? E.g. bitmap subtitles will not have any useful content for customers without the image IMO.

src/events.ts
```diff
@@ -735,3 +735,39 @@ export interface PlaybackSpeedChangedEvent extends Event {
...
+   */
+  start: number;
+  /**
+   * The playback time in seconds when the subtitle should be hidden.
+   */
+  end: number;
+  /**
+   * The textual content of this subtitle.
+   */
+  text?: string;
```
strangesource, line 754: Does it make sense to have the text as optional? What is the use case for a cue without text?

strangesource (commented):
Looking good in general, just some questions.

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
123mpozzi: That's a valid point, thank you for bringing it up. 👍    
But the idea for this PR was to implement the same baseline API we did for the Flutter SDK. So supporting other `Cue` fields would be out of scope here.  
Filtering the cue events which are not supported yet could be object of discussion, but I think the plan for now is to keep it simple.  

Let me know however if you think we should go that direction with this PR 🙌

rolandkakonyi (approved):
Nice job! 🎉

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
strangesource: > supporting other Cue fields would be out of scope here.
Makes sense. 👍 

> the plan for now is to keep it simple.

Fine for me. Should we add some documentation around what is supported and what now? I still think it is confusing to get `Cue`s without any useful information.

123mpozzi: Nice idea. I added a note in ca321df922423048775deb723bc9b567e9f6ba7a

strangesource (approved):

PR #408, opened 2024-03-05
Extract environment setup for the CI into separate action

matamegger:
## Description

There is a lot of redundancy in the CI workflows regarding the setup of the tooling. Which is especially noticeable when updating or changing tooling setups and versions.

## Changes

Extract the setup into a configurable action and replace the setup with the newly introduced action.

## Checklist
- [x] 🗒 `CHANGELOG` entry

### DIFF
--- a/.github/actions/setup-build/action.yml
+++ b/.github/actions/setup-build/action.yml
@@ -1,0 +1,85 @@
+name: Setup build environment
+description: Setup the build environment
+
+inputs:
+  node:
+    description: Setup node and install root node modules
+    default: false
+    required: false
+  subprojects:
+    description: Setup node and install root and subproject node modules (overrides `node` input)
+    default: false
+    required: false
+  ios-build:
+    description: Setup xcode and install dependencies
+    default: false
+    required: false
+  restore-pods:
+    description: Restores the Pod cache
+    default: false
+    required: false
+  java:
+    description: Setup Java and Gradle cache
+    default: false
+    required: false
+
+runs:
+  using: composite
+  steps:
+    - name: Set up Java
+      if: ${{ inputs.java == 'true' }}
+      uses: actions/setup-java@v3
+      with:
+        distribution: 'zulu'
+        java-version: '17'
+
+    - name: Set up Gradle cache
+      if: ${{ inputs.java == 'true' }}
+      uses: gradle/gradle-build-action@v2
+      with:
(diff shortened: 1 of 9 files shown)


.github/actions/setup-build/action.yml
```diff
@@ -0,0 +1,85 @@
...
+        registry-url: 'https://registry.npmjs.org/'
+        cache: 'yarn'
+
+    - name: Install node_modules
+      if: ${{ inputs.node == 'true' || inputs.subprojects == 'true' }}
+      shell: bash
+      run: yarn install --frozen-lockfile
+
+    - name: Install node_modules (example/)
+      if: ${{ inputs.subprojects == 'true' || inputs.subprojects == 'true' }}
```
rolandkakonyi, line 56: Is this a typo? `inputs.subprojects == 'true'` is there 2x!

.github/actions/setup-build/action.yml
rolandkakonyi, line 1: Could we call this `prepare-dependencies`?
`setup-build` is a bit unclear to me if it actually builds or not.

matamegger: "dependencies" is hear a bit broadly used IMO. Installing xcode or java is strictly preparing a dependency, but not what we usually talk about when talking about dependencies I think.

`prepare-environment` or `setup-environment`?

rolandkakonyi: It is a dependency for the upcoming build process. We use the term in the iOS SDK, but I like both `prepare-environment` and `setup-environment`. Either is fine!

.github/actions/setup-build/action.yml
matamegger: Embarrassing. Seems to be a copy past error on my side. should be just once there.

.github/actions/setup-build/action.yml
matamegger: I see. I guess `environment` is a good compromise then :)

matamegger: https://github.com/bitmovin/bitmovin-player-react-native/pull/408/commits/e25bdbf939ee04d175fbeea0abe4c22b680f4a90

.github/actions/setup-build/action.yml
matamegger: https://github.com/bitmovin/bitmovin-player-react-native/pull/408/commits/443a4be8423bfb1c314ceef4760ff22aa6502365

PR #412, opened 2024-03-19
Fix event listener conflicts

123mpozzi:
## Description

The names we are using for events are too generic (eg. `onSeek`): this may cause conflicts with other 3rd party libraries in case both are included in the same project.

> **Note**
> The Android part did not have conflict issues. However, the bubbled events have been renamed also there to keep them synced with the iOS part, to avoid future confusion.

## Changes

- Rename the bubbled listeners which are communicating with our native SDKs using a `onBmp` prefix instead of the generic `on` prefix
  > I decided to keep using a prefix starting with `on`(instead of eg. `bmpOn`) because in iOS [RCTBubblingEventBlock must be prefixed with on](https://reactnative.dev/docs/native-components-ios#events).
- Create a copy of `events.ts` named `nativeEvents.ts` which contains the bubbled event listeners
  > The original file now contains the un-bubbled event listeners together with the un-bubbling function

## Testing

Run integration tests to check if events are working correctly

## Checklist
- [x] 🗒 `CHANGELOG` entry


matamegger (commented):
Android side looks fine!

Have you also done some manual expectation of the sample apps on iOS and android? :)
Do all of them still work?

123mpozzi:
> Android side looks fine!
> 
> Have you also done some manual expectation of the sample apps on iOS and android? :) Do all of them still work?

Yes 👍 , I tested all the samples apart from casting and offline

matamegger (approved):

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 - `CueEnterEvent` and `CueExitEvent` to signal when a subtitle entry transitions into an active or inactive status respectively
 
+### Fixed
+
+- Use `onBmp` as a prefix to registered event listeners to prevent event name conflicts with other 3rd party libraries
```
rolandkakonyi, line 11: I would not go into that much detail but say something like this instead:
```suggestion
- Fix potential event name conflicts with other 3rd party libraries
```

rolandkakonyi (commented):
Changes look good!
Manual testing looks good as well!

Please delete the testing related branches once merging this!

123mpozzi:
> Changes look good! Manual testing looks good as well!
> 
> Please delete the testing related branches once merging this!

Sure, I will 👍

rolandkakonyi (approved):

PR #414, opened 2024-03-22
Remove commitlint

123mpozzi:
## Description

This PR is a proposal to remove `commitlint` from this project.

In my opinion, it has more drawbacks than advantages. The main ones being the following:
- The time spent on formatting commit messages, albeit it is little time, could be spent somewhere else to achieve greater impact.
- For casual contributors, which are not working full-time on this repository, it can be challenging to always remember this additional barrier, causing the initial commits to be annoyingly rejected.
- Since it can be annoying, people may be tempted at skipping hooks altogether, risking to skip meaningful checks.

#### Remove
- `commitlint` via `yarn`
- `commit-msg` hook from `husky`
- Section _Commit message convention_ from README

## Testing
- Make some changes (_but keep in mind that the code-linting hook is still there, so follow proper code style if the changes are on the code itself_)
- Do a testing commit with a message that previously would have been rejected: it should now be successful

## Checklist
- [x] 🗒 `CHANGELOG` entry: None as this is internal

### DIFF
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -1,4 +1,0 @@
-#!/bin/sh
-. "$(dirname "$0")/_/husky.sh"
-
-yarn commitlint --edit $1
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -183,19 +183,6 @@ export default (spec: TestScope) => {
 };
 ```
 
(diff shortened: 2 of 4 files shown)


rolandkakonyi (approved):
😍

PR #419, opened 2024-03-29
Expose hideFirstFrame config

dweinber:
## Description
Our native SDKs contains a configuration to ensure the first frame of the main content is not shown, which is useful for pre-roll ads in some scenarios. This configuration was not exposed to the React Native app.

## Changes
Expose `PlayerViewConfig.hideFirstFrame`, which maps to
- `PlayerViewConfig.hideFirstFrame` on Android, and
- `BitmovinUserInterfaceConfig.hideFirstFrame` and `SystemUserInterfaceConfig.hideFirstFrame` on iOS/tvOS.

## Checklist
- [x] 🗒 `CHANGELOG` entry

### DIFF
--- a/android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
@@ -750,6 +750,9 @@ fun toPlayerViewConfig(json: ReadableMap) = PlayerViewConfig(
             ?.getBooleanOrNull("playbackSpeedSelectionEnabled")
             ?: true,
     ),
+    hideFirstFrame = json.getMap("hideFirstFrame")
+        ?.getBooleanOrNull("hideFirstFrame")
+        ?: false,
 )
 
 private fun ReadableMap.toUserInterfaceTypeFromPlayerConfig(): UserInterfaceType? =
--- a/android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
@@ -751,9 +751,7 @@ fun toPlayerViewConfig(json: ReadableMap) = PlayerViewConfig(
             ?.getBooleanOrNull("playbackSpeedSelectionEnabled")
             ?: true,
     ),
-    hideFirstFrame = json.getMap("hideFirstFrame")
-        ?.getBooleanOrNull("hideFirstFrame")
-        ?: false,
+    hideFirstFrame = json.getBooleanOrNull("hideFirstFrame") ?: false,
 )
 
 private fun ReadableMap.toUserInterfaceTypeFromPlayerConfig(): UserInterfaceType? =
--- a/ios/RCTConvert+BitmovinPlayer.swift
+++ b/ios/RCTConvert+BitmovinPlayer.swift
@@ -1157,7 +1157,7 @@ extension RCTConvert {
         return RNPlayerViewConfig(
             uiConfig: rnUiConfig(json["uiConfig"]),
             pictureInPictureConfig: pictureInPictureConfig(json["pictureInPictureConfig"]),
-            hideFirstFrame: json["hideFirstFrame"] as? Bool ?? false
+            hideFirstFrame: json["hideFirstFrame"] as? Bool
         )
     }
 
--- a/ios/RCTConvert+BitmovinPlayer.swift
+++ b/ios/RCTConvert+BitmovinPlayer.swift
@@ -1265,7 +1265,7 @@ internal struct RNPlayerViewConfig {
      * To reliably hide the first frame before a pre-roll ad, please ensure that you are using the
      * `AdvertisingConfig` to schedule ads and not the `scheduleAd` API call.
      */
-    var hideFirstFrame: Bool
+    var hideFirstFrame: Bool?
 }
 
 /**
--- a/ios/RCTConvert+BitmovinPlayer.swift
+++ b/ios/RCTConvert+BitmovinPlayer.swift
@@ -1156,7 +1156,8 @@ extension RCTConvert {
 
         return RNPlayerViewConfig(
             uiConfig: rnUiConfig(json["uiConfig"]),
-            pictureInPictureConfig: pictureInPictureConfig(json["pictureInPictureConfig"])
+            pictureInPictureConfig: pictureInPictureConfig(json["pictureInPictureConfig"]),
+            hideFirstFrame: json["hideFirstFrame"] as? Bool ?? false
         )
     }
 
@@ -1256,6 +1257,15 @@ internal struct RNPlayerViewConfig {
         }
         return config
     }
+
+    /**
+     * When set to `true` the first frame of the main content will not be rendered before playback starts.
+     * Default is `false`.
+     *
+     * To reliably hide the first frame before a pre-roll ad, please ensure that you are using the
+     * `AdvertisingConfig` to schedule ads and not the `scheduleAd` API call.
+     */
+    var hideFirstFrame: Bool
 }
 
 /**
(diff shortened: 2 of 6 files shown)


android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
```diff
@@ -751,6 +751,9 @@ fun toPlayerViewConfig(json: ReadableMap) = PlayerViewConfig(
             ?.getBooleanOrNull("playbackSpeedSelectionEnabled")
             ?: true,
     ),
+    hideFirstFrame = json.getMap("hideFirstFrame")
```
rolandkakonyi, line 754: Should that first `getMap` be with `uiConfig`?

Note: I didn't run this yet but based on the code above that would make sense.

```suggestion
    hideFirstFrame = json.getMap("uiConfig")
```

ios/RCTConvert+BitmovinPlayer.swift
```diff
@@ -1156,7 +1156,8 @@ extension RCTConvert {
 
         return RNPlayerViewConfig(
             uiConfig: rnUiConfig(json["uiConfig"]),
-            pictureInPictureConfig: pictureInPictureConfig(json["pictureInPictureConfig"])
+            pictureInPictureConfig: pictureInPictureConfig(json["pictureInPictureConfig"]),
+            hideFirstFrame: json["hideFirstFrame"] as? Bool ?? false
```
rolandkakonyi, line 1160: I would make `hideFirstFrame` as an optional bool (`Bool?`), so we can indicate we want the default SDK value, indicated by `nil`.
This way we don't have to sync the default value both here in RN and on the native SDK.

That way we would only modify this flag when we have a known value.

What do you think?

src/components/PlayerView/playerViewConfig.ts
```diff
@@ -19,6 +19,13 @@ export interface PlayerViewConfig {
    * Provides options to configure Picture in Picture playback.
    */
   pictureInPictureConfig?: PictureInPictureConfig;
+
+  /**
+   * When set to `true`, the first frame of the main content will not be rendered before playback starts. Default is `false`.
```
rolandkakonyi, line 24: Please mention that this will have no effect with `UserInterfaceType.Subtitle`, at least on iOS.

ios/RNPlayerViewManager.swift
```diff
@@ -74,6 +62,42 @@ public class RNPlayerViewManager: RCTViewManager {
...
+#if os(iOS)
+        if styleConfig.userInterfaceType == .bitmovin {
+            let bitmovinUserInterfaceConfig = styleConfig
+                .userInterfaceConfig as? BitmovinUserInterfaceConfig ?? BitmovinUserInterfaceConfig()
+
+            if let uiConfig = playerViewConfig?.uiConfig {
+                bitmovinUserInterfaceConfig
+                    .playbackSpeedSelectionEnabled = uiConfig.playbackSpeedSelectionEnabled
+            }
+            bitmovinUserInterfaceConfig.hideFirstFrame = playerViewConfig?.hideFirstFrame ?? false
```
rolandkakonyi, line 78: In case we decide changing `hideFirstFrame` to optional.

```suggestion
            if let hideFirstFrame = playerViewConfig?.hideFirstFrame {
                bitmovinUserInterfaceConfig.hideFirstFrame = hideFirstFrame
            }
```

ios/RNPlayerViewManager.swift
```diff
@@ -74,6 +62,42 @@ public class RNPlayerViewManager: RCTViewManager {
...
+            }
+
+            return bitmovinUserInterfaceConfig
+        }
+#endif
+        if styleConfig.userInterfaceType == .system {
+          let systemUserInterfaceConfig = styleConfig
+                .userInterfaceConfig as? SystemUserInterfaceConfig ?? SystemUserInterfaceConfig()
+
+          systemUserInterfaceConfig.hideFirstFrame = playerViewConfig?.hideFirstFrame ?? false
```
rolandkakonyi, line 93: same here

src/components/PlayerView/playerViewConfig.ts
dweinber: Added in d63c39cd.

Android doesn't mention any restrictions etc in the API doc, plus for Android it's on the `PlayerViewConfig`, so I assume no such limitation exists there. But would be good to get confirmation from the Android team before marking this conversation as resolved.

ios/RNPlayerViewManager.swift
dweinber: Changed in 758e17b7

ios/RCTConvert+BitmovinPlayer.swift
dweinber: Makes definitely sense. Done in c833751b (although I'm not 100% sure that's how it's done in Swift, but it seems to work)/

src/components/PlayerView/playerViewConfig.ts
```diff
@@ -19,6 +19,14 @@ export interface PlayerViewConfig {
...
+   * This configuration has no effect for the ${@link UserInterfaceType.Subtitle} on iOS/tvOS.
```
rolandkakonyi, line 25: I think there is an extra `$` there
```suggestion
   * This configuration has no effect for the {@link UserInterfaceType.Subtitle} on iOS/tvOS.
```

src/components/PlayerView/playerViewConfig.ts
rolandkakonyi: @matamegger could you share some insights?

matamegger: The property should work for all UI types of the android player :)

rolandkakonyi: Actualy here is no good reason to not have it on iOS, I think it was just never requested.
@dweinber It would be a quite simple change to support it if needed.

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
dweinber: I think you're right that `json.getMap("hideFirstFrame")` is not correct, thanks for pointing out. However, changing it to your suggestion would mean that the `hideFirstFrame` flag is expected in `PlayerViewConfig.uiConfig` but it is in `PlayerViewConfig.hideFirstFrame`.

I think `hideFirstFrame = json.getBooleanOrNull("hideFirstFrame"),` would be the correct approach, but that doesn't work as I get a `Boolean?` but `PlayerViewConfig` expects a `Boolean`. Based on the conversation in the iOS part of this PR, we don't necessarily want to add the default value in RN, so I'm wondering what's the best approach for this. Maybe @matamegger has a suggestion?

src/components/PlayerView/playerViewConfig.ts
dweinber: Good to know, not needed at the moment but would be nice to have 🙂

rolandkakonyi (commented):
@dweinber could you enable the new flag in `BasicAds` for showcasing it?

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
rolandkakonyi: My bad for the wrong suggestion!
I see we already have some default values in there, which I missed. So no strong opinion about the default value topic.

dweinber:
Good idea @rolandkakonyi, done in 055f2458

rolandkakonyi (approved):
Changes and manual testing looks good from the iOS and TypeScript side!

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
dweinber: Makes all sense. I updated this in a108c8cd now.

matamegger (approved):

PR #424, opened 2024-04-04
Fix PiP inconsistencies on Android

zigavehovec:
## Description

The PiP implementation on Android was very complex and made some wrong assumptions, which caused a few bugs. Most notably: 
- Auto PiP mode entering was always enabled, which caused the app to go into PiP mode on every screen when navigating the app to the background.
Note that we do have `PictureInPictureConfig.shouldEnterOnBackground`, but it works on iOS only and implementing it carries some lifecycle handling complexity (this bug), so implementing this should be considered as a feature request. 
Reference: https://github.com/bitmovin/bitmovin-player-react-native/pull/424/files#diff-884e35b040de326e767bfbc98ce73c052358fe2319d57d4716e044dbe68b6aa1L176
- `RNPictureInPictureHandler` did toolbar hiding from the native side which is a wrong layer to do that. This caused the app to be without a toolbar after: Going into PiP mode -> Dismissing PiP window (stopping the app) -> Opening the app again. 
Reference: https://github.com/bitmovin/bitmovin-player-react-native/pull/424/files#diff-884e35b040de326e767bfbc98ce73c052358fe2319d57d4716e044dbe68b6aa1L115

## Changes

- Greatly simplify `RNPictureInPictureHandler`, aligning it with the Flutter implementation: https://github.com/bitmovin/bitmovin-player-flutter/pull/87
  - Fix unwanted auto entering into PiP mode
  - Fix toolbar gone after PiP mode has been dismissed and app opened again

## Checklist
- [x] 🗒 `CHANGELOG` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## [unreleased]
+## [Unreleased]
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixed
 
-- Android: Entering Picture-in-Picture automatically from all screens when navigating the app to the background (after navigating to Basic Picture-in-Picture screen once)
+- Android: Entering Picture-in-Picture automatically when navigating the app to the background after activating Picture-in-Picture mode once
 - Android: Example app Toolbar not visible after going into PiP mode -> Dismissing PiP window (stopping the app) -> Opening the app again
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,8 @@
 
 ### Fixed
 
-- Android: Entering Picture-in-Picture automatically from all screens when navigating the app to the background (after navigating to Picture-in-Picture sample once)
-- Android: Example app Toolbar not visible after going into PiP mode -> Dismissing PiP window (stopping the app) -> Opening the app again+
+- Android: Entering Picture-in-Picture automatically from all screens when navigating the app to the background (after navigating to Basic Picture-in-Picture screen once)
+- Android: Example app Toolbar not visible after going into PiP mode -> Dismissing PiP window (stopping the app) -> Opening the app again
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+## [unreleased]
+
+### Fixed
+
+- Android: Entering Picture-in-Picture automatically from all screens when navigating the app to the background (after navigating to Picture-in-Picture sample once)
+- Android: Example app Toolbar not visible after going into PiP mode -> Dismissing PiP window (stopping the app) -> Opening the app again+
+
+### Changed
+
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 # Changelog
 
+## [unreleased]
```
strangesource, line 3:
```suggestion
## [Unreleased]
```
NIT

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
...
+
+### Fixed
+
+- Android: Entering Picture-in-Picture automatically from all screens when navigating the app to the background (after navigating to Basic Picture-in-Picture screen once)
```
strangesource, line 7: This entry is very specific for our samples, is this something that only happens in our samples? If so we should make this clear. Otherwise, can we make it a bit more generic?

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -232,61 +218,38 @@ class RNPlayerView(
...
-    }
-
-    /**
-     * Called when the player has just exited PiP mode.
-     */
-    override fun onExitPictureInPicture() {
-        // Explicitly call `exitPictureInPicture()` on PlayerView when exiting PiP state, otherwise
-        // the `PictureInPictureExit` event won't get dispatched.
-        playerView?.exitPictureInPicture()
+    private fun isInPictureInPictureMode(): Boolean {
```
strangesource, line 221: This function has a very similar name to the property `isPictureInPictureMode` which makes it quite confusing. Should we rename to e.g. `isCurrentActivityInPictureInPictureMode ` or convert it to an extension function on the `activity`?

strangesource (approved):
Very nice simplifications/improvements! Just some NIT from my side.

CHANGELOG.md
zigavehovec: It's kind of both, I would suppose that it should happen to other single activity apps. I'll write it in a more generic way.

zigavehovec: https://github.com/bitmovin/bitmovin-player-react-native/pull/424/commits/132dbd68bb0f1cb2e3b5105ecdf45e5362088f59

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
zigavehovec: https://github.com/bitmovin/bitmovin-player-react-native/pull/424/commits/d6e1247212145546f65b7f5bbc8b3ea54568bcc1

rolandkakonyi (commented):
iOS still works :)

I can't judge the Android part.

Could you also update the guide here?
https://developer.bitmovin.com/playback/docs/enabling-picture-in-picture-mode

rolandkakonyi (approved):
Approving but please still update the guide!

zigavehovec:
> Approving but please still update the guide!

Added a section in the guide here: https://developer.bitmovin.com/playback/docs/enabling-picture-in-picture-mode#adjust-the-ui-for-picture-in-picture-mode-on-android

PR #432, opened 2024-04-11
Improve `TweaksConfig` docs

123mpozzi:
## Description

Some fields in `TweaksConfig` do not state explicitly their tvOS availability.

## Changes

- Sync tvOS availability with native SDK
- Format default values to be at the same place for iOS/tvOS related fields, and to use backticks
- Format object references to use backticks

## Checklist
- [ ] 🗒 `CHANGELOG` entry

### DIFF
--- a/src/tweaksConfig.ts
+++ b/src/tweaksConfig.ts
@@ -11,7 +11,7 @@ export interface TweaksConfig {
    * Default value in iOS is `1.0`.
    * Default value in Android is `0.2`.
    *
-   * @platform iOS, Android
+   * @platform Android, iOS, tvOS
    */
   timeChangedInterval?: number;
   /**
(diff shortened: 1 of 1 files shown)


src/tweaksConfig.ts
```diff
@@ -8,69 +8,73 @@ export interface TweaksConfig {
...
-   * Default is 0.5.
+   * Default is `0.5`.
    *
-   * @platform iOS
+   * @platform iOS, tvOS
    */
   seekToEndThreshold?: number;
   /**
-   * Specifies the player behaviour when `Player.play` is called. Default is 'relaxed'.
+   * Specifies the player behaviour when `Player.play` is called.
```
123mpozzi, line 59: Moved the default value section before `@platform`

PR #437, opened 2024-04-18
Use more verbose run names for GitHub Action workflows

rolandkakonyi:
## Description

Extends current GitHub Action run names with context specific information (e.g. inputs)

## Changes

Utilized [`run-name`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#run-name) attribute to define context aware name for the GitHub Action workflow runs.

## Checklist
- [ ] 🗒 `CHANGELOG` entry - not needed, internal change only


matamegger (approved):
Nice!

PR #441, opened 2024-05-02
Move RN function call into the lock block, to avoid missing the condition signal

matamegger:
## Description

When sending many messages in a short time over the custom message bridge, which connects the native/RN code to the javascript code in the WebUi, it can happen that the synchronous communication channel blocks due to a race condition in the condition signaling.

## Changes

Move the RN function call block into the lock guarded block, so that we for sure wait for the condition signal, before sending it. The same behaviour is already present for iOS as far as I can see. (cc @rolandkakonyi)

## Checklist
- [x] 🗒 `CHANGELOG` entry


rolandkakonyi (approved):
Thanks for bringing this to my attention.

I checked and we don't use locks in the iOS modules for our semaphores as they are thread safe AFAIK.

testcenter (approved):

PR #449, opened 2024-05-27
Introduce `NetworkConfig` API

123mpozzi:
## Description

Introduce **NetworkConfig** with two APIs for pre-processing HTTP requests and responses. 

### PRs planning 📝 

This PR contains the **public API in typescript**, and the implementation of one of the two APIs (`preprocessHttpResponse`).

At least **three more PRs will follow**. Specifically: #452 to implement the APIs in Android, #450 to implement the second API in iOS, and one to cover the API with system tests. 

However, all PRs will not be merged into development until they're all ready. In fact we will use a base branch for this feature.

### Typescript

Introduce the `NetworkConfig` API, accessible through `playerConfig.networkConfig`

#### Config definition in `network/networkConfig.ts`
This file contains the config type definition together with its other related types.

NetworkConfig:
- `preprocessHttpRequest(type, request) => Promise<HttpRequest>` to pre-process an HTTP request before the request is made
- `preprocessHttpResponse(type, response) => Promise<HttpResponse>` to pre-process an HTTP response before it gets into the player

Other relevant types
  - Notably, `HttpRequestType` has been moved here from `src/events.ts`

#### Introduce Network module in `network/index.ts`
Handles the Network module and binds the actual native calls to JS through custom listeners.

_Why a module? We need a module here because to serialise callbacks we have to call functions on a platform from the other._
_Note: this is very similar to what we did for `drm/index.ts`, although there is no asynchronousity in those callbacks._

### iOS

- Implement the network module to handle the bridging to and from JS. The bridging is used for passing the config object and running the callback on the JS side
- Add serialisers for network config types

<details>
<summary><h3>Implementation details 🔬 (*)</h3></summary> 

The main struggle here was serialising a config object containing two APIs which are both async callbacks. We went for the following approach:

- Create a native module to make the JS code callable from native
- On native side, initialise the native configuration object by fetching the JS one (e.g. using json format)
    - The config will be initialised as an empty shell initially as the callbacks cannot be serialised directly
- Check if the wanted API is defined in the fetched json config. If so, start defining the native callback: serialise the parameters needed by the native API to a JS-friendly format (1), and call the JS method responsible of running the callback using the params
- On the JS side, the method gets the serialised parameters and applies the callback (2). Finally it gets the eventual resulting object and sends it back to the native side
- The native side can now acknowledge the fact that the callback has been called on the JS side, and hence runs any additional logic (3)

(1) In this case the API has a one of its parameter being a completion handler, which is required to finalise the pre-processing. So we also assign an identifier to the response to keep track of it. We store the completion handlers on the native side
(2) In this case, given that the callback is async, this method also handles the resolving and rejection of the promise
(3) In this case it calls the completion handler assigned to the response

</details>

_Reading this additional section may be helpful for serialising callbacks in future PRs_

## Testing `preprocessHttpResponse` (iOS) 🧪 

Use [httpCallbacks.patch](https://github.com/bitmovin/bitmovin-player-react-native/files/15489165/httpCallbacks.patch) which modifies both **BasicPlayback** and **DrmPlayback**.  
_To test for a non-empty `body` field, use DRM stream samples_.   

However note that, in this PR, the request callback is not yet implemented, so it will not log or do anything!  

And finally: don't forget to run `yarn bootstrap`!

Things to look for:
- fields are serialised correctly
  - not randomly `null`
  - values make sense
- performance is fine

## Checklist
- [x] 🗒 `CHANGELOG` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 - `NetworkConfig` to configure the network request manipulation functionality
   - `preprocessHttpRequest(type, request) => Promise<HttpRequest>` to pre-process an HTTP request before the request is executed by the player
-  - `preprocessHttpResponse(type, response) => Promise<HttpResponse>` to pre-process an HTTP response before it gets into the player
+  - `preprocessHttpResponse(type, response) => Promise<HttpResponse>` to pre-process an HTTP response before it gets passed to the player
 
 ## [0.23.0] (2024-05-08)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@
   - `preprocessHttpRequest(type, request) => Promise<HttpRequest>` to pre-process an HTTP request before the request is executed by the player
   - `preprocessHttpResponse(type, response) => Promise<HttpResponse>` to pre-process an HTTP response before it gets passed to the player
 
+### Changed
+
+- Update Bitmovin's native iOS SDK version to `3.64.0`
+
 ## [0.23.0] (2024-05-08)
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Added
 
 - `NetworkConfig` to configure the network request manipulation functionality
-  - `preprocessHttpRequest(type, request) => Promise<HttpRequest>` to pre-process an HTTP request before the request is made
+  - `preprocessHttpRequest(type, request) => Promise<HttpRequest>` to pre-process an HTTP request before the request is executed by the player
   - `preprocessHttpResponse(type, response) => Promise<HttpResponse>` to pre-process an HTTP response before it gets into the player
 
 ## [0.23.0] (2024-05-08)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## [Unreleased]
+
+### Added
+
+- `NetworkConfig` to configure the network request manipulation functionality
+  - `preprocessHttpRequest(type, request) => Promise<HttpRequest>` to pre-process an HTTP request before the request is made
+  - `preprocessHttpResponse(type, response) => Promise<HttpResponse>` to pre-process an HTTP response before it gets into the player
+
 ## [0.23.0] (2024-05-08)
 
 ### Changed
--- a/ios/NetworkModule.swift
+++ b/ios/NetworkModule.swift
@@ -57,13 +57,13 @@ public class NetworkModule: NSObject, RCTBridgeModule {
     }
 
     private func initPreprocessHttpResponse(_ nativeId: NativeId, networkConfigJson: [String: Any]) {
-        guard let networkConfig = retrieve(nativeId) else {
+        guard let networkConfig = retrieve(nativeId),
+              networkConfigJson["preprocessHttpResponse"] != nil else {
             return
         }
-        if networkConfigJson["preprocessHttpResponse"] != nil {
-            networkConfig.preprocessHttpResponse = { [weak self] type, response, completionHandler in
-                self?.preprocessHttpResponseFromJS(nativeId, type, response, completionHandler)
-            }
+
+        networkConfig.preprocessHttpResponse = { [weak self] type, response, completionHandler in
+            self?.preprocessHttpResponseFromJS(nativeId, type, response, completionHandler)
         }
     }
 
--- a/ios/NetworkModule.swift
+++ b/ios/NetworkModule.swift
@@ -68,7 +68,7 @@ public class NetworkModule: NSObject, RCTBridgeModule {
         }
     }
 
-    func preprocessHttpResponseFromJS(
+    private func preprocessHttpResponseFromJS(
         _ nativeId: NativeId,
         _ type: HttpRequestType,
         _ response: HttpResponse,
--- a/ios/NetworkModule.swift
+++ b/ios/NetworkModule.swift
@@ -1,0 +1,98 @@
+import BitmovinPlayer
+
+@objc(NetworkModule)
+public class NetworkModule: NSObject, RCTBridgeModule {
+    // swiftlint:disable:next implicitly_unwrapped_optional
+    @objc public var bridge: RCTBridge!
+
+    /// In-memory mapping from `nativeId`s to `NetworkConfig` instances.
+    private var networkConfigs: Registry<NetworkConfig> = [:]
+    private var preprocessHttpResponseCompletionHandlers: Registry<(_ response: HttpResponse) -> Void> = [:]
+
+    // swiftlint:disable:next implicitly_unwrapped_optional
+    public static func moduleName() -> String! {
+        "NetworkModule"
+    }
+
+    public static func requiresMainQueueSetup() -> Bool {
+        true
+    }
+
+    // swiftlint:disable:next implicitly_unwrapped_optional
+    public var methodQueue: DispatchQueue! {
+        bridge.uiManager.methodQueue
+    }
+
+    @objc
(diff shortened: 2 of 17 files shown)


ios/RCTConvert+BitmovinPlayer.swift
```diff
@@ -215,6 +218,67 @@ extension RCTConvert {
...
+            body: body
+        )
+    }
+
+    static func networkConfig(_ json: Any?) -> NetworkConfig? {
+        guard let json = json as? [String: Any?] else {
+            return nil
+        }
+        return NetworkConfig()
+    }
```
123mpozzi, line 280: This func has no logic with the json because, as of now, the only things to serialise inside `NetworkConfig` are callbacks, which have to be set up in a different way and in their corresponding module.

ios/RCTConvert+BitmovinPlayer.swift
```diff
@@ -215,6 +218,67 @@ extension RCTConvert {
...
+            let headers = json["headers"] as? [String: String]
+        else {
+            return nil
+        }
+        var request = HttpRequest(url: url, method: method)
+        request.headers = NSMutableDictionary(dictionary: headers)
+
+        if let bodyBase64EncodedString = json["body"] as? String {
+            request.body = Data(base64Encoded: bodyBase64EncodedString)
+        }
```
123mpozzi, line 240: In iOS, we are serialising a `Data` object using base64-encoded strings

123mpozzi:
I am adding an Android reviewer to look at the public API changes in typescript files (and the changelog)

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## [Unreleased]
+
+### Added
+
+- `NetworkConfig` to configure the network request manipulation functionality
+  - `preprocessHttpRequest(type, request) => Promise<HttpRequest>` to pre-process an HTTP request before the request is made
+  - `preprocessHttpResponse(type, response) => Promise<HttpResponse>` to pre-process an HTTP response before it gets into the player
```
rolandkakonyi, line 9:
```suggestion
  - `preprocessHttpResponse(type, response) => Promise<HttpResponse>` to pre-process an HTTP response before gets passed to the player
```

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## [Unreleased]
+
+### Added
+
+- `NetworkConfig` to configure the network request manipulation functionality
+  - `preprocessHttpRequest(type, request) => Promise<HttpRequest>` to pre-process an HTTP request before the request is made
```
rolandkakonyi, line 8:
```suggestion
  - `preprocessHttpRequest(type, request) => Promise<HttpRequest>` to pre-process an HTTP request before the request is executed by the player
```

ios/NetworkModule.swift
```diff
@@ -0,0 +1,98 @@
...
+    func destroy(_ nativeId: NativeId) {
+        networkConfigs.removeValue(forKey: nativeId)
+
+        preprocessHttpResponseCompletionHandlers.keys.filter { $0.starts(with: nativeId) }.forEach {
+            preprocessHttpResponseCompletionHandlers.removeValue(forKey: $0)
+        }
+    }
+
+    private func initConfigBlocks(_ nativeId: NativeId, _ config: Any?) {
+        if let json = config as? [String: Any] {
```
rolandkakonyi, line 55: nit: we could turn this into a `guard`

ios/NetworkModule.swift
```diff
@@ -0,0 +1,98 @@
...
+            initPreprocessHttpResponse(nativeId, networkConfigJson: json)
+        }
+    }
+
+    private func initPreprocessHttpResponse(_ nativeId: NativeId, networkConfigJson: [String: Any]) {
+        guard let networkConfig = retrieve(nativeId) else {
+            return
+        }
+        if networkConfigJson["preprocessHttpResponse"] != nil {
```
rolandkakonyi, line 64: nit: We could move this statement into the `guard` above

ios/RCTConvert+BitmovinPlayer.swift
```diff
@@ -1237,6 +1301,52 @@ extension RCTConvert {
...
+        case .manifestHlsVariant:
+            return "manifest/hls/variant"
+        case .keyHlsAes:
+            return "key/hls/aes"
+        case .unknown:
+            return "unknown"
+        default:
+            return nil
+        }
+    }
```
rolandkakonyi, line 1326: This can be simplified!
```suggestion
    static func toJson(httpRequestType: HttpRequestType) -> String {
        return httpRequestType.rawValue
    }
```

src/network/networkConfig.ts
```diff
@@ -0,0 +1,162 @@
...
+  request: HttpRequest;
+  /** The HTTP status code of the response. */
+  status: number;
+  /** The URL of the response. May differ from {@link HttpRequest.url} when redirects have happened. */
+  url: string;
+}
+
+/**
+ * The network API gives the ability to influence network requests.
+ * It enables preprocessing requests and processing responses.
```
rolandkakonyi, line 67:
```suggestion
 * It enables preprocessing requests and preprocessing responses.
```

src/network/networkConfig.ts
```diff
@@ -0,0 +1,162 @@
...
+   *  });
+   * ```
+   */
+  preprocessHttpRequest?: (
+    type: HttpRequestType,
+    request: HttpRequest
+  ) => Promise<HttpRequest>;
+  /**
+   * Called before an HTTP response is accessed by the player.
+   * Can be used to the access or change properties of the response.
```
rolandkakonyi, line 117: 1. There is an extra "the"
2. It can be simplified to "change the response"
```suggestion
   * Can be used to access or change the response.
```

ios/RCTConvert+BitmovinPlayer.swift
123mpozzi: Amazing 🤩 , thanks for catching this! I really had an oversight here 😅

ios/NetworkModule.swift
123mpozzi: Nice idea! 💪 
Address in f12ec936277274ec7a74bbf9f314bc621d855169

ios/NetworkModule.swift
123mpozzi: Let's do it 🙌 , it's always a nice idea to simplify the code 🙂 
Address in c91309699d564b216c061ede528d66f8a2c6a5e6

rolandkakonyi (approved):

src/network/networkConfig.ts
```diff
@@ -0,0 +1,162 @@
...
+  /** The HTTP method of the request. */
+  method: string;
+  /** The URL of the request. */
+  url: string;
+}
+
+/**
+ * Base64-encoded string representing the HTTP response body.
+ */
+export type HttpResponseBody = string;
```
matamegger, line 46: Why base64 encoded? 🤔

src/network/networkConfig.ts
```diff
@@ -0,0 +1,162 @@
...
+  DrmCertificateFairplay = 'drm/certificate/fairplay',
+  DrmLicenseWidevine = 'drm/license/widevine',
+  KeyHlsAes = 'key/hls/aes',
+  Unknown = 'unknown',
+}
+
+/**
+ * Base64-encoded string representing the HTTP request body.
+ */
+export type HttpRequestBody = string;
```
matamegger, line 26: Why base64 encoded?

src/network/networkConfig.ts
rolandkakonyi: We decided on base64 encoded string as there is no native support for RN to send raw data between TS and native. Introducing this might require bigger changes.
This is something that could be improved later (with breaking changes).

matamegger (approved):

PR #450, opened 2024-05-29
Implement `NetworkConfig.preprocessHttpRequest` for iOS and tvOS

rolandkakonyi:
## Description

Implement `NetworkConfig.preprocessHttpRequest` on iOS and tvOS after the API was defined in #449.

## Changes

Implemented connecting the TS `NetworkConfig.preprocessHttpRequest` with the native `NetworkConfig.preprocessHttpRequestDelegate` on the iOS/tvOS side to enable the React Native SDK to manipulate HTTP requests before executed.

No public API changes were made.

CI checks fail for iOS/tvOS because we use an API from the iOS Player SDK that is not released yet.

## Checklist
- [x] 🗒 `CHANGELOG` entry - already exists

### DIFF
--- a/ios/NetworkModule.swift
+++ b/ios/NetworkModule.swift
@@ -7,6 +7,8 @@ public class NetworkModule: NSObject, RCTBridgeModule {
 
     /// In-memory mapping from `nativeId`s to `NetworkConfig` instances.
     private var networkConfigs: Registry<NetworkConfig> = [:]
+    private var preprocessHttpRequestDelegateBridges: Registry<PreprocessHttpRequestDelegate> = [:]
+    private var preprocessHttpRequestCompletionHandlers: Registry<(_ response: HttpRequest) -> Void> = [:]
     private var preprocessHttpResponseCompletionHandlers: Registry<(_ response: HttpResponse) -> Void> = [:]
 
     // swiftlint:disable:next implicitly_unwrapped_optional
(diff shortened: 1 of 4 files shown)


ios/PreprocessHttpRequestDelegateBridge.swift
```diff
@@ -0,0 +1,30 @@
...
+        self.bridge = bridge
+        super.init()
+    }
+
+    func preprocessHttpRequest(
+        _ type: String,
+        httpRequest: HttpRequest,
+        completionHandler: @escaping (
+            HttpRequest
+        ) -> Void
```
123mpozzi, line 18: nit

```suggestion
        completionHandler: @escaping (
            _ request: HttpRequest
        ) -> Void
```

ios/NetworkModule.swift
```diff
@@ -68,6 +87,35 @@ public class NetworkModule: NSObject, RCTBridgeModule {
...
+        type: String,
+        request: HttpRequest,
+        completionHandler: @escaping (
+            _ request: HttpRequest
+        ) -> Void
+    ) {
+        let requestId = "\(nativeId)@\(ObjectIdentifier(request).hashValue)"
+        let args: [Any] = [
+            requestId,
+            type,
```
123mpozzi, line 100: Just a note here: I see we are directly passing a String param instead of going through `RCTConvert` because that's the way it is defined in the `PreprocessHttpRequestDelegate` protocol, as opposed to what we do for the `HttpResponse` variation.

Still, it is working well in my testing 💪

ios/NetworkModule.swift
```diff
@@ -68,6 +87,35 @@ public class NetworkModule: NSObject, RCTBridgeModule {
         }
     }
 
+    internal func preprocessHttpRequestFromJS(
+        nativeId: NativeId,
+        type: String,
+        request: HttpRequest,
+        completionHandler: @escaping (
+            _ request: HttpRequest
+        ) -> Void
```
123mpozzi, line 95: A note here: the style of arguments is inconsistent compared to `preprocessHttpResponseFromJS`. I think we should align them, but I honestly prefer this way.
What do you think if I change `preprocessHttpResponseFromJS` to use named parameters in the main PR?

123mpozzi (changes requested):
Nice job! 💪  The code is working well!
I've just added comments on some minor things

ios/NetworkModule.swift
rolandkakonyi: If you like it this way, feel free to update the other PR

123mpozzi (approved):

PR #453, opened 2024-05-29
Update iOS player to 3.64.0

bitPlayerGHActions:
Automated iOS player version update to 3.64.0


123mpozzi (approved):
Smoke-tested every sample on the demo app apart from casting and it is looking good to me 💪

PR #456, opened 2024-06-03
Allow multiple TS players to control the same native player.

krocard:
## Problem (PRN-122)
Some clients implement fullscreen by creating a new fullscreen Player instance
with the name nativeId.
This was documented as supported, but on Android, was not.

## Changes
When a native ID is reused, don't do initialize a new native instance.
Instead, reuse the same native instance.

## Test
Have 2 players on the same screen with the same native-id. Pausing one pauses the other.

image

## Checklist
- [x] 🗒 `CHANGELOG` entry if applicable

### DIFF
--- a/android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt
@@ -1,6 +1,8 @@
 package com.bitmovin.player.reactnative
 
+import android.util.Log
 import com.bitmovin.analytics.api.DefaultMetadata
+import com.bitmovin.player.PlayerView
 import com.bitmovin.player.api.Player
 import com.bitmovin.player.api.PlayerConfig
 import com.bitmovin.player.api.analytics.create
@@ -14,6 +16,7 @@ import com.bitmovin.player.reactnative.extensions.mapToReactArray
 import com.facebook.react.bridge.*
 import com.facebook.react.module.annotations.ReactModule
 import java.security.InvalidParameterException
+import kotlin.math.log
 
 private const val MODULE_NAME = "PlayerModule"
 
@@ -63,6 +66,9 @@ class PlayerModule(context: ReactApplicationContext) : BitmovinBaseModule(contex
         promise: Promise,
     ) = promise.unit.resolveOnUiThread {
         if (players.containsKey(nativeId)) {
+            if (playerConfigJson != null || analyticsConfigJson != null) {
+                Log.w("BitmovinPlayerModule", "Cannot reconfigure an existing player")
+            }
             return@resolveOnUiThread // key can be reused to access the same native instance (see NativeInstanceConfig)
         }
         val playerConfig = playerConfigJson?.toPlayerConfig() ?: PlayerConfig()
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -39,7 +39,7 @@
     "bootstrap": "yarn bootstrap:example && yarn bootstrap:integration-test",
     "bootstrap:example": "yarn && yarn example && yarn example pods && yarn brew",
     "bootstrap:integration-test": "yarn && yarn integration-test && yarn integration-test pods && yarn brew",
-    "brew": "[ \"$(uname)\" = \"Darwin\" ] && brew bundle install --no-lock",
+    "brew": "[ \"$(uname)\" != Darwin ] || brew bundle install --no-lock",
```
rolandkakonyi, line 42: Did you test this if we really don't need `"` around?

krocard, line 42: Fix boostrap failing on Linux.

krocard: :thinking: This is `sh`, you only need quotes if you have special characters. `Darwin` has none.

android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt
```diff
@@ -63,7 +63,7 @@ class PlayerModule(context: ReactApplicationContext) : BitmovinBaseModule(contex
         promise: Promise,
     ) = promise.unit.resolveOnUiThread {
         if (players.containsKey(nativeId)) {
-            throw IllegalArgumentException("Duplicate player creation for id $nativeId")
+            return@resolveOnUiThread // key can be reused to access the same native instance (see NativeInstanceConfig)
```
matamegger, line 66: If the new TS instance has different configurations, they are ignored (which is fine IMO).
Should we check the configs here and show a warning if they do not match?

android/src/main/java/com/bitmovin/player/reactnative/SourceModule.kt
```diff
@@ -76,12 +76,12 @@ class SourceModule(context: ReactApplicationContext) : BitmovinBaseModule(contex
         analyticsSourceMetadata: ReadableMap?,
         promise: Promise,
     ) = promise.unit.resolveOnUiThread {
+        if (sources.containsKey(nativeId)) {
+            return@resolveOnUiThread // key can be reused to access the same native instance (see NativeInstanceConfig)
```
matamegger, line 80: If the new TS instance has different configurations, they are ignored (which is fine IMO).
Should we check the configs here and show a warning if they do not match?

android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt
krocard: Added a warning: https://github.com/bitmovin/bitmovin-player-react-native/pull/456/commits/6619a6ef28c496ebf4ed620019c7076c5e24a51d

Unfortunately, it is not a player event as JS is not yet listening for events when the player is create (only when it is attached to a view).

android/src/main/java/com/bitmovin/player/reactnative/SourceModule.kt
krocard: Added a warning: https://github.com/bitmovin/bitmovin-player-react-native/pull/456/commits/6619a6ef28c496ebf4ed620019c7076c5e24a51d

android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt
matamegger: That seems like a simple but effective solution 😄 👍

matamegger (approved):

rolandkakonyi (approved):

PR #463, opened 2024-06-05
Add ref support to `PlayerView`

rolandkakonyi:
## Description

This PR enables future integrations to have access to the underlying native `PlayerView` instances.

## Changes

Added the possibility to inject a `React.MutableRefObject` for setting initial value and track value changes of `nativeView` reference inside `PlayerView`.
This can be then used to fetch the related native `PlayerView` based on this reference.
This is crucial for integrations accessing `PlayerView` natively.

Note: I couldn't use the property name `ref` as that triggers the error of `Function components cannot be given refs. Attempts to access this ref will fail.`

It can be used like this:
```tsx
const player = usePlayer();
const viewRef = useRef(null);

return <PlayerView viewRef={viewRef} player={player} />
```

## Checklist
- [x] 🗒 `CHANGELOG` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [Unreleased]
+
+### Added
+
+- `viewRef` property to `PlayerView` to allow setting a reference to the native view
+
 ## [0.24.0] - 2024-06-04
 
 ### Added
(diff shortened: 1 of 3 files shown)


123mpozzi (approved):
Nice job 💪

PR #469, opened 2024-06-13
Feature/set video quality

saravanans-github:
## Description
new feature to setVideoQuality on Android. This feature is not available on iOS and tvOS devices

## Changes
- Integrated with the `source.setVideoQuality` Bitmovin Android Player API
- Platform check to throw warning on non-Android devices

## Checklist
- [ ] 🗒 `CHANGELOG` entry

### DIFF
--- a/src/player.ts
+++ b/src/player.ts
@@ -438,6 +438,20 @@ export class Player extends NativeInstance<PlayerConfig> {
   };
 
   /**
+   * Sets the video quality.
+   * ONLY works on Android. No effect on iOS and tvOS devices.
+   */
+  setVideoQuality = (qualityId: String) => {
+    if (Platform.OS !== 'android') {
+      console.warn(
+        `[Player ${this.nativeId}] Method setVideoQuality is not available for iOS and tvOS devices. Only Android devices.`
+      );
+    } else {
+      PlayerModule.setVideoQuality(this.nativeId, qualityId);
+    }
+  };
+
+  /**
    * Sets the playback speed of the player. Fast forward, slow motion and reverse playback are supported.
    * @note
    * - Slow motion is indicated by values between `0` and `1`.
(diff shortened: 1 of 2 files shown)


src/player.ts
```diff
@@ -437,6 +437,20 @@ export class Player extends NativeInstance<PlayerConfig> {
...
+   * ONLY works on Android. No effect on iOS and tvOS devices.
+   */
+  setVideoQuality = (qualityId: String) => {
+    if (Platform.OS !== 'android') {
+      console.warn(
+        `[Player ${this.nativeId}] Method setVideoQuality is not available for iOS and tvOS devices. Only Android devices.`
+      );
+    } else {
+      PlayerModule.setVideoQuality(this.nativeId, qualityId);
+    }
```
rolandkakonyi, line 451: nit: I think it would be easier to read if we add an early return inside the `if` statement and extract the content of the `else` statement, like here:

```suggestion
    if (Platform.OS !== 'android') {
      console.warn(
        `[Player ${this.nativeId}] Method setVideoQuality is not available for iOS and tvOS devices. Only Android devices.`
      );
      return;
    }
    PlayerModule.setVideoQuality(this.nativeId, qualityId);
```

rolandkakonyi (commented):
Note: I have only reviewed the TypeScript code changes

src/player.ts
```diff
@@ -437,6 +437,20 @@ export class Player extends NativeInstance<PlayerConfig> {
     return PlayerModule.getAvailableVideoQualities(this.nativeId);
   };
 
+  /**
+   * Sets the video quality.
+   * ONLY works on Android. No effect on iOS and tvOS devices.
+   */
```
rolandkakonyi, line 443: Please also document the parameter. The most important part here is to mention the special value `'auto'`!

```suggestion
   *
   * @param qualityId value obtained from `VideoQuality.id`, which can be obtained via `Player.getAvailableVideoQualities()` to select a specific quality. To use automatic quality selection, 'auto' can be passed here.
   */
```

rolandkakonyi (changes requested):
I have reviewed the Android code too!

Manual testing looks good!

Please address the documentation comments!

rolandkakonyi (approved):

rolandkakonyi:
I missed that there was no CHANGELOG entry added, added it here retroactively: aed31c9f

PR #475, opened 2024-06-17
Fix integration-test `Podfile.lock` not being staged during release

123mpozzi:
## Description

The release action is not staging the changes to `integration_test/ios/Podfile.lock`, so it can stay misaligned with the pod file of the examples folder

## Changes

Stage `integration_test/ios/Podfile.lock` too

## Checklist
- [x] 🗒 `CHANGELOG` entry: `no need`


rolandkakonyi (approved):
Thanks for catching!

PR #487, opened 2024-07-04
Fix `player.setMaxSelectableBitrate` by fixing Xcode 16 build

rolandkakonyi:
## Description

Found an actual issue when trying to fix Xcode 16 build for iOS and tvOS.

## Changes

Fixed the `UInt` value usage by using `Int` for comparison with `-1` then convert to `UInt` when assigning to the player API.

## Checklist
- [x] 🗒 `CHANGELOG` entry


123mpozzi (approved):
Nice catch! 💪

PR #498, opened 2024-08-05
Feat expose ad item preload offset

saravanans-github:
## Description
Expose the AdItem. preloadOffset property. This property is NOT available for iOS. Android documentation [reference](https://cdn.bitmovin.com/player/android/3/docs/player-core/com.bitmovin.player.api.advertising/-ad-item/preload-offset.html)

## Changes
1. added the property with type `Number`
2. copied the existing documentation from the Android API, including mention of default value
3. add @platform Android to the documentation to indicate it's ONLY for Android
4. Parsed the RN value and return it as an Android AdItem object in JsonConvertor.toAdItem

## Checklist
- [ ] 🗒 `CHANGELOG` entry

### DIFF
--- a/android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
@@ -206,7 +206,7 @@ fun ReadableMap.toAdItem(): AdItem? {
     return AdItem(
         sources = getArray("sources") ?.toMapList()?.mapNotNull { it?.toAdSource() }?.toTypedArray() ?: return null,
         position = getString("position") ?: "pre",
-        preloadOffset = getString("preloadOffset") ?: 0,
+        preloadOffset = getDoubleOrNull("preloadOffset") ?: 0.0,
     )
 }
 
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
```diff
@@ -206,6 +206,7 @@ fun ReadableMap.toAdItem(): AdItem? {
     return AdItem(
         sources = getArray("sources") ?.toMapList()?.mapNotNull { it?.toAdSource() }?.toTypedArray() ?: return null,
         position = getString("position") ?: "pre",
+        preloadOffset = getString("preloadOffset") ?: 0,
```
rolandkakonyi, line 209: Please read the value as a `Double` to match type. See related build error: https://github.com/bitmovin/bitmovin-player-react-native/actions/runs/10245493463/job/28340612293#step:4:274

```suggestion
        preloadOffset = getDoubleOrNull("preloadOffset") ?: 0,
```

src/advertising.ts
```diff
@@ -76,6 +76,14 @@ export interface AdItem {
    * The fallback ad sources need to have the same `AdSourceType` as the main ad source.
    */
   sources: AdSource[];
+
+  /**
+   * @platform Android
+   * The amount of seconds the ad manifest is loaded in advance;
+   * i.e as compared to when the ad break is scheduled for playback.
+   * Default value is 0.0
```
rolandkakonyi, line 84: Please match the documentation here: v

Like:
```suggestion
   * The amount of seconds the ad manifest is loaded in advance
   * compared to when the ad break is scheduled for playback.
   *
   * Default value is 0.0
   *
   * @platform Android
```

rolandkakonyi (changes requested):

PR #509, opened 2024-08-28
Feat/initial bandwidth estimate override

saravanans-github:
## Description
Added support for the `adaptationConfig.initialBandwidthEstimateOverride` config parameter.

## Changes
Android initialBandwidthEstimateOverride is type Long. No such type on JS hence converted from int to long. Hope that's ok. Max value of Int `2,147,483,647 ` should be sufficient for this usecase I believe.

## Checklist
- [ ] 🗒 `CHANGELOG` entry

### DIFF
--- a/android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
@@ -144,6 +144,7 @@ private fun String.toTimelineReferencePoint(): TimelineReferencePoint? = when (t
  */
 private fun ReadableMap.toAdaptationConfig(): AdaptationConfig = AdaptationConfig().apply {
     withInt("maxSelectableBitrate") { maxSelectableVideoBitrate = it }
+    withInt("initialBandwidthEstimateOverride") { initialBandwidthEstimateOverride = it.toLong(); }
 }
 
 /**
--- a/src/adaptationConfig.ts
+++ b/src/adaptationConfig.ts
@@ -7,4 +7,12 @@ export interface AdaptationConfig {
    * Can be set to `undefined` for no limitation.
    */
   maxSelectableBitrate?: number;
+
+  /**
+   * The initial bandwidth estimate in bits per second the player uses to select the optimal media tracks before actual bandwidth data is available. Overriding this value should only be done in specific cases and will most of the time not result in better selection logic.
+   *
+   * @platform Android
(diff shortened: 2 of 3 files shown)


src/adaptationConfig.ts
```diff
@@ -7,4 +7,12 @@ export interface AdaptationConfig {
    * Can be set to `undefined` for no limitation.
    */
   maxSelectableBitrate?: number;
+
+  /**
+   * The initial bandwidth estimate in bits per second the player uses to select the optimal media tracks before actual bandwidth data is available. Overriding this value should only be done in specific cases and will most of the time not result in better selection logic.
+   *
+   * @platform Android
+   * @see https://cdn.bitmovin.com/player/android/3/docs/player-core/com.bitmovin.player.api.media/-adaptation-config/initial-bandwidth-estimate-override.html?query=var%20initialBandwidthEstimateOverride:%20Long?
```
strangesource, line 15:
```suggestion
   * @see https://cdn.bitmovin.com/player/android/3/docs/player-core/com.bitmovin.player.api.media/-adaptation-config/initial-bandwidth-estimate-override.html
```

Actually not sure if this is very helpful. The documentation here is the same as the linked one. I would suggest to not include the link, also as it could get outdated.

krocard, line 15:
```suggestion
   * @see https://cdn.bitmovin.com/player/android/3/docs/player-core/com.bitmovin.player.api.media/-adaptation-config/initial-bandwidth-estimate-override.html

File: android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt [hunk 2]
```diff
@@ -144,6 +144,7 @@ private fun String.toTimelineReferencePoint(): TimelineReferencePoint? = when (t
  */
 private fun ReadableMap.toAdaptationConfig(): AdaptationConfig = AdaptationConfig().apply {
     withInt("maxSelectableBitrate") { maxSelectableVideoBitrate = it }
+    withInt("initialBandwidthEstimateOverride") { initialBandwidthEstimateOverride = it.toLong(); }
```
krocard, line 147: JS `number` is a double. Not sure why this is an Int :thinking: Should both be changes to `withDouble`?

krocard (approved):

strangesource (approved):

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
saravanans-github: I'll leave it as-is for now please as this seems inline with the `withInt` implementation of both `bandwidthEstimateWeightLimit` and `maxSelectableBitrate`

PR #539, opened 2024-10-18
Android: support Media Controls

123mpozzi:
## Description

Implement the **media controls** in Android (via the MediaSession API) to make the system display the current media information on the lock-screen, in notifications, and within the control-center.

Additionally, this PR also introduces **background-playback** support

### PR-planning
- ✅    iOS part: #532 
- 📍   Android part
Both will be merged together on a base branch

### Overview pics
image
image

#### Implement media controls
- Use a MediaSession service to hold a strong reference to the player instance
- Add serializers for `MediaControlConfig` 
- Update manifest permissions to allow running the background service

#### Implement Background playback
- Modify the PlayerView lifecycle to prevent auto-pausing the background playback on app minimize
  - Fix `playerEventRelay` being null in some scenarios

#### Changes to samples
To allows having playback without the main app being opened
- **Media Controls** enabled by default in all samples
  - No background playback, so it pauses the app on minimizing it
- **Background Playback**
  - It does not pause the playback when minimizing the app 
  - Has media controls enabled (since we need a notification, we favor media controls over a simple notification)

> [!NOTE]
> Note: both samples will destroy the player instance (hence destroying the media control widgets) when going back to the homepage. That is expected, a more advanced sample could be part of a follw-up (e.g. letting the widget live in the homescreen as well, until we open a sample with media session disabled).

### Config Flags Design

- `MediaControlConfig.isEnabled` decides whether the service starts or not
- `PlayerConfig.PlaybackConfig.isBackgroundPlaybackEnabled` decides whether to detach the player from the view, basically handling whether the playback should automatically pause (when set to`false`) or not (when set to `true`)
  - `MediaControlConfig.isEnabled` has to be `true` for background playback to work, since it needs a notification

## Testing

Please do some manual testing, with a physical device if possible 🙏 

#### Media Controls (use any Media Control sample apart from Background Playback)
- widget is created
  - in notification-centre
  - in lock-screen
- minimizing the app stops the playback
  - when re-opening the app, the player data is synced (e.g. current playback time)

#### Background Playback feature (use the new "Background Playback" sample)
- widget is created
  - in notification-centre
  - in lock-screen
- minimizing the app does NOT stop the playback
  - when re-opening the app, the player data is synced (e.g. current playback time)
- does not work without media session being enabled as well (via `MediaControlConfig`)

#### Activity lifecycle changes
- basic samples have no background playback
  - by default
  - after running the new samples
- basic samples playerview is not empty after running the new samples

## Checklist
- [x] 🗒 `CHANGELOG` entry - `entry added already in base PR`


android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -116,13 +134,24 @@ class RNPlayerView(
...
 
         override fun onDestroy(owner: LifecycleOwner) = dispose()
     }
 
     init {
+        playerEventRelay = EventRelay<Player, Event>(
+            EVENT_CLASS_TO_REACT_NATIVE_NAME_MAPPING,
+            ::emitEventFromPlayer,
+        )
+
```
123mpozzi, line 154: I had to move `playerEventRelay` initialization here because it was `null` during the first invocations of `onStop`.
This is due to the custom `setPlayer` we have, which also sets `playerEventRelay`

src/lockScreenControlConfig.ts
123mpozzi, line 1: Note that the **Limitations** section is not modified.
It will be update once the lock-screen feature is fully released on the Android SDK

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
```diff
@@ -883,6 +884,21 @@ fun String.toMediaType(): MediaType? = when (this) {
...
+    withBoolean("isEnabled") { isEnabled = it }
+}
+
+private val lockScreenConfigMap = mutableMapOf<PlayerConfig, LockScreenControlConfig>()
+var PlayerConfig.lockScreenControlConfig: LockScreenControlConfig
+    get() = lockScreenConfigMap[this] ?: LockScreenControlConfig()
+    set(value) {
+        lockScreenConfigMap[this] = value
+    }
+
```
123mpozzi, line 901: This config is a cross-platform wrapper, it does not exist in either native SDK.
It acts as a common name to reach the feature

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
```diff
@@ -883,6 +883,14 @@ fun String.toMediaType(): MediaType? = when (this) {
...
 }
 
+data class LockScreenControlConfig(
+    var isEnabled: Boolean = false,
+)
+
+fun ReadableMap.toLockScreenControlConfig(): LockScreenControlConfig = LockScreenControlConfig().apply {
+    withBoolean("isEnabled") { isEnabled = it }
+}
+
```
123mpozzi, line 893: This config is a cross-platform wrapper, it does not exist in either native SDK.
It acts as a common name to reach the feature

android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt
```diff
@@ -74,6 +78,10 @@ class PlayerModule(context: ReactApplicationContext) : BitmovinBaseModule(contex
         val playerConfig = playerConfigJson?.toPlayerConfig() ?: PlayerConfig()
         val analyticsConfig = analyticsConfigJson?.toAnalyticsConfig()
         val defaultMetadata = analyticsConfigJson?.getMap("defaultMetadata")?.toAnalyticsDefaultMetadata()
+        val enableMediaSession = playerConfigJson?.getMap("lockScreenControlConfig")
+            ?.toLockScreenControlConfig()?.isEnabled ?: false
+        enableBackgroundPlayback = playerConfigJson?.getMap("playbackConfig")
+            ?.getBoolean("isBackgroundPlaybackEnabled") ?: false
```
123mpozzi, line 84: This needs special mapping as it is a React-Native (and iOS as well) flag

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -102,8 +103,26 @@ class RNPlayerView(
     private val activityLifecycle = (context.currentActivity as? ReactActivity)?.lifecycle
         ?: error("Trying to create an instance of ${this::class.simpleName} while not attached to a ReactActivity")
 
+    /**
+     * Relays the provided set of events, emitted by the player, together with the associated name
+     * to the `eventOutput` callback.
+     */
+    private var playerEventRelay: EventRelay<Player, Event>
```
zigavehovec, line 110: You can just initialize the eventRelay here, no need for an extra init block 🙂

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -102,8 +103,26 @@ class RNPlayerView(
...
+
+    private var playerInMediaSessionService: Player?
+        get() = context.playerModule?.mediaSessionPlaybackManager?.serviceBinder?.player
+        set(value) {
+            value?.let {
```
zigavehovec, line 115: Don't we want to set the player in service binder to null if we set playerInMediaSessionService to null like on the line 139? With this `playerInMediaSessionService = null` has no effect.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -102,8 +103,26 @@ class RNPlayerView(
...
+                context.playerModule?.mediaSessionPlaybackManager?.serviceBinder?.player = it
+            }
+        }
+
     private val activityLifecycleObserver = object : DefaultLifecycleObserver {
+        // Don't stop the player when going to background
```
zigavehovec, line 121: This comment looks out of place for the onStart lifecycle.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -102,8 +103,26 @@ class RNPlayerView(
...
         override fun onStart(owner: LifecycleOwner) {
+            if (playerInMediaSessionService != null) {
+                player = playerInMediaSessionService
```
zigavehovec, line 124: Does this mean that if the media session is active and then the user opens up another video with a different player configuration it will be ignored right? 

So this scenario: 
- App has a list of video items (all have background playback enabled)
- User opens the first video and we create a media session
- User goes to background (where the app might be killed)
- Player persists in the media session
- User opens the app and navigates to a different video item
- Instead of playing the other video the player instance will be overriten by the player in the media session

I'm just trying to understand if this is a known limitation that we are ok with 🙂

android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt
```diff
@@ -74,6 +78,10 @@ class PlayerModule(context: ReactApplicationContext) : BitmovinBaseModule(contex
         val playerConfig = playerConfigJson?.toPlayerConfig() ?: PlayerConfig()
         val analyticsConfig = analyticsConfigJson?.toAnalyticsConfig()
         val defaultMetadata = analyticsConfigJson?.getMap("defaultMetadata")?.toAnalyticsDefaultMetadata()
+        val enableMediaSession = playerConfigJson?.getMap("lockScreenControlConfig")
+            ?.toLockScreenControlConfig()?.isEnabled ?: false
+        enableBackgroundPlayback = playerConfigJson?.getMap("playbackConfig")
```
zigavehovec, line 83: I'm not sure about double source of truth (kind of). But basically my understanding is that once media session is created we automatically support background playback as the player instance is held in the media session service and thus won't be killed by the system. Do we really need an extra flag to enable/disable background playback? Has this been discussed at all?

From my point of view playing on the lock screen is already playing in the background (the app isn't visible - in foreground anymore), so I don't really see the value in this extra flag. Just thinking from the Android point of view, I know that this flag exists for iOS already.

What I'm getting at is that in my opinion we might only want to enable backgorund playback where lockscreen controls would be there automatically, but I might not have enough context what the expectations are on iOS or from the customers.

android/src/main/java/com/bitmovin/player/reactnative/services/MediaSessionPlaybackService.kt
```diff
@@ -0,0 +1,55 @@
...
+                    connectSession()
+                }
+            }
+    }
+
+    private var player: Player? = null
+    private val binder = ServiceBinder()
+    private var mediaSession: MediaSession? = null
+
+    override fun onGetSession(): MediaSession? = null
```
zigavehovec, line 28: Shouldn't this return the mediaSession instance?

zigavehovec (commented):
Nice work 💪! I commented on a few things I noticed. 
I'm still going to do some testing and report back 🙂

zigavehovec:
Testing looks good. 
The only thing that I noticed is the "app gets killed in the background" scenario, where if we navigate to the home screen the playback correctly continues. Then we kill the app and the playback continues. When we open the app again and go to the same video we create a new media session, so there are effectively 2 active sessions playing back at the same time. I guess we should probably destroy the old session or continue the old one - up to discussion.

android/src/main/java/com/bitmovin/player/reactnative/services/MediaSessionPlaybackService.kt
123mpozzi: As this function is not used anywhere, we decided to 'restrict' its usage by setting it to `null`

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
123mpozzi: Address in 503a4fc2cad5854751751937e115b988c9f53646

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
123mpozzi: Yeah I agree, it is confusing.
Address in 5a4ff3c379dcaef1dab6fbaa5d6ed9349b976770

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
123mpozzi: You're right, nice catch! 💪 
Address in abaedf2fe5aedfd66baa95d12785e257ea35c0b4

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -116,6 +131,11 @@ class RNPlayerView(
         }
 
         override fun onStop(owner: LifecycleOwner) {
+            if (enableBackgroundPlayback) {
+                playerInMediaSessionService = player
+                player = null
+            }
```
hawk23, line 137: As far as I understand the situation, we still have the same bug here. Imagine we have two players and both have media session controls and background playback enabled. Only one of them will be referenced in the service (the one that was created last). Here in `onStop`, we should not store `playerInMediaSessionService` if the player in the service is not matching the one that is managed here.

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
```diff
@@ -883,6 +883,14 @@ fun String.toMediaType(): MediaType? = when (this) {
     else -> null
 }
 
+data class MediaControlConfig(
+    var isEnabled: Boolean = false,
+)
```
hawk23, line 888: Shouldn't the default value be `true`?

android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt
```diff
@@ -74,6 +77,8 @@ class PlayerModule(context: ReactApplicationContext) : BitmovinBaseModule(contex
         val playerConfig = playerConfigJson?.toPlayerConfig() ?: PlayerConfig()
         val analyticsConfig = analyticsConfigJson?.toAnalyticsConfig()
         val defaultMetadata = analyticsConfigJson?.getMap("defaultMetadata")?.toAnalyticsDefaultMetadata()
+        val enableMediaSession = playerConfigJson?.getMap("mediaControlConfig")
+            ?.toMediaControlConfig()?.isEnabled ?: true
```
hawk23, line 81: Nit: I would prefer defaulting to `false` here as this is rather an error case (I assume a `mediaControlConfig` should always be present here?). The correct default value should be handled somewhere else I guess.

android/src/main/java/com/bitmovin/player/reactnative/services/MediaSessionPlaybackService.kt
```diff
@@ -0,0 +1,55 @@
...
+    inner class ServiceBinder : Binder() {
+        var player: Player?
+            get() = this@MediaSessionPlaybackService.player
+            set(value) {
+                disconnectSession()
+                this@MediaSessionPlaybackService.player = value
+                value?.let {
+                    createSession(it)
+                    connectSession()
+                }
```
hawk23, line 20: Nit: Would be a nice improvement to not destroy and recreate the media session in case the same player is set again

android/src/main/java/com/bitmovin/player/reactnative/services/MediaSessionPlaybackService.kt
hawk23: It's fine returning `null` here as we call `addSession` manually in `connectSession`

example/src/App.tsx
```diff
@@ -20,6 +20,8 @@ import LandscapeFullscreenHandling from './screens/LandscapeFullscreenHandling';
 import SystemUI from './screens/SystemUi';
 import OfflinePlayback from './screens/OfflinePlayback';
 import Casting from './screens/Casting';
+import LockScreenControls from './screens/LockScreenControls';
```
hawk23, line 23: I guess we should not use the naming "Lock screen controls" anymore to stay consistent with the API naming. Would be good to rename every occurrence in the whole code base (incl. file names)

android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt
123mpozzi: Address in fe1e6517f1fe2b63efff420572a4f4f1d439b537 , following the design we decided in the meeting

android/src/main/java/com/bitmovin/player/reactnative/services/MediaSessionPlaybackService.kt
zigavehovec: Ok, thanks!

android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt
123mpozzi: I'd prefer to keep it this way since a `mediaControlConfig` may not be present at all. Because it is a ReactNative-only object, so Native SDKs are not setting this explicitly

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
123mpozzi: Address in 2ee1f112df2f06585a9716b6227c7de4168b1ee8

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
123mpozzi: Address with the changes to state-management in 36e88e6d28a8066797fef76bb963fb2dc930c50b

example/src/App.tsx
123mpozzi: Address in b97c8c1eb5b660354be1b087db575c04ff51321c

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
123mpozzi: Address with the changes to state-management in 36e88e6d28a8066797fef76bb963fb2dc930c50b

123mpozzi:
> Testing looks good. The only thing that I noticed is the "app gets killed in the background" scenario, where if we navigate to the home screen the playback correctly continues. Then we kill the app and the playback continues. When we open the app again and go to the same video we create a new media session, so there are effectively 2 active sessions playing back at the same time. I guess we should probably destroy the old session or continue the old one - up to discussion.

Address with the changes to state-management in 36e88e6d28a8066797fef76bb963fb2dc930c50b

android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt
```diff
@@ -25,6 +26,8 @@ class PlayerModule(context: ReactApplicationContext) : BitmovinBaseModule(contex
      */
     private val players: Registry<Player> = mutableMapOf()
 
+    var mediaSessionPlaybackManager: MediaSessionPlaybackManager? = null
```
hawk23: As discussed offline, let's init the `mediaSessionPlaybackManager` once here, instead of every time there is a player created with media controls support enabled.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
```diff
@@ -296,6 +297,7 @@ class RNPlayerViewManager(private val context: ReactApplicationContext) : Simple
             block()
         } catch (e: Exception) {
             Log.e(MODULE_NAME, "Error while executing command", e)
+            // TODO: re throw
```
hawk23, line 300: Please remove and create ticket for this

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -190,6 +215,7 @@ class RNPlayerView(
         // is responsible for destroying the player in the end.
         playerView.player = null
         playerView.onDestroy()
+        (playerView.parent as? ViewGroup)?.removeView(playerView)
```
hawk23, line 218: That was just for debugging, please remove

android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt
123mpozzi: Address in cf7c34f41d4b752a1d2209df1029babd30b823e9 and finally in f0b06d7f338a79ed86b7567b248400a773bad994

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
123mpozzi: Address in 554967836fc4e69ceaed9cca7bd9b446cee85ad8

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
123mpozzi: Address in 554967836fc4e69ceaed9cca7bd9b446cee85ad8

hawk23 (approved):

123mpozzi:
Since 6d625727e94186932894d5da40136fe2177db670 a patch is no longer needed for testing

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -116,10 +132,28 @@ class RNPlayerView(
...
             playerView?.onStop()
         }
 
         override fun onDestroy(owner: LifecycleOwner) = dispose()
+
+        // When background playback is enabled,
+        // remove player from view so it does not get paused when entering background
+        private fun removePlayerForBackgroundPlayback() {
+            playerInMediaSessionService = null
+            playerView?.player?.let {
```
zigavehovec, line 145: Nit: I'm not a big fan of using `let` for nullability checks as the code gets an extra indentation, consider replacing it with:
`val player = playerView?.player ?: return`

It's also generally nice to avoid `it` and name the parameter if the code is more than 1 line.

zigavehovec (approved):
Nice work 💪 Only some nit left from my side.

android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt
```diff
@@ -211,6 +222,7 @@ class PlayerModule(context: ReactApplicationContext) : BitmovinBaseModule(contex
      */
     @ReactMethod
     fun destroy(nativeId: NativeId, promise: Promise) {
+        mediaSessionPlaybackManager.destroy(nativeId)
```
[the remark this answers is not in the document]
123mpozzi: We spotted a bug here: the media session was not destroyed if put inside the resolve block.

The issue is that, at this point, the main thread is already killed from JS, so if we put this inside the resolve block, the code is not reached

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
123mpozzi: Address in b12d6b09405e5f226371433351402a081b5e60cf

CHANGELOG.md
```diff
@@ -4,15 +4,16 @@
 
 ### Added
 
-- `LockScreenControlConfig` to configure the lock screen information for the application. When `isEnabled` is `true`, the current media information will be shown on the lock-screen and within the control center
+- `MediaControlConfig` to configure the media control information for the application. When `isEnabled` is `true`, the current media information will be shown on the lock-screen, in notifications, and within the control center
```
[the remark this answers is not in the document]
123mpozzi: I'd prefer to keep it as it is since the widget also pops up in the lock-screen in Android as well

PR #541, opened 2024-10-23
Android: support Background Playback

123mpozzi:
## Description

Support background playback on Android as well

## Changes

- Use a service to keep the player alive via a strong reference when the app goes into background
  - Update app lifecycle as well, to prevent the playback to automatically pause
  - Update Android's manifest to allow the service to run
- Support `playerConfig.playbackConfig.isBackgroundEnabled` on Android as well
- Introduce a new sample: **Background Playback**

## Testing

Please, if possible, do some manual testing on a physical device 🙏 

#### Background Playback (use the new sample)
- minimizing the app does not stop the playback
  - when re-opening the app, the player data is synced (e.g. current playback time)
> Note: the "Background Playback" sample will destroy the player instance (hence stopping background playback) when going back to the homepage. That is expected, a more complex sample could be published later on.

#### Activity lifecycle
- basic samples have no background playback
  - by default
  - after running the new "Background Playback" sample
- basic samples' playerview works correctly after running the new "Background Playback" sample

## Checklist
- [x] 🗒 `CHANGELOG` entry

PR #542, opened 2024-10-23
Android: mediaSession - support simple background playback as well

123mpozzi:
## Description

As of the current integration, background playback is only supported through media session.
This PR allows to run playback in the background without having to create any media session.

### Related work
#541 implements background-playback independently from media session.
Since media session is a subset of background playback, the idea is to unify and re-use most of the logic

## Checklist
- [x] 🗒 `CHANGELOG` entry

### DIFF
--- a/android/src/main/java/com/bitmovin/player/reactnative/BackgroundPlaybackManager.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/BackgroundPlaybackManager.kt
@@ -1,6 +1,5 @@
 package com.bitmovin.player.reactnative
 
-
 import android.content.ComponentName
 import android.content.Context
 import android.content.Intent
@@ -15,16 +14,14 @@ import com.facebook.react.bridge.*
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.asStateFlow
 
-// TODO: rename BackgroundPlaybackConnectionManager as both are form of bg playback
-// TODO: mediaSessionPlayback<whatever> to mediaSession<whatever>
-class MediaSessionConnectionManager(val context: ReactApplicationContext) {
+class BackgroundPlaybackManager(val context: ReactApplicationContext) {
     private var isServiceStarted = false
     private lateinit var playerId: NativeId
 
     private val _serviceBinder = MutableStateFlow<PlayerServiceBinder?>(null)
     val serviceBinder = _serviceBinder.asStateFlow()
 
-    inner class BackgroundPlaybackServiceConnection: ServiceConnection {
+    inner class BackgroundPlaybackServiceConnection : ServiceConnection {
         override fun onServiceConnected(className: ComponentName, service: IBinder) {
             val binder = service as PlayerServiceBinder
             _serviceBinder.value = binder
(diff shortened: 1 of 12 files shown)


example/android/app/src/main/AndroidManifest.xml
```diff
@@ -62,5 +62,11 @@
           </intent-filter>
       </service>
 
+        <service
+            android:name="com.bitmovin.player.reactnative.services.BackgroundPlaybackService"
+            android:foregroundServiceType="mediaPlayback"
+            android:exported="true">
+        </service>
+
```
123mpozzi, line 70: **Note**
This is necessary also for correctly updating the playerview after the app has been minimized. In fact, lifecycle changes seems to not be enough without this (the whole playerview will be black).

android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt
```diff
@@ -80,6 +81,11 @@ class PlayerModule(context: ReactApplicationContext) : BitmovinBaseModule(contex
         val defaultMetadata = analyticsConfigJson?.getMap("defaultMetadata")?.toAnalyticsDefaultMetadata()
         isMediaSessionPlaybackEnabled = playerConfigJson?.getMap("lockScreenControlConfig")
             ?.toLockScreenControlConfig()?.isEnabled ?: false
+        isBackgroundPlaybackEnabled = if (isMediaSessionPlaybackEnabled) {
+            isMediaSessionPlaybackEnabled
+        } else {
+            playerConfigJson?.getMap("playbackConfig")?.getBoolean("isBackgroundPlaybackEnabled") ?: false
+        }
```
123mpozzi, line 88: ### Open Question about Config Design

Currently, I am making the media session flag forcibly treat the background flag as true when enabled, but that still has to be decided.

This way makes sense for me since media-session is a form of background-playback.

android/src/main/java/com/bitmovin/player/reactnative/services/MediaSessionPlaybackService.kt
```diff
@@ -1,15 +1,15 @@
...
 
 import android.content.Intent
-import android.os.Binder
 import android.os.IBinder
 import com.bitmovin.player.api.Player
 import com.bitmovin.player.api.media.session.MediaSession
 import com.bitmovin.player.api.media.session.MediaSessionService
+import com.bitmovin.player.reactnative.PlayerServiceBinder
 
 class MediaSessionPlaybackService : MediaSessionService() {
```
123mpozzi, line 10: ### Open Question about Services Design

Unfortunately, I don't see a clean way of unifying the two services (maybe someone more proficient in Kotlin can help tho 🙏 )

I see mainly 2 drawbacks when unifying the services:
- a media session is always initialized in `onCreate`. Even if I pass data via the intent, `onBind` will be called after `onCreate` from my understanding, so the 'fake' media session is already created
- here we inherit from `MediaSessionService()`. This is not necessary for simple background-playback. Simple backgorund playback, on the other hand, only needs `Service()`, without the additional wrappers

android/src/main/java/com/bitmovin/player/reactnative/BackgroundPlaybackManager.kt
```diff
@@ -52,3 +57,5 @@ class MediaSessionConnectionManager(val context: ReactApplicationContext) {
         playerModule: PlayerModule? = context.playerModule,
     ): Player = playerModule?.getPlayerOrNull(nativeId) ?: throw IllegalArgumentException("Invalid PlayerId $nativeId")
 }
+
+open class PlayerServiceBinder(open var player: Player?) : Binder()
```
123mpozzi, line 61: I am not sure if this is the best way for extending `Binder()` for our use-case. I would like input here from someone more knowledgeable in Kotlin 🙏

PR #550, opened 2024-10-31
Add Basic TV playback sample

zigavehovec:
## Changes

- Add TV sample showcasing how to set up the TV UI
- Remove the previous auto TV UI in BasicPlayback sample

I placed the TV sample as the second item so that it's easily noticeable 
image

The sample will also render the TvUI on phones: 
image

## Checklist
- [ ] 🗒 `CHANGELOG` entry

### DIFF
--- a/example/src/App.tsx
+++ b/example/src/App.tsx
@@ -65,6 +65,7 @@ export type RootStackParamsList = {
 const RootStack = createNativeStackNavigator<RootStackParamsList>();
 
 const isTVOS = Platform.OS === 'ios' && Platform.isTV;
+const isAndroidTV = Platform.OS === 'android' && Platform.isTV;
 
 export default function App() {
   useEffect(() => {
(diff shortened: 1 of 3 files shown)


example/src/App.tsx
```diff
@@ -186,6 +192,11 @@ export default function App() {
           component={BasicPlayback}
           options={{ title: 'Basic playback' }}
         />
+        <RootStack.Screen
+          name="BasicTvPlayback"
+          component={BasicTvPlayback}
+          options={{ title: 'Basic TV playback' }}
+        />
```
rolandkakonyi, line 199: What about having this on the list only when running on an Android TV?
Especially as this has no useful effect on other platforms.

```suggestion
        {Platform.OS === 'android' && Platform.isTV && (
          <RootStack.Screen
            name="BasicTvPlayback"
            component={BasicTvPlayback}
            options={{ title: 'Basic TV playback' }}
          />
        )}
```

strangesource (approved):
Nice! Good starting point to add additional TV specific features in the future. 🎉

example/src/App.tsx
zigavehovec: Good idea!

zigavehovec: https://github.com/bitmovin/bitmovin-player-react-native/pull/550/commits/4a6e7ab222ca937df91c0d67be06c4de9e81e9b8

rolandkakonyi (approved):

PR #559, opened 2024-11-13
Null safe check if `isBackgroundPlaybackEnabled` property is set

strangesource:
## Description
`isBackgroundPlaybackEnabled` was checked in an unsafe way which caused attaching a player to a view to crash if  `isBackgroundPlaybackEnabled` was not set.

## Changes
- make `isBackgroundPlaybackEnabled` check null safe

## Checklist
- [x] 🗒 `CHANGELOG` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [Unreleased]
+
+### Fixed
+
+- Error where setting `PlaybackConfig.isAutoplayEnabled = true` causes the player view creation to fail on Android
+
 ## [0.31.0] - 2024-11-07
 
 ### Added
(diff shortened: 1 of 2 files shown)


strangesource:
@123mpozzi is it fine that the property is not set in this scenario or do we have to have a closer look why it is unset?

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## [Unreleased]
+
+### Fixed
+
+- Error where setting `PlaybackConfig.isAutoplayEnabled = true` causes the player view creation to fail on Android
+
```
123mpozzi, line 8: Nit -- Wording: isn't the crash happening when the flag is not set?

```suggestion
- Error where not setting `PlaybackConfig.isAutoplayEnabled` causes the player view creation to fail on Android

```

123mpozzi (approved):
Nice catch!! 💪 
The flag should absolutely be null safe

CHANGELOG.md
strangesource: No, the crash happens on the `isBackgroundPlaybackEnabled` flag if `isAutoplayEnabled` is set to true.

123mpozzi: Ah I see, thanks for clarifying

PR #568, opened 2024-11-22
Introduce Debug Logging

123mpozzi:
## Description

Debug logging provides detailed information primarily for debugging purposes, helping to diagnose problems and trace the flow of execution within the Player.

## Changes

New **static** APIs to check and set whether debug logging is enabled:
- `isDebugLoggingEnabled`
- `setDebugLogging(value: boolean)`

_Internally, this has been handled by adding a new module, because this is not a config like the others since it is static/global._

## Test and Basic Usage

> [!IMPORTANT]
> For iOS, Use player version `3.77.3-a.1` or newer

Add the following to BasicPlayback:

```diff
// other imports ...
import { useTVGestures } from '../hooks';

+ DebugConfig.setDebugLogging(true); // Add it right after the imports
```

Can also leverage a player event to test the flag:
```typescript
  const onPaused = useCallback((_: Event) => {
    if (DebugConfig.isDebugLoggingEnabled) {
      DebugConfig.setDebugLogging(false);
    } else {
      DebugConfig.setDebugLogging(true);
    }
  }, []);
```
> Changing flag value after initialization, however, does not work for Android because of a limitation

## Checklist
- [x] 🗒 `CHANGELOG` entry


123mpozzi:
We should merge only after iOS SDK version `3.78.0`

src/debug.ts
```diff
@@ -0,0 +1,56 @@
...
+   * ## Usage Notes
+   * ---
+   * - We recommend setting this flag during your app's initialization phase, such as in the
+   *   `Application.onCreate` function on Android or equivalent initialization on iOS.
+   *
+   * @defaultValue `false`
+   */
+  static async setDebugLogging(value: boolean): Promise<void> {
+    DebugConfig._isDebugEnabled = value;
+    await DebugModule.setLoggingEnabled(value);
```
rolandkakonyi, line 54: What happened with the fire and forget option?
Did you drop that idea? Or didn't work?

123mpozzi: That did not work in the end

src/debug.ts
```diff
@@ -0,0 +1,56 @@
...
+   *
+   * ## Limitations
+   * ---
+   * **Android**
+   * - This flag **must** be set **before** creating any Bitmovin component to take effect.
+   *
+   * ## Usage Notes
+   * ---
+   * - We recommend setting this flag during your app's initialization phase, such as in the
+   *   `Application.onCreate` function on Android or equivalent initialization on iOS.
```
rolandkakonyi, line 48: This doesn't translate to React Native as is.

We should phrase this differently, like:
```suggestion
   * - We recommend setting this flag during your app's initialization phase, such as in the
   *   application's entry point (e.g. `App.tsx`).
```

src/debug.ts
rolandkakonyi: What was the issue? Can you elaborate?

src/debug.ts
123mpozzi: Nice catch!

src/debug.ts
123mpozzi: You get the classic red screen because the `set` function is supposed to return a promise, and instead it does not. You can try it if you want, it should take little time to test

ios/DebugModule.swift
```diff
@@ -0,0 +1,38 @@
...
+    /// - Parameter enabled: Whether to set verbose logging as enabled or disabled.
+    @objc(setLoggingEnabled:)
+    func setLoggingEnabled(enabled: Bool) {
+        bridge.uiManager.addUIBlock { [weak self] _, _ in
+            let defaultMinimumLevel: LogLevel = .warning
+            if enabled {
+                DebugConfig.logging.logger?.level = .verbose
+            } else {
+                DebugConfig.logging.logger?.level = defaultMinimumLevel
+            }
```
rolandkakonyi, line 35: This can be simplified as:
```suggestion
            DebugConfig.logging.logger?.level = enabled ? .verbose : .warning
```

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## [Unreleased]
+
+### Added
+
+- `isDebugLoggingEnabled` and `setDebugLogging(value: boolean)` to check and set whether debug logging is enabled. Debug logging helps diagnose problems and trace the flow of execution within the Player and **should not be enabled in production** as it may log sensitive or confidential information to the console.
```
rolandkakonyi, line 7:
```suggestion
- `DebugConfig.isDebugLoggingEnabled` and `DebugConfig.setDebugLogging(value: boolean)` to check and set whether debug logging is enabled. Debug logging helps diagnose problems and trace the flow of execution within the Player and **should not be enabled in production** as it may log sensitive or confidential information to the console.
```

src/debug.ts
```diff
@@ -0,0 +1,56 @@
...
+   * - This flag **must** be set **before** creating any Bitmovin component to take effect.
+   *
+   * ## Usage Notes
+   * ---
+   * - We recommend setting this flag during your app's initialization phase, such as in the
+   *   application's entry point (e.g. `App.tsx`).
+   *
+   * @defaultValue `false`
+   */
+  static async setDebugLogging(value: boolean): Promise<void> {
```
rolandkakonyi, line 52: Just noticed that the naming here is inconsistent.
Getter is `isDebugLoggingEnabled` while the setter is `setDebugLogging`.
It should be probable `setDebugLoggingEnabled`.

Please follow up with all occurrences in both docs and changelog entry.

ios/DebugModule.swift
```diff
@@ -0,0 +1,33 @@
...
+    public var methodQueue: DispatchQueue! {
+        bridge.uiManager.methodQueue
+    }
+}
+
+extension DebugModule {
+    /// Enable/disable verbose logging for the console logger.
+    /// - Parameter enabled: Whether to set verbose logging as enabled or disabled.
+    @objc(setLoggingEnabled:)
+    func setLoggingEnabled(enabled: Bool) {
```
rolandkakonyi, line 28: This should be `setDebugLoggingEnabled`, we only say logging

src/debug.ts
123mpozzi: Address in 6ccf48a777249ede7d1d107d255ee9b2320f0b77

ios/DebugModule.swift
123mpozzi: Address in d1b0ed2232af581273c5e7a03380f27d484d4a5e , also for Android

rolandkakonyi (approved):
Changes and testing looks good!

src/debug.ts
```diff
@@ -0,0 +1,56 @@
...
+   * - We recommend setting this flag during your app's initialization phase, such as in the
+   *   application's entry point (e.g. `App.tsx`).
+   *
+   * @defaultValue `false`
+   */
+  static async setDebugLoggingEnabled(value: boolean): Promise<void> {
+    DebugConfig._isDebugEnabled = value;
+    await DebugModule.setDebugLoggingEnabled(value);
+  }
+}
```
123mpozzi, line 56: TypeDoc note: I've discovered we can use [@defaultValue](https://typedoc.org/tags/defaultValue/) 😮

PR #593, opened 2024-12-20
Expose Android SurfaceType

krocard:
## Problem (untracked)
Android `SurfaceType` was not exposed.

## Changes
Add it to the TS PlayerViewConfig


zigavehovec (approved):
👌

PR #596, opened 2024-12-24
Feat/video adaptation

saravanans-github:
## Description
Added support for Android onVideoAdaptation callback.
## Changes
In a similar design to the NetworkConfig I abstracted the AdaptationConfig to an AdaptationModule to support the callback 
## Checklist
- [ ] 🗒 `CHANGELOG` entry


src/adaptation/index.ts
```diff
@@ -0,0 +1,69 @@
+import { NativeModules } from 'react-native';
+import BatchedBridge from 'react-native/Libraries/BatchedBridge/BatchedBridge';
+import NativeInstance from '../nativeInstance';
+import { VideoAdaptationData, AdaptationConfig } from './adaptationConfig';
+
+// Export config types from Adaptation module.
+export { VideoAdaptationData, AdaptationConfig };
```
rolandkakonyi, line 7: Hi @saravanans-github!
To fix the documentation error, and possibly a potential integration error regarding imports, please replace this line here with the suggestion.

```suggestion
export * from './adaptationConfig';
```

The reason is that all exported types should be re-exported on the top-level and now this is not the case.

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 ## [Unreleased]
 
+### Changed
+
+- Android: Refactored AdaptationConfig to enable the `onVideoAdaptation` callback
```
zigavehovec, line 7: I think this changelog isn't necessary, as the existing api's can be called the same way as before, or am I missing something?

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
...
+
+### Added
+
+- Android: Added support for the `onVideoAdaptation` callback
```
zigavehovec, line 11: What do you think about this?
```suggestion
- Android: Added `AdaptationConfig.videoAdaptation` to customize the player's adaptation logic
```

android/src/main/java/com/bitmovin/player/reactnative/AdaptationModule.kt
```diff
@@ -0,0 +1,99 @@
...
+
+    @ReactMethod
+    fun destroy(nativeId: NativeId) {
+        adaptationConfigs.remove(nativeId)
+        onVideoAdaptationCompleters.keys.filter { it.startsWith(nativeId) }.forEach {
+            onVideoAdaptationCompleters.remove(it)
+        }
+    }
+
+    private fun initConfigBlocks(nativeId: String, config: ReadableMap) {
```
zigavehovec, line 52: I would just directly call `initVideoAdaptationData` from `initWithConfig`. I guess this pattern comes form NetworkModule, where there were multiple init blocks to trigger.

android/src/main/java/com/bitmovin/player/reactnative/AdaptationModule.kt
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    private fun initVideoAdaptationData(nativeId: NativeId, adaptationConfigJson: ReadableMap) {
+        val adaptationConfig = getConfig(nativeId) ?: return
+        if (!adaptationConfigJson.hasKey("videoAdaptation")) return
+
+        adaptationConfig.videoAdaptation = VideoAdaptation { data ->
+            val future = onVideoAdaptationFromJS(nativeId, data)
+            try {
+                val callbackValue = future.get(1, TimeUnit.SECONDS) // set timeout to mimize playback performance impact
```
zigavehovec, line 63: Have you seen problems with the promise taking longer to resolve? Is this really needed, I don't see a reason why this could take more than a second, do we really need this?

If so, we might want to consider if we need this also in other similar APIs like `preprocessHttpResponseCallback` and `preprocessHttpRequestCallback` in the NetworkModule - not a part of this PR of course.

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
```diff
@@ -144,11 +145,17 @@ private fun String.toTimelineReferencePoint(): TimelineReferencePoint? = when (t
 /**
  * Converts an arbitrary `json` to `AdaptationConfig`.
  */
-private fun ReadableMap.toAdaptationConfig(): AdaptationConfig = AdaptationConfig().apply {
+fun ReadableMap.toAdaptationConfig(): AdaptationConfig = AdaptationConfig().apply {
     withInt("maxSelectableBitrate") { maxSelectableVideoBitrate = it }
     withInt("initialBandwidthEstimateOverride") { initialBandwidthEstimateOverride = it.toLong(); }
 }
 
+fun ReadableMap.toVideoAdaptationData(): VideoAdaptationData? {
```
zigavehovec, line 153: This is not used, do is it a leftover or are we missing something?

src/adaptation/index.ts
```diff
@@ -0,0 +1,69 @@
...
+  /**
+   * Applies the user-defined `onVideoAdaptation` function to native's `data` and store
+   * the result back in `AdaptationModule`.
+   *
+   * Called from native code when `AdaptationConfig.videoAdaptation` is dispatched.
+   *
+   * @param requestId Passed through to identify the completion handler of the request on native.
+   * @param data The quality according to VideoQuality.id that is returned will be downloaded next.
+   */
+
```
zigavehovec, line 58:
```suggestion
```

zigavehovec (commented):
Nice work! Mostly nitpicks from me 👍

PR #597, opened 2025-01-14
Replace Akamai CDN URLs with Bitmovin CDN URLs

rolandkakonyi:
## Description

We are moving away from Akamai CDN in favor of Bitmovin CDN.

## Changes

Updated URLs to use the same assets from the Bitmovin CDN.

## Checklist
- [x] 🗒 `CHANGELOG` entry - not needed


123mpozzi (approved):
Nice job! 💪

PR #598, opened 2025-01-15
Fix UninitializedPropertyAccessException when destroying the Player

testcenter:
## Description
When a player with the config `mediaControlConfig.isEnabled = false` is created and the player gets destroyed, it will throw an `UninitializedPropertyAccessException` while trying to destroy the `MediaSessionPlaybackManager`

## Changes
- Change `MediaSessionPlaybackManager.playerId` from `lateinit` to nullable.
- Remove unnecessary `resolveOnUiThread` call

## Checklist
- [x] 🗒 `CHANGELOG` entry

### DIFF
--- a/android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt
@@ -97,9 +97,7 @@ class PlayerModule(context: ReactApplicationContext) : BitmovinBaseModule(contex
         }
 
         if (enableMediaSession) {
-            promise.unit.resolveOnUiThread {
-                mediaSessionPlaybackManager.setupMediaSessionPlayback(nativeId)
-            }
+            mediaSessionPlaybackManager.setupMediaSessionPlayback(nativeId)
         }
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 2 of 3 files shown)


rolandkakonyi (approved):

android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt
```diff
@@ -97,9 +97,7 @@ class PlayerModule(context: ReactApplicationContext) : BitmovinBaseModule(contex
         }
 
         if (enableMediaSession) {
-            promise.unit.resolveOnUiThread {
-                mediaSessionPlaybackManager.setupMediaSessionPlayback(nativeId)
-            }
+            mediaSessionPlaybackManager.setupMediaSessionPlayback(nativeId)
         }
```
strangesource, line 101: Why this change? 🤔

strangesource (approved):

android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt
testcenter: it is not necessary, as we are already running inside the uiThread, see [the first line of the function](https://github.com/bitmovin/bitmovin-player-react-native/pull/598/files/6c7c2c7757051e0b3ccdaae94eaff1443291df79#diff-dc6df38ca918c4ebe88d22bf0fc3b8c3a8d000be83da6d9701e9dbaac459cfc8L70)

PR #613, opened 2025-03-06
PRN-164 - Add `CueEnterEvent.image` and `CueExitEvent.image`

rolandkakonyi:
## Description

We don't expose the image from the cues with subtitle related events.

## Changes

Added support for exposing the `image` property from the native `CueEnterEvent` and `CueExitEvent`.

## Checklist
- [x] 🗒 `CHANGELOG` entry

### DIFF
--- a/src/events.ts
+++ b/src/events.ts
@@ -736,7 +736,7 @@ export interface CueEnterEvent extends Event {
   /**
    * Data URI for image data of this subtitle.
    */
-  imageUri?: string;
+  image?: string;
 }
 
 /**
--- a/src/events.ts
+++ b/src/events.ts
@@ -733,6 +733,10 @@ export interface CueEnterEvent extends Event {
    * The textual content of this subtitle.
    */
   text?: string;
+  /**
+   * Data URI for image data of this subtitle.
+   */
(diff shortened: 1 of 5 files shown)


123mpozzi (approved):
Good job! 💪

src/events.ts
```diff
@@ -733,6 +733,10 @@ export interface CueEnterEvent extends Event {
    * The textual content of this subtitle.
    */
   text?: string;
+  /**
+   * Data URI for image data of this subtitle.
+   */
+  image?: string;
```
krocard, line 739: Are we not constructing the images for performance reasons? Aka:
```suggestion
  image?: Image;
```

krocard (approved):
Looks good. I'm wondering why we are not building the Images instead of exposing them as URIs.

Please add a test

rolandkakonyi:
> Looks good. I'm wondering why we are not building the Images instead of exposing them as URIs.
> 
> **Please add a test**

1. React Native doesn't support sending `Image` type over events. Also our Web player exposes the data URI the same way in the `image` property.
2. We don't have tests in this project.

PR #658, opened 2025-07-14
Add expect helper for integration tests and add more tests

rolandkakonyi:
## Description

We don't have `expect` integration into the integration tests as `expect()` library doesn't work with React Native when running the tests on the mobile platforms. (They only work for unit tests that run on the host machine)

## Changes

Found a nice and lightweight `expect` implementation to allow us adding expectations.

+1 added tests utilizing these.

## Checklist


integration_test/package.json
```diff
@@ -7,10 +7,11 @@
...
-    "start-test:ios": "yarn cavy run-ios --simulator \"$(xcrun simctl list devices available -e -j | jq --raw-output -s '[.[].devices.[].[] | select( .deviceTypeIdentifier | contains(\"iPhone\") )|.name] | first')\"",
+    "start-test:android": "yarn cavy run-android --terminal bash --deviceId \"$([ -z \"$(adb devices | grep -v List | grep device | grep -m 1 emulator | cut -f 1)\" ] && (emulator -list-avds | head -n 1 | xargs -I NAME emulator -avd NAME 1> /dev/null 2> /dev/null & adb wait-for-device shell 'while [[ -z $(getprop sys.boot_completed) ]]; do sleep 1; done;'); (adb devices | grep -v List | grep device | grep -m 1 emulator | cut -f 1))\"",
+    "start-test:ios": "yarn cavy run-ios --terminal bash --simulator \"$(xcrun simctl list devices available -e -j | jq --raw-output -s '[.[].devices.[].[] | select( .deviceTypeIdentifier | contains(\"iPhone\") )|.name] | first')\"",
```
rolandkakonyi, line 14: This fixes running the test in some terminal apps, like Ghostty

integration_test/tests/advertisingTest.ts
```diff
@@ -153,29 +158,60 @@ export default (spec: TestScope) => {
...
         });
       });
     });
   }
 
-  advertisingErrorTests(
-    { sources: [{ tag: AdTags.error, type: AdSourceType.IMA }] },
-    'IMA'
-  );
+  advertisingErrorTests(() => {
```
rolandkakonyi, line 206: Changed to always fetch new tags with new correlator to avoid IMA not playing the same URL

integration_test/tests/helper/Expect.ts
```diff
@@ -0,0 +1,208 @@
+export function expect(actual: any, desc?: string) {
```
rolandkakonyi, line 1: New helper to allow using `expect(..).to...` in tests

integration_test/tests/helper/Sources.ts
```diff
@@ -11,8 +11,8 @@ export const Sources = {
     type: SourceType.DASH,
   } as SourceConfig,
 
-  akamaiTestLiveHls: {
-    url: 'https://cph-msl.akamaized.net/hls/live/2000341/test/master.m3u8',
+  iReplayTestLiveHls: {
```
rolandkakonyi, line 14: Replaced outdated stream

integration_test/tests/helper/Expect.ts
```diff
@@ -0,0 +1,208 @@
...
+        const errorMessage = `Expected ${actual} to have length ${expected}, but got ${
+          actual ? actual.length : 'undefined'
+        }`;
+        console.error(`[EXPECTATION NOT MET]: ${errorMessage}`);
+        throwErr(errorMessage);
+      }
+    },
+
+    toHaveProperty(property: string) {
+      if (actual && actual.hasOwnProperty(property))
```
123mpozzi, line 161: Can we go through the object's prototype so that we are sure we're calling the native `hasOwnProperty`, only in case it exists?

Applies to `toHavePropertyValue` as well.

```suggestion
      if (actual && Object.prototype.hasOwnProperty.call(actual, property))
```

---

Otherwise this can fail in edge cases like:
```ts
const noProto = Object.create(null);
noProto.key = 123;
console.log(noProto.hasOwnProperty('key')); // Error: noProto.hasOwnProperty is not a function

const overridden = {
  key: 123,
  hasOwnProperty: () => false,
};
console.log(overridden.hasOwnProperty('key')); // false

console.log(Object.prototype.hasOwnProperty.call(overridden, 'key')) // true
console.log(Object.prototype.hasOwnProperty.call(noProto, 'key'))    // true
```

integration_test/tests/helper/Expect.ts
```diff
@@ -0,0 +1,208 @@
...
+        const errorMessage = `Expected ${actual} to have property '${property}' with value ${value}, but got ${actualValue}`;
+        console.error(`[EXPECTATION NOT MET]: ${errorMessage}`);
+        throwErr(errorMessage);
+      }
+    },
+
+    toMatchRegex(regex: RegExp) {
+      if (regex.test(actual)) logPass(`${actual} matches regex ${regex}`);
+      else {
+        const errorMessage = `Expected ${actual} to match regex ${regex}`;
```
123mpozzi, line 188: Nit: we might want to explicitly convert `actual` to a string before calling `.test()` to avoid unexpected behavior when actual is a number, object, null, etc.

```suggestion
      const actualStr = String(actual);
      if (regex.test(actualStr)) logPass(`${actualStr} matches regex ${regex}`);
      else {
        const errorMessage = `Expected ${actualStr} to match regex ${regex}`;
```

integration_test/tests/helper/Expect.ts
```diff
@@ -0,0 +1,208 @@
...
+        console.error(`[EXPECTATION NOT MET]: ${errorMessage}`);
+        throwErr(errorMessage);
+      }
+    },
+
+    toBeCloseTo(expected: number, precision: number = 2) {
```
123mpozzi, line 194: Nit: here as well, we could check for `actual` to be a number before doing any math:

```suggestion
    toBeCloseTo(expected: number, precision: number = 2) {
        if (typeof actual !== 'number') {
            throwErr(`Both ${actual} and ${expected} must be numbers`);
        }
```

integration_test/tests/captionTest.ts
```diff
@@ -36,5 +38,92 @@ export default (spec: TestScope) => {
...
+        expect(
+          typeof cueEnterEvent.timestamp,
+          'timestamp should be a number'
+        ).toBe('number');
+        expect(
+          cueEnterEvent.timestamp,
+          'timestamp should be > 0'
+        ).toBeGreaterThan(0);
+      });
+    });
```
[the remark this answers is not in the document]
rolandkakonyi: Added more here: be063a91665583eaabc4cc981ae0b740b48f4b92

integration_test/tests/helper/Expect.ts
```diff
@@ -0,0 +1,208 @@
...
+    toBeNull() {
+      if (actual === null) logPass(`${actual} is null`);
+      else {
+        const errorMessage = `Expected ${actual} to be null`;
+        console.error(`[EXPECTATION NOT MET]: ${errorMessage}`);
+        throwErr(errorMessage);
+      }
+    },
+
+    toBeNotNull() {
```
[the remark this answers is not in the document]
rolandkakonyi: 3b93a8f8

integration_test/tests/helper/Expect.ts
```diff
@@ -0,0 +1,208 @@
...
+
+    toEqual(expected: any) {
+      if (JSON.stringify(actual) === JSON.stringify(expected))
+        logPass(`Expected ${actual} to equal ${expected}`);
+      else {
+        const errorMessage = `Expected ${actual} to equal ${expected}`;
+        console.error(`[EXPECTATION NOT MET]: ${errorMessage}`);
+        throwErr(errorMessage);
+      }
+    },
```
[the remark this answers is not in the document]
rolandkakonyi: added in ba5b2760

integration_test/tests/helper/Expect.ts
```diff
@@ -0,0 +1,208 @@
...
+      if (actual !== expected) logPass(`${actual} != ${expected}`);
+      else {
+        const errorMessage = `Expected ${actual} not to be ${expected}`;
+        console.error(`[EXPECTATION NOT MET]: ${errorMessage}`);
+        throwErr(errorMessage);
+      }
+    },
+
+    toEqual(expected: any) {
+      if (JSON.stringify(actual) === JSON.stringify(expected))
```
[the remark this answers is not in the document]
123mpozzi: Great find for that lib!

integration_test/tests/helper/Expect.ts
```diff
@@ -0,0 +1,307 @@
...
+      assert(
+        diff <= tolerance,
+        `Expected ${formatValue(actual)} to be close to ${formatValue(
+          expected
+        )} (precision: ${precision}), but difference is ${diff}`
+      );
+    },
+
+    // Async assertion support
+    async toResolve() {
```
123mpozzi, line 239: Non-promise values seem to pass this check as well.

I would propose to type-check this function too to avoid silently accepting non-promises.

```suggestion
    async toResolve() {
      if (!(actual instanceof Promise)) {
        throwErr(`Expected a Promise to resolve, but got a non-promise: ${actual}`);
      }
```

---

Without type-checking:

```ts
async function testToResolve(actual: any) {
  try {
    const result = await Promise.resolve(actual);
    console.error(`Expected resolve, got: ${result}`);
  } catch (error) {
    console.log(`Expected promise to resolve, but it rejected with: ${error}`);
  }
}

(async () => {
  console.log("\nCase 1: Rejected promise");
  await testToResolve(Promise.reject("boom")); // Rejected

  console.log("\nCase 2: Resolved promise");
  await testToResolve(Promise.resolve("ok")); // Resolved

  console.log("\nCase 3: Non-promise value");
  await testToResolve(123); // Resolved   <----
})();
```

123mpozzi (commented):
Good job! I have just left one last comment

integration_test/tests/helper/Expect.ts
rolandkakonyi: fixed in 8faf3e73

123mpozzi (approved):

PR #696, opened 2025-09-01
Feature/fix live stream crash react native reanimated

rolandkakonyi:
When using `react-native-reanimated` along with our library, crashes happen during JSON serialization due to an infinite value for duration.

## Changes

Introduced sanitation JSON event payload values to avoid crashes.

## Checklist
- [x] 🗒 `CHANGELOG` entry

### DIFF
--- a/android/src/main/java/com/bitmovin/player/reactnative/util/NonFiniteSanitizer.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/util/NonFiniteSanitizer.kt
@@ -1,0 +1,35 @@
+package com.bitmovin.player.reactnative.util
+
+object NonFiniteSanitizer {
+  fun sanitize(value: Any?): Any? = when (value) {
+    null -> null
+    is Double -> if (value.isFinite()) value else toSentinel(value)
+    is Float -> {
+      val d = value.toDouble()
+      if (d.isFinite()) d else toSentinel(d)
+    }
+    is Map<*, *> -> value.entries.associate { (k, v) -> k.toString() to sanitize(v) }
+    is List<*> -> value.map { sanitize(it) }
+    is Array<*> -> value.map { sanitize(it) }
+    is Number -> value // Int/Long/Short/Byte
+    else -> value
+  }
+
+  /**
+   * Type-safe method specifically for event data maps.
+   * Sanitizes event data while preserving type safety and handling null filtering.
+   */
+  fun sanitizeEventData(eventData: Map<String, Any>): Map<String, Any> {
+    val sanitized = mutableMapOf<String, Any>()
+    for ((key, value) in eventData) {
+      sanitize(value)?.let { sanitized[key] = it }
+    }
+    return sanitized
+  }
+
+  private fun toSentinel(d: Double): String = when {
+    d == Double.POSITIVE_INFINITY -> "Infinity"
+    d == Double.NEGATIVE_INFINITY -> "-Infinity"
+    else -> "NaN"
+  }
+}
--- a/ios/NonFiniteSanitizer.swift
+++ b/ios/NonFiniteSanitizer.swift
@@ -4,19 +4,19 @@ enum NonFiniteSanitizer {
   // Keep existing generic method for backward compatibility
   static func sanitize(_ value: Any) -> Any {
     switch value {
-    case let d as Double:
-      if d.isFinite { return d }
-      return toSentinel(d)
-    case let f as Float:
-      let d = Double(f)
-      if d.isFinite { return d }
-      return toSentinel(d)
+    case let doubleValue as Double:
+      if doubleValue.isFinite { return doubleValue }
+      return toSentinel(doubleValue)
+    case let floatValue as Float:
+      let doubleValue = Double(floatValue)
+      if doubleValue.isFinite { return doubleValue }
+      return toSentinel(doubleValue)
     case let dict as [AnyHashable: Any]:
       var out: [AnyHashable: Any] = [:]
-      for (k, v) in dict { out[k] = sanitize(v) }
+      for (ket, value) in dict { out[ket] = sanitize(value) }
       return out
-    case let arr as [Any]:
-      return arr.map { sanitize($0) }
+    case let array as [Any]:
+      return array.map { sanitize($0) }
     default:
       return value
     }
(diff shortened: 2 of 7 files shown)


android/src/main/java/com/bitmovin/player/reactnative/util/NonFiniteSanitizer.kt
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  /**
+   * Type-safe method specifically for event data maps.
+   * Sanitizes event data while preserving type safety and handling null filtering.
+   */
+  fun sanitizeEventData(eventData: Map<String, Any>): Map<String, Any> {
+    val sanitized = mutableMapOf<String, Any>()
+    for ((key, value) in eventData) {
+      sanitize(value)?.let { sanitized[key] = it }
```
123mpozzi, line 25: This drops keys with `null` values, right?
Not sure if expected since on iOS it seems they are not dropped - if that's the case, can we align on the same behaviour?

ios/NonFiniteSanitizer.swift
```diff
@@ -0,0 +1,55 @@
...
+    case let doubleValue as Double:
+      if doubleValue.isFinite { return doubleValue }
+      return toSentinel(doubleValue)
+    case let floatValue as Float:
+      let doubleValue = Double(floatValue)
+      if doubleValue.isFinite { return doubleValue }
+      return toSentinel(doubleValue)
+    case let dict as [AnyHashable: Any]:
+      var out: [AnyHashable: Any] = [:]
+      for (ket, value) in dict { out[ket] = sanitize(value) }
```
123mpozzi, line 16: Typo: 

```suggestion
      for (key, value) in dict { out[key] = sanitize(value) }
```

ios/NonFiniteSanitizer.swift
```diff
@@ -0,0 +1,55 @@
...
+    switch value {
+    case let doubleValue as Double:
+      if doubleValue.isFinite { return doubleValue }
+      return toSentinel(doubleValue)
+    case let floatValue as Float:
+      let doubleValue = Double(floatValue)
+      if doubleValue.isFinite { return doubleValue }
+      return toSentinel(doubleValue)
+    case let dict as [AnyHashable: Any]:
+      var out: [AnyHashable: Any] = [:]
```
123mpozzi, line 15: I see on Android we convert keys to String, How do you feel about converting them here as well instead of keeping it `AnyHashable`? This would achieve key/value handling parity between the platforms

src/utils/normalizeNonFinite.ts
```diff
@@ -0,0 +1,13 @@
+export function normalizeNonFinite<T>(input: T): T {
+  function walk(v: any): any {
+    if (v === 'Infinity') return Infinity;
+    if (v === '-Infinity') return -Infinity;
+    if (v === 'NaN') return NaN;
+    if (Array.isArray(v)) return v.map(walk);
+    if (v && typeof v === 'object') {
+      for (const k of Object.keys(v)) (v as any)[k] = walk((v as any)[k]);
```
123mpozzi, line 8: I think it would be safer to avoid in-place mutation when normalizing nested objects.

Right now, in `useProxy` we shallow-copy the event, but its `nativeEvent` still holds references to nested objects. If `normalizeNonFinite` mutates those in place, we might accidentally modify objects that other listeners or RN internals are still holding on to.

Instead, we can build a fresh object during recursion to avoids side effects and keeps the normalizer a pure transformation function:

```suggestion
      const out: any = {};
      for (const k of Object.keys(v)) out[k] = walk((v as any)[k]);
      return out;
```

android/src/main/java/com/bitmovin/player/reactnative/util/NonFiniteSanitizer.kt
```diff
@@ -0,0 +1,35 @@
...
+    }
+    return sanitized
+  }
+
```
123mpozzi, line 29: Should we provide the counterpart of `sanitizeDictionary` we have in iOS to Android?

```suggestion

     /**
      * Generic method for sanitizing maps with type preservation
      */
    fun sanitizeMap(dictionary: Map<*, *>): Map<String, Any?> {
        val sanitized = mutableMapOf<String, Any?>()
        for ((key, value) in dictionary) {
            val stringKey = key?.toString() ?: continue
            sanitized[stringKey] = sanitize(value)
        }
        return sanitized
    }
    
```

123mpozzi (commented):
Looks good overall! 💪 
Just left a few minor comments

android/src/main/java/com/bitmovin/player/reactnative/util/NonFiniteSanitizer.kt
rolandkakonyi: `null` values mean no value, so this is fine. We are filtering null values already before this.

android/src/main/java/com/bitmovin/player/reactnative/util/NonFiniteSanitizer.kt
```diff
@@ -0,0 +1,35 @@
...
+object NonFiniteSanitizer {
+  fun sanitize(value: Any?): Any? = when (value) {
+    null -> null
+    is Double -> if (value.isFinite()) value else toSentinel(value)
+    is Float -> {
+      val d = value.toDouble()
+      if (d.isFinite()) d else toSentinel(d)
+    }
+    is Map<*, *> -> value.entries.associate { (k, v) -> k.toString() to sanitize(v) }
```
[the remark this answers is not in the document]
rolandkakonyi: This was a mistake, fixed in 3126b21d

android/src/main/java/com/bitmovin/player/reactnative/util/NonFiniteSanitizer.kt
```diff
@@ -0,0 +1,35 @@
+package com.bitmovin.player.reactnative.util
+
+object NonFiniteSanitizer {
+  fun sanitize(value: Any?): Any? = when (value) {
```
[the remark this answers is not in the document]
rolandkakonyi: Fixed it in 3126b21d

src/utils/normalizeNonFinite.ts
rolandkakonyi: Nice catch, fixed it in 15ec8e96

123mpozzi (approved):

rolandkakonyi:
Closes #695

PR #697, opened 2025-09-01
Automated changelog entry for SDK update

matamegger:
## Description
The automatic workflow creates the version update PR, but is missing the changelog entry.

## Changes
Introduce a python script (because the workflow is too hard for a readable bash script) to automatically insert/update the changelog entries for the SDK version update.

Tested locally and will automatically create missing sections and override the version of existing entries for the platform.
The workflow assumes that `python3` is present, which should be the case on ubuntu or mac machines (though I have not explicitly verified if this is indeed the case for all the runners)

## Checklist
- [x] 🗒 `CHANGELOG` entry


rolandkakonyi:
@matamegger Copilot review found a couple issues, please address them!

update_changelog.py
rolandkakonyi, line 1: I think the name of the script is a bit too generic for what it does in practice.

What about `update_player_sdk_update_changelog.py`?

PR #705, opened 2025-09-09
Support git worktrees

123mpozzi:
## Description

Setting up on worktrees gets you:

```
Error: Not a git repository. Please run this from the root of the project.
```

The problem is that we check for a .git directory `([ ! -d ".git" ])`, but in Git worktrees, `.git` is a file that points to the actual Git directory, not a directory itself.

## Changes

- Check for  .git directory or **file** exist instead of only checking the directory

## Checklist
- [x] 🗒 `CHANGELOG` entry - `no need since it is build tooling`

### DIFF
--- a/scripts/setup-hooks.sh
+++ b/scripts/setup-hooks.sh
@@ -3,8 +3,8 @@
 # Setup git pre-commit hooks for the project
 echo "Setting up pre-commit hooks..."
 
-# Check if .git directory exists
-if [ ! -d ".git" ]; then
+# Check if .git directory or file exists (supports both regular repos and worktrees)
+if [ ! -d ".git" ] && [ ! -f ".git" ]; then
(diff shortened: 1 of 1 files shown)


scripts/setup-hooks.sh
```diff
@@ -3,8 +3,8 @@
 # Setup git pre-commit hooks for the project
 echo "Setting up pre-commit hooks..."
 
-# Check if .git directory exists
-if [ ! -d ".git" ]; then
+# Check if .git directory or file exists (supports both regular repos and worktrees)
+if [ ! -d ".git" ] && [ ! -f ".git" ]; then
```
rolandkakonyi, line 7: nit: could be just this:

```suggestion
if [ ! -e ".git" ]; then
```

PR #845, opened 2026-02-25
iOS: Warn on `playerView` reparenting due to native IMA limitation

123mpozzi:
## Description

Re-Parenting the PlayerView is not a pattern we currently support natively on iOS. When using IMA ads, and the player view is re-parented, the ad video could show up as a black screen, hearing only the audio.

## Changes

- Detect re-parenting of the playerView, and emit a warning log message

## Testing

Apply [reparentWarning.patch](https://github.com/user-attachments/files/25549593/reparentWarning.patch):
- Landscape Fullscreen sample has ads added now: play it and look at Xcode logs - nothing should appear
- Open new sample `Modal Re-Parenting`: play it until the mid-roll at ~10 starts - log should appear in Xcode console

## Checklist
- [x] 🗒 `CHANGELOG` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,11 @@
 
 - Android: TweaksConfig option `enableDrmLicenseRenewRetry` to retry renewing a DRM license when the first renewal attempt fails
 
+### Known Issues
+
+- iOS: Moving a live `PlayerView` between different React Native view trees while using IMA ads in the same source is unsupported and can result in a black screen during playback
+  - A native warning log is now emitted when a rebind of this type is detected
+
 ## [1.9.0] - 2026-02-13
 
 ### Changed
--- a/ios/RNPlayerView.swift
+++ b/ios/RNPlayerView.swift
@@ -328,7 +328,8 @@ private extension RNPlayerView {
 
         NSLog(
             """
-            [RNPlayerView][Warning][playerId=\(playerId)] Detected this player attached to multiple RNPlayerView hosts. \
+            [RNPlayerView] [Warning] [playerId=\(playerId)] Detected this player attached \
+            to multiple RNPlayerView hosts. \
             Re-parenting a live PlayerView is unsupported and can cause black screens when paired with IMA ads. \
             Keep one stable PlayerView host for the session.
             """
--- a/ios/RNPlayerView.swift
+++ b/ios/RNPlayerView.swift
@@ -3,6 +3,15 @@ import BitmovinPlayer
 import ExpoModulesCore
 
 public class RNPlayerView: ExpoView {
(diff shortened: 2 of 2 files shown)


ios/RNPlayerView.swift
```diff
@@ -3,6 +3,16 @@ import BitmovinPlayer
 import ExpoModulesCore
 
 public class RNPlayerView: ExpoView {
+    private final class WeakViewReference {
+        weak var value: RNPlayerView?
+        init(_ value: RNPlayerView) {
+            self.value = value
+        }
+    }
```
rolandkakonyi: This could be a struct as we only need the `value` to be a reference.
Additionally the `value` property is mutable, which shouldn't be the case as we only read it or create a new weak reference.

Please use `private(set) weak var`

ios/RNPlayerView.swift
```diff
@@ -265,6 +286,55 @@ private extension RNPlayerView {
...
+
+    func maybeWarnAboutPlayerRebind(playerId: NativeId, player: Player) {
+        let existingHost = Self.playerHostRegistryLock.withLock { () -> RNPlayerView? in
+            guard let existingHost = Self.playerHostRegistry[playerId]?.value else {
+                Self.playerHostRegistry.removeValue(forKey: playerId)
+                return nil
+            }
+            return existingHost
+        }
```
rolandkakonyi: Should this be in a helper method for reading the optional value?

ios/RNPlayerView.swift
```diff
@@ -265,6 +286,55 @@ private extension RNPlayerView {
...
+        guard let existingHost else {
+            return
+        }
+        guard existingHost !== self else {
+            return
+        }
+
+        NSLog(
```
rolandkakonyi: We could use the built-in Expo logger.
```suggestion
        log.warn(
```

CHANGELOG.md
```diff
@@ -17,6 +17,11 @@
 
 - Android: TweaksConfig option `enableDrmLicenseRenewRetry` to retry renewing a DRM license when the first renewal attempt fails
 
+### Known Issues
+
+- iOS: Moving a live `PlayerView` between different React Native view trees while using IMA ads in the same source is unsupported and can result in a black screen during playback
+  - A native warning log is now emitted when a rebind of this type is detected
```
rolandkakonyi:
```suggestion
  - A native warning log is now emitted when a rebind of this kind is detected
```

rolandkakonyi (changes requested):

123mpozzi:
Reparenting a player view using IMA is going to be supported natively on iOS soon, so closing this

PR #864, opened 2026-03-12
PiP Android onPictureInPictureExit event sync with IOS flow

ibocharnikov-3ss-tv:
## Description
With the current 1.11.0 package version there is no way to distinguish 2 PiP use cases on Android:

1. When user exits from PiP view by pressing "expand to fullscreen" button on the PiP view.
2. When user exits from PiP view by pressing "close (X)" button on the PiP view.

The root cause why it is not possible to distinguish is because `onPictureInPictureExit` event listener is triggered before `AppState` would change accordingly.

## Changes
With the current change `onPictureInPictureExit` event listener would be called after AppState would change which will allow us to check if PiP view was closed completely or simply expanded to fullscreen view.

## Checklist
- [x] 🗒 `CHANGELOG` entry


testcenter:
FYI: we are working on it and try to fix it in the android native player.
I will let you know about further updates.

testcenter:
@ibocharnikov-3ss-tv
fyi:
We added `PictureInPictureEntered`  and `PictureInPictureExited` events in the native player (similar to the events already available on iOS).
Those events + a small lifecycle flow update are part of this PR: https://github.com/bitmovin/bitmovin-player-react-native/pull/927

The change in this PR should make your changes obsolete. As you can track the AppState and check the current AppState in the `PiPEntered` or `PiPExited` events then. Can you please confirm this indeed fixes your problem?

testcenter:
fyi, we fixed a known limitation (known since the PR mentioned in the previous comment):
With the latest release ([v1.20.1](https://github.com/bitmovin/bitmovin-player-react-native/releases/tag/v1.20.1)) the issue should be resolved.
@ibocharnikov-3ss-tv can you please verify?

testcenter:
closing this, as there is no input since weeks.
in case this is still important, re-open the PR

PR #885, opened 2026-03-24
Adjust DebugConfig for JS console.log calls.

ibocharnikov-3ss-tv:
## Description
There were a set of unguarded `console.log` calls for every HTTP request/response, they were triggered in production/release mode. On Android, together with perf. monitoring it is causing performance issues and slowness during the playback.

## Changes
DebugConfig class was adjusted to support logging on JS side same way it supports it on IOS/Android.

## Checklist
- [x] 🗒 `CHANGELOG` entry


rolandkakonyi (approved):

rolandkakonyi:
Thank you @ibocharnikov-3ss-tv! This is a great improvement!

PR #887, opened 2026-03-25
Android: add `PlayerView.isPictureInPictureEnabled` prop handling

ibocharnikov-3ss-tv:
## Description
There is a need to enable/disable PiP feature after player initialisation, for the same player instance. As far as `PlayerViewConfig.pictureInPictureConfig` can't be changed after player initialisation - new `PlayerView.isPictureInPictureEnabled` property was added to handle such kind of behavior.

## Changes
Added `PlayerView.isPictureInPictureEnabled` property and its handling on iOS side to enable/disable PiP feature in runtime.

## Checklist
- [x] 🗒 `CHANGELOG` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,14 +2,13 @@
 
 ### Added
 
-- iOS: `PlayerView.isPictureInPictureEnabled` prop to enable or disable Picture in Picture dynamically after player initialization
+- Android: `PlayerView.isPictureInPictureEnabled` prop to enable or disable Picture in Picture dynamically after player initialization
 
 ## [1.14.0] - 2026-03-20
 
 ### Added
 
 - `SourceConfig.cmcdConfig.isEnabled` to enable support for Common Media Client Data (CMCD) on iOS/tvOS 18+ (Android currently ignores `cmcdConfig`)
-- iOS: `PlayerView.isPictureInPictureEnabled` prop to enable or disable Picture in Picture dynamically after player initialization
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,25 @@
 # Changelog
 
+## [Unreleased]
+
 ### Added
 
-- Android: `PlayerView.isPictureInPictureEnabled` prop to enable or disable Picture in Picture dynamically after player initialization
+- Android: `PlayerView.isPictureInPictureEnabled` to enable or disable Picture in Picture availability dynamically after player initialization
+
+## [1.15.0] - 2026-03-27
+
+### Added
+
+- Android: `PlaybackConfig.handleAudioFocus` to control whether the player pauses playback when audio focus is lost
+- iOS: `PlayerView.isPictureInPictureEnabled` to enable or disable Picture in Picture availability dynamically after player initialization
+
+### Changed
+
+- Update Bitmovin's native iOS SDK version to `3.110.0`
+
+### Fixed
+
+- iOS, tvOS: Error/Warning code not forwarded to the React Native bridge on error and warning events
 
 ## [1.14.0] - 2026-03-20
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Changed
+
+- `DebugConfig.setDebugLoggingEnabled` now also controls JS-side debug logging; `console.log` calls in the `Network` and `Drm` modules are suppressed unless debug logging is enabled
+
 ### Added
 
 - Android: `PlayerView.isPictureInPictureEnabled` to enable or disable Picture in Picture availability dynamically after player initialization
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Changed
 
 - `DebugConfig.setDebugLoggingEnabled` now also controls JS-side debug logging; `console.log` calls in the `Network` and `Drm` modules are suppressed unless debug logging is enabled
+- Update Bitmovin's native iOS SDK version to `3.111.0`
 
 ## [1.15.0] - 2026-03-27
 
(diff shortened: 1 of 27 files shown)


luckygoyal-bitmovin:
Hi @ibocharnikov-3ss-tv , thanks for proposing this PR. Can you please resolve the conflicts in the PR branch?

ibocharnikov-3ss-tv:
> Hi @ibocharnikov-3ss-tv , thanks for proposing this PR. Can you please resolve the conflicts in the PR branch?

hi @luckygoyal-bitmovin , done, conflicts merged, thanks for pointing out.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -762,6 +760,24 @@ class RNPlayerView(context: Context, appContext: AppContext) : ExpoView(context,
         }
     }
 
+    fun setIsPictureInPictureEnabled(isEnabled: Boolean) {
```
123mpozzi, line 768: This method is not reachable as of now from JS due to the early iOS-only release. We need to modify `src/components/PlayerView/index.tsx:113` removing that additional guard:

```diff
  useEffect(() => {
    if (
      isPlayerInitialized &&
      isPictureInPictureEnabled != null
-      isPictureInPictureEnabled != null &&
-      Platform.OS === 'ios'
    ) {
      void nativeView.current?.setIsPictureInPictureEnabled(
        isPictureInPictureEnabled
      );
    }
  }, [isPlayerInitialized, isPictureInPictureEnabled]);
```

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 - `DebugConfig.setDebugLoggingEnabled` now also controls JS-side debug logging; `console.log` calls in the `Network` and `Drm` modules are suppressed unless debug logging is enabled
 
+### Added
+
+- Android: `PlayerView.isPictureInPictureEnabled` to enable or disable Picture in Picture availability dynamically after player initialization
+
```
123mpozzi, line 8: Nit: we usually put the `Added` section above the `Changed` one. We basically follow this convention: https://keepachangelog.com/en/1.1.0/

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
```diff
@@ -43,8 +43,6 @@ class RNPlayerViewManager : Module() {
                     ?: throw IllegalArgumentException("Player info must contain 'playerId' field")
                 val customMessageHandlerBridgeId = playerInfo.getString("customMessageHandlerBridgeId")
                 val enableBackgroundPlayback = playerInfo.getBooleanOrNull("enableBackgroundPlayback") ?: false
-                val isPictureInPictureEnabledOnPlayer =
-                    playerInfo.getBooleanOrNull("isPictureInPictureEnabledOnPlayer") ?: false
```
123mpozzi, line 47: I don't think we can remove reading `isPictureInPictureEnabledOnPlayer`.

Otherwise, integrations using `playbackConfig.isPictureInPictureEnabled` would break. That property is deprecated, but we should still support it.

Applies to `RNPlayerView.kt` as well, I'd revert those removals and keep the original handling that also includes the deprecated property as a fallback to check whether PiP is enabled

123mpozzi (commented):
Just leaving a few early-review comments since I am no Android dev 🙂

ibocharnikov-3ss-tv:
@123mpozzi , hi!
All your suggestions and comments were fixes/applied. Thanks for review and for findings!

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerViewManager.kt
```diff
@@ -174,7 +178,7 @@ class RNPlayerViewManager : Module() {
         activePlayerViewsToPlayerIds.remove(view)
     }
 
-    private fun updateAutoPictureInPictureRegistration(view: RNPlayerView) {
+    internal fun updateAutoPictureInPictureRegistration(view: RNPlayerView) {
```
testcenter: Please keep this function `private` and call it in the `AsyncFunction("setIsPictureInPictureEnabled")`.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
testcenter, line 1: There is the function `setPictureInPicture` (in `RnPlayerView`) which needs to handle the new logic as well.

the change you did is only affecting the Player-UI; it is still possible to call `setPictureInPicture` (eg via a button) even though it should be disabled.
(for testing, you could add a button in the `BassicPictureInPicture.tsx` file and call `setPictureInPicture` on it. test with the combination of the `Enter/ Exit PiP` button)

testcenter (changes requested):
All in all good work 👍 
Just some minor changes necessary.

ibocharnikov-3ss-tv:
Hello @testcenter! Thanks for review!
Fixes and adjustments based on your comments were applied and pushed.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -725,6 +725,9 @@ class RNPlayerView(context: Context, appContext: AppContext) : ExpoView(context,
             if (it.isPictureInPicture == isPictureInPicture) {
                 return
             }
+            if (isPictureInPicture && !pictureInPictureConfig.isEnabled) {
```
testcenter: Please move the if block outside the `playerView.let` block. so this is checked at first (otherwise the internal state is bogus (`requestedPictureInPictureValue`))

ibocharnikov-3ss-tv: Fixed. Great catch!
Thank you!

testcenter (approved):
good job 👍

PR #889, opened 2026-03-26
Fix build issues when using Xcode 26.4

123mpozzi:
## Description

React Native fails to build iOS when using Xcode 26.4, see: https://github.com/facebook/react-native/issues/55601
To fix this, we need to bump the `fmt` internal dependency from `11.0.2` to `12.1.0`

This PR uses `patch-package` to apply a patch on post-installs script doing the bump
👉  After the script runs, make sure to regenerate the pods via: `yarn example bootstrap`!

## Checklist
- [x] 🗒 `CHANGELOG` entry -  no need, it's a dev-only change


rolandkakonyi (changes requested):
This happens almost every year with an Xcode release.

Instead of using a hand-rolled patch solution please look into [`patch-package`](https://www.npmjs.com/package/patch-package).

We used this in the past in https://github.com/bitmovin/bitmovin-player-react-native/pull/372

123mpozzi:
> This happens almost every year with an Xcode release.
> 
> Instead of using a hand-rolled patch solution please look into [`patch-package`](https://www.npmjs.com/package/patch-package).
> 
> We used this in the past in #372

Good point! Address in 85b33744404069f1918060cd7a863c5bf2672f85

rolandkakonyi (approved):

PR #902, opened 2026-04-02
Fix PiP layouting

testcenter:
## Description
When in PiP mode the video is sometimes cropped. (this happens mostly when resizing the window)

## Changes
- Remove old layouting / measuring logic
- Add helper to re-parent the playerView parent to the root view and manually measure + layout the playerView in PiP layout changes
- Overwrite `shouldUseAndroidLayout` + remove overwritten `requestLayout` function (when the mentioned flag returns true, internally it does the same as we did when overwriting)

## Checklist
- [x] 🗒 `CHANGELOG` entry

### DIFF
--- a/android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
@@ -651,7 +651,7 @@ class RNPlayerView(context: Context, appContext: AppContext) : ExpoView(context,
             val playerParentIndex: Int,
         )
 
-        private inline fun <reified T : View> View.getParent(): T? {
+        private inline fun <reified T : View> View.findParentOfType(): T? {
             var view = this
             do {
                 view = view.parent as? View ?: return null
@@ -675,9 +675,9 @@ class RNPlayerView(context: Context, appContext: AppContext) : ExpoView(context,
         }
 
         fun reparent() {
-            val playerParent = this@RNPlayerView.getParent<ReactViewGroup>() ?: return
+            val playerParent = this@RNPlayerView.findParentOfType<ReactViewGroup>() ?: return
             val playerParentParent = playerParent.parent as? ViewGroup ?: return
-            val reactRoot = playerParent.getParent<ReactRootView>() ?: return
+            val reactRoot = playerParent.findParentOfType<ReactRootView>() ?: return
 
             viewHolder = ViewHolder(
                 reactRoot = reactRoot,
--- a/android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
@@ -170,7 +170,7 @@ class RNPlayerView(context: Context, appContext: AppContext) : ExpoView(context,
             requestLayout()
         }
 
-    private val reparentRestoreHelper = RNPlayerViewReparentRestoreHelper()
+    private val reparentHelper = RNPlayerViewReparentHelper()
 
     init {
         // React Native has a bug that dynamically added views sometimes aren't laid out again properly.
@@ -214,7 +214,7 @@ class RNPlayerView(context: Context, appContext: AppContext) : ExpoView(context,
 
         activityLifecycle?.removeObserver(activityLifecycleObserver)
         viewTreeObserver.takeIf { it.isAlive }?.removeOnGlobalLayoutListener(globalLayoutListener)
-        reparentRestoreHelper.dispose()
+        reparentHelper.dispose()
 
         // cleanup all children views explicitly,
         // so that in case react native does some view caching we are 100% the child views of this view
@@ -439,15 +439,15 @@ class RNPlayerView(context: Context, appContext: AppContext) : ExpoView(context,
                 playerView.enterPictureInPicture()
             }
 
-            if (!reparentRestoreHelper.isActive) {
-                reparentRestoreHelper.reparent()
+            if (!reparentHelper.isActive) {
+                reparentHelper.reparent()
             }
         } else {
             if (playerView.isPictureInPicture) {
                 playerView.exitPictureInPicture()
             }
 
-            reparentRestoreHelper.tryRestore()
+            reparentHelper.tryRestore()
         }
     }
 
@@ -643,7 +643,7 @@ class RNPlayerView(context: Context, appContext: AppContext) : ExpoView(context,
     // During PiP mode, we temporarily move the view higher up the hierarchy to the ReactRoot.
     // This prevents fragmented rendering when the user resizes the PiP window.
     // On PiP close, we re-arrange the view hierarchy to its original state
-    private inner class RNPlayerViewReparentRestoreHelper {
+    private inner class RNPlayerViewReparentHelper {
         private inner class ViewHolder(
             val reactRoot: ReactRootView,
             val playerParentParent: ViewGroup,
--- a/android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
@@ -699,7 +699,14 @@ class RNPlayerView(context: Context, appContext: AppContext) : ExpoView(context,
         fun tryRestore() {
             val viewHolder = viewHolder ?: return
             viewHolder.reactRoot.removeView(viewHolder.playerParent)
-            viewHolder.playerParentParent.addView(viewHolder.playerParent, viewHolder.playerParentIndex)
+            try {
+                viewHolder.playerParentParent.addView(viewHolder.playerParent, viewHolder.playerParentIndex)
+            } catch (_: Exception) {
+                // In case the view hierarchy layout has changed an exception will be thrown while adding the view
+                // This should never happen, but we can not be sure what react-native does under the hood.
+                // As a fallback add the view without index (will be added as last view)
+                viewHolder.playerParentParent.addView(viewHolder.playerParent)
+            }
 
             dispose()
         }
--- a/android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
@@ -170,7 +170,7 @@ class RNPlayerView(context: Context, appContext: AppContext) : ExpoView(context,
             requestLayout()
         }
 
-    private val reparentRestoreHolder = RNPlayerViewReparentRestoreHelper()
+    private val reparentRestoreHelper = RNPlayerViewReparentRestoreHelper()
 
     init {
         // React Native has a bug that dynamically added views sometimes aren't laid out again properly.
@@ -214,7 +214,7 @@ class RNPlayerView(context: Context, appContext: AppContext) : ExpoView(context,
 
         activityLifecycle?.removeObserver(activityLifecycleObserver)
         viewTreeObserver.takeIf { it.isAlive }?.removeOnGlobalLayoutListener(globalLayoutListener)
-        reparentRestoreHolder.dispose()
+        reparentRestoreHelper.dispose()
 
         // cleanup all children views explicitly,
         // so that in case react native does some view caching we are 100% the child views of this view
@@ -439,15 +439,15 @@ class RNPlayerView(context: Context, appContext: AppContext) : ExpoView(context,
                 playerView.enterPictureInPicture()
             }
 
-            if (!reparentRestoreHolder.isActive) {
-                reparentRestoreHolder.reparent()
+            if (!reparentRestoreHelper.isActive) {
+                reparentRestoreHelper.reparent()
             }
         } else {
             if (playerView.isPictureInPicture) {
                 playerView.exitPictureInPicture()
             }
 
-            reparentRestoreHolder.tryRestore()
+            reparentRestoreHelper.tryRestore()
         }
     }
 
--- a/android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
@@ -116,7 +116,7 @@ class RNPlayerView(context: Context, appContext: AppContext) : ExpoView(context,
 
     private var playerInMediaSessionService: Player? = null
 
-    // This replaces the manual overwriting of the requestLayout function
+    // Setting this flag to `true` in order for React Native to re-render our view when [requestLayout] is triggered
     override val shouldUseAndroidLayout: Boolean = true
 
     private val activityLifecycleObserver = object : DefaultLifecycleObserver {
--- a/android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
@@ -1,14 +1,12 @@
 package com.bitmovin.player.reactnative
 
 import android.annotation.SuppressLint
-import android.app.PictureInPictureParams
 import android.content.Context
 import android.content.res.Configuration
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -118,6 +116,9 @@ class RNPlayerView(context: Context, appContext: AppContext) : ExpoView(context,
 
     private var playerInMediaSessionService: Player? = null
 
+    // This replaces the manual overwriting of the requestLayout function
```
zigavehovec: I don't understand this comment fully. Could we more accurately document why this is needed?

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -118,6 +116,9 @@ class RNPlayerView(context: Context, appContext: AppContext) : ExpoView(context,
...
+    override val shouldUseAndroidLayout: Boolean = true
```
zigavehovec, line 120: Do we really need this? I quickly tested without this flag and the pip resizing seems to be working. This flag states that enabling it can missbehave in some cases so it would be preferable to avoid it if not really needed.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -782,19 +639,70 @@ class RNPlayerView(context: Context, appContext: AppContext) : ExpoView(context,
...
-        super.requestLayout()
-        post {
-            measure(
-                MeasureSpec.makeMeasureSpec(width, MeasureSpec.EXACTLY),
-                MeasureSpec.makeMeasureSpec(height, MeasureSpec.EXACTLY),
+    // React Native doesn't properly handle PiP layout transitions.
+    // During PiP mode, we temporarily move the view higher up the hierarchy to the ReactRoot.
+    // This prevents fragmented rendering when the user resizes the PiP window.
+    // On PiP close, we re-arrange the view hierarchy to its original state
+    private inner class RNPlayerViewReparentRestoreHelper {
```
zigavehovec: Can we move this to separate file?

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -782,19 +639,70 @@ class RNPlayerView(context: Context, appContext: AppContext) : ExpoView(context,
...
+        private inner class ViewHolder(
+            var reactRoot: ReactRootView,
```
zigavehovec: Why are these not immutable?

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -782,19 +639,70 @@ class RNPlayerView(context: Context, appContext: AppContext) : ExpoView(context,
...
+            var playerParentParent: ViewGroup,
+            var playerParent: ReactViewGroup,
+            var playerParentIndex: Int,
+        )
+
+        private inline fun <reified T : View> View.getParent(): T? {
```
zigavehovec: This function name shadows the build in getParent. Consider `findParentOfType` or something similar

zigavehovec: I would also move this outside and after RNPlayerViewReparentRestoreHelper class, once everything is moved into a separate file.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -782,19 +639,70 @@ class RNPlayerView(context: Context, appContext: AppContext) : ExpoView(context,
...
+                reactRoot = reactRoot,
+                playerParentParent = playerParentParent,
+                playerParent = playerParent,
+                playerParentIndex = playerParentParent.indexOfChild(playerParent),
+            )
+
+            playerParentParent.removeView(playerParent)
+            reactRoot.addView(playerParent)
+
+            playerParent.viewTreeObserver.addOnGlobalLayoutListener(globalLayoutListener)
```
zigavehovec, line 696: We attach the global layout listener to a playerParent, but inside the callback we are measuring reach root. Is this expected?

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -782,19 +639,70 @@ class RNPlayerView(context: Context, appContext: AppContext) : ExpoView(context,
...
+                MeasureSpec.makeMeasureSpec(reactRoot.width, MeasureSpec.EXACTLY),
+                MeasureSpec.makeMeasureSpec(reactRoot.height, MeasureSpec.EXACTLY),
             )
-            layout(left, top, right, bottom)
+            view.layout(0, 0, view.measuredWidth, view.measuredHeight)
+        }
+
+        fun reparent() {
+            val playerParent = this@RNPlayerView.getParent<ReactViewGroup>() ?: return
+            val playerParentParent = playerParent.parent as? ViewGroup ?: return
```
zigavehovec, line 679: What is expected this to be? What's between the first react view group parent and react root?

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -782,19 +639,70 @@ class RNPlayerView(context: Context, appContext: AppContext) : ExpoView(context,
...
+            playerParentParent.removeView(playerParent)
+            reactRoot.addView(playerParent)
+
+            playerParent.viewTreeObserver.addOnGlobalLayoutListener(globalLayoutListener)
+        }
+
+        fun tryRestore() {
+            val viewHolder = viewHolder ?: return
+            viewHolder.reactRoot.removeView(viewHolder.playerParent)
+            viewHolder.playerParentParent.addView(viewHolder.playerParent, viewHolder.playerParentIndex)
```
zigavehovec: My agent gave this feedback: 
tryRestore re-inserts playerParent at its original index, but the parent’s child count may have changed while in PiP (e.g. siblings removed by React reconciliation). If playerParentIndex is now beyond the child count, addView(view, index) will throw IndexOutOfBoundsException.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
zigavehovec:
```suggestion
    private inner class RNPlayerViewReparentHelper {
```

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -782,19 +639,70 @@ class RNPlayerView(context: Context, appContext: AppContext) : ExpoView(context,
...
-     * in `playerView` due to being previously attached to a different parent.
-     */
-    override fun requestLayout() {
-        super.requestLayout()
-        post {
-            measure(
-                MeasureSpec.makeMeasureSpec(width, MeasureSpec.EXACTLY),
-                MeasureSpec.makeMeasureSpec(height, MeasureSpec.EXACTLY),
+    // React Native doesn't properly handle PiP layout transitions.
+    // During PiP mode, we temporarily move the view higher up the hierarchy to the ReactRoot.
```
zigavehovec, line 643: Do we know why moving the view higher up helps the situation? If yes some more details here would be helpful.

zigavehovec (commented):
The removal of previously needed manual measuring of multiple views is nice!

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
testcenter: Yes, we need it.
The problem in this case is not PiP, but when starting normal playback -> the video is then weirdly stretched.

having this flag set, we don't need to override `requestLayout` and have a custom logic, as it was actually the same logic as in the base class when this flag is set, see [ExpoView](https://github.com/expo/expo/blob/main/packages/expo-modules-core/android/src/main/java/expo/modules/kotlin/views/ExpoView.kt#L49-L55)

Anyway: I will update the comment :+1:

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
testcenter: Could be tricky, as it is an `inner` class, accessing the `RNPlayerView` to call `measure` + `layout`.

Moving it to a separate file could lead to having an inconsistent state, so I would like to avoid it. (especially as we are dealing with custom views + layouting in react-native)

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
testcenter: it was a leftover while debugging.
fixed: https://github.com/bitmovin/bitmovin-player-react-native/pull/902/commits/6611f30cc7ae40b4cda9e5c67e50b816357f19db

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
testcenter: renamed to `findParentOfType`: https://github.com/bitmovin/bitmovin-player-react-native/pull/902/commits/489bcca088e3fbb7e47a6f418ae68786f88940b9

as mentioned in an other comment, this (`RNPlayerViewReparentHelper`) is an `inner` class and I would prefer to keep it here.
so in this case, I would also prefer to keep this function (`findParentOfType`) here, as it is specifically only for this use-case and might never be of use outside of this class.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
testcenter: yes this is expected. it is because the root is the first view layer that gets transformed in pip mode.
measuring and layouting this should cascade down the viewHierarchy.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
testcenter: there could be multiple other layers between, depending on the UI layout of the library consumer.
I am not sure I fully understand the question. :sweat_smile:

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
testcenter: renamed: https://github.com/bitmovin/bitmovin-player-react-native/pull/902/commits/0c0097ff1cf2ef98d6484ee23f0dfaa0b66db820

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
testcenter: I am not sure how this can happen... do you think it makes sense to handle this case? :thinking:

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
testcenter: Actually I don't know why it behaves like this in PiP mode... might be due to some Yoga engine handling or maybe related to Expo in combination with a SurfaceView or something else :shrug:

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
testcenter: updated comment: https://github.com/bitmovin/bitmovin-player-react-native/pull/902/commits/e1d682b88ff5587ee45d7024ef65e54ff34d9340

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
zigavehovec: Could you please add this explanation as a comment?

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
zigavehovec: Ah yes, I was under the impression that we are already in the layers that get added automatically be RN outside of what the application does. All good!

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
zigavehovec: Ok. I'm a bit worried as the `shouldUseAndroidLayout` documents that some issues might occur when layouting. I guess we should spend some time testing the other scenarios in the test app to make sure this doesn't break anything.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
zigavehovec: Ah I see. Ok, let's keep it.

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
testcenter: it should not have any affect on our code, as the logic stays the same:
this is the [current state](https://github.com/bitmovin/bitmovin-player-react-native/blob/development/android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt#L790-L799) on the `development` branch.
and [this](https://github.com/expo/expo/blob/main/packages/expo-modules-core/android/src/main/java/expo/modules/kotlin/views/ExpoView.kt#L41-L55) is the logic in `ExpoView` which we will use in future.
So we are actually removing our own logic in favor of the same logic in ExpoView

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
testcenter: added comment: https://github.com/bitmovin/bitmovin-player-react-native/pull/902/commits/6fd2cadf1217bdb085a38f140b85d9b31989542d

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -782,19 +639,74 @@ class RNPlayerView(context: Context, appContext: AppContext) : ExpoView(context,
...
+                playerParentParent = playerParentParent,
+                playerParent = playerParent,
+                playerParentIndex = playerParentParent.indexOfChild(playerParent),
+            )
+
+            playerParentParent.removeView(playerParent)
+            reactRoot.addView(playerParent)
+
+            // We attach the global layout listener to a playerParent,
+            // but inside the callback we are measuring reach root.
```
zigavehovec:
```suggestion
            // but inside the callback we are measuring each root.
```

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
zigavehovec: Ah ok, thanks for the info!

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
zigavehovec: Not sure.. Maybe we could just guard against this case so that we don't crash. Not sure if this can really happen though.

zigavehovec (approved):

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
testcenter: Added fallback handling: https://github.com/bitmovin/bitmovin-player-react-native/pull/902/commits/85b3348587fe4c2c1f2edd850a96a4db35c24519

PR #908, opened 2026-04-09
Feature/ima dai integration

joshreed104:
## Description
This addition enables the use of Google IMA server-side dynamic ad insertion. The user passes an asset id which the IMA SDK uses to insert midroll ads into a live stream video and loads that new video file with ads into the Bitmovin Player.

An IMAAdsWrapper instance is created along with the Bitmovin player.

RNPlayerView attaches an adContainer to the Player which allows the ads to be played and metadata reported, plus it shows the Learn More link and ad timer.

There are callbacks which detect ad events and report back to Google to log ad visibility. There are also events exposed to allow the user to take actions on ad events, though many of these are currently not implemented.

In React Native, the user passes an assetId, a fallbackUrl, and adTagParams to a new function in player.ts called `loadDaiStream`. This calls a corresponding method in the PlayerModule which sets some properties in the IMAAdsWrapper and uses the IMA functionality of `requestAndLoadStream` to stitch the ads and load the new video.

One quirk is I've had to ad a `onBmpAdContainerReady` event that tells the JS side when the container is ready. This has to be true before `loadDaiStream` can be called to kick off the actual stream request. Right now the user has to utilize this callback to load the video in React Native instead of calling `player.load()` in a regular useEffect. Perhaps there's a better way to do this natively.

It also adds an sample screen to the example app with a Big Buck Bunny sample stream assetId from Google, and the Bitmovin sample video as a fallback. If the Bitmovin video plays, it means there was a problem with the IMA flow.

## Changes
Many changes are noted above. The largest changes are the new files added like `IMAAdsWrapper`, `BitmovinVideoDisplay.swift` on the iOS side which helps the IMA stream communicate with the base Bitmovin player, and the changes to the creation of the player itself to also create the ad container and get things ready for an IMA DAI stream.

TBH it's been a while since I implemented the Android side, and recently I implemented the iOS side with lots of help from Cursor and this POC from the Bitmovin team on an older version of the library: https://github.com/bitmovin/bitmovin-player-react-native/pull/679/changes#diff-b8e2561e633a82b40ce1931673d42270e00645bcdeb76b21d69ff2c77e1dcde0

Documentation from Google on IMA integration found here: https://developers.google.com/ad-manager/dynamic-ad-insertion/sdk/android/get-started?service=full

## Checklist
- [ ] 🗒 `CHANGELOG` entry

### DIFF
--- a/android/src/main/java/com/bitmovin/player/reactnative/IMAAdsWrapper.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/IMAAdsWrapper.kt
@@ -207,7 +207,7 @@ class IMAAdsWrapper(
                 String.format("Event: %s\n", adEvent.type),
             )
             AdEventType.TAPPED -> {
-                // Tap on non-clickthrough portion of ad (e.g. video area) — toggle play/pause like iOS.
+                // Tap on non-clickthrough portion of ad (e.g. video area) — toggle play/pause
                 if (videoPlayer.isPlaying) {
                     videoPlayer.pause()
                     playerCallbacks.forEach { it.onPause() }
--- a/android/src/main/java/com/bitmovin/player/reactnative/IMAAdsWrapper.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/IMAAdsWrapper.kt
@@ -1,0 +1,250 @@
+package com.bitmovin.player.reactnative
+
+import android.util.Log
+import android.view.ViewGroup
+import com.bitmovin.player.api.Player
+import com.bitmovin.player.api.event.PlayerEvent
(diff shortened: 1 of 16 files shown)


strangesource:
Hey! Thanks a lot for the contribution — we appreciate the effort that went into this, it looks substantial.

We’ll take some time to properly review and evaluate it from our side, and get back to you once we’ve had a closer look. Thanks again for putting this together!

strangesource:
After reviewing the approach, we’ve decided to implement this in a different way to better align with the direction of the project. We’ll close this PR for now, but we really appreciate your contribution and the discussion it started.

We'll keep you posted on the progress.

PR #920, opened 2026-04-20
Add `pictureInPictureConfig.shouldExitOnForeground`support.

ibocharnikov-3ss-tv:
## Description
Currently, when a user returns to the app while the player is playing in PiP mode, the PiP window stays active — the user must manually tap "Return to App" in the PiP window to restore inline playback (iOS) or interact with the PiP window to expand it. The request is to add a config option that automatically exits PiP when the app becomes active again.

## Changes
Add `pictureInPictureConfig.shouldExitOnForeground` property support to enable "auto" return to fullscreen player when app becomes active.

## Checklist
- [x] 🗒 `CHANGELOG` entry


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
```
123mpozzi, line 2: Unrelated to CHANGELOG, but we need to add `shouldExitOnForeground` to `pictureInPictureConfig.ts`, so that the feature is configurable. E.g.:

```ts
  /**
   * Defines whether Picture in Picture should exit automatically when the app transitions back to foreground.
   *
   * Does not have any effect when Picture in Picture is disabled.
   *
   * @defaultValue `false`
   * @platform iOS
   */
  shouldExitOnForeground?: boolean;
```

android/src/main/java/com/bitmovin/player/reactnative/Types.kt
```diff
@@ -18,6 +18,7 @@ enum class UserInterfaceType {
 data class PictureInPictureConfig(
     val isEnabled: Boolean = false,
     val shouldEnterOnBackground: Boolean = false,
+    val shouldExitOnForeground: Boolean = false,
```
123mpozzi: The feature is not actually implemented on Android, but just parsed into config - is Android in the scope? If not, I'd just keep it iOS-only for simplicity.

Note: it looks like Android by default exits PiP when coming to foreground already

ios/RCTConvert+BitmovinPlayer.swift
```diff
@@ -1054,9 +1054,15 @@ extension RCTConvert {
...
 
+        let shouldExitPiPOnForeground = json["pictureInPictureConfig"].flatMap { pipJson -> Bool? in
+            guard let dict = pipJson as? [String: Any] else { return nil }
+            return dict["shouldExitOnForeground"] as? Bool
+        } ?? false
+
         return RNPlayerViewConfig(
             uiConfig: json["uiConfig"].flatMap(rnUiConfig),
             pictureInPictureConfig: json["pictureInPictureConfig"].flatMap(pictureInPictureConfig),
+            shouldExitPictureInPictureOnForeground: shouldExitPiPOnForeground,
```
123mpozzi: **RN-only config Part 2** - see [part1](https://github.com/bitmovin/bitmovin-player-react-native/pull/920#discussion_r3113853069) for context

Then, we'd need to only change this `flatMap` argument to `rnPictureInPictureConfig`:
```diff
- pictureInPictureConfig: json["pictureInPictureConfig"].flatMap(pictureInPictureConfig),
+ pictureInPictureConfig: json["pictureInPictureConfig"].flatMap(rnPictureInPictureConfig),
```

```suggestion
```

ios/RCTConvert+BitmovinPlayer.swift
```diff
@@ -1054,9 +1054,15 @@ extension RCTConvert {
             return nil
         }
 
+        let shouldExitPiPOnForeground = json["pictureInPictureConfig"].flatMap { pipJson -> Bool? in
+            guard let dict = pipJson as? [String: Any] else { return nil }
+            return dict["shouldExitOnForeground"] as? Bool
+        } ?? false
```
123mpozzi: **RN-only config Part 1**

Can we instead create a ReactNative-only config wrapper? Since this feature does not exist on iOS player native, I'd follow the same pattern we use for `rnPlayerViewConfig` and `rnUiConfig`:

```swift
    static func rnPictureInPictureConfig(_ json: Any?) -> RNPictureInPictureConfig? {
        guard let json = json as? [String: Any?] else {
            return nil
        }

        return RNPictureInPictureConfig(
            nativeConfig: pictureInPictureConfig(json),
            shouldExitOnForeground: json["shouldExitOnForeground"] as? Bool ?? false
        )
    }
```

With this new structure at the bottom:

```swift
/**
 * React Native specific PictureInPictureConfig.
 */
internal struct RNPictureInPictureConfig {
    /**
     * Native SDK PiP configuration.
     */
    let nativeConfig: PictureInPictureConfig?

    /**
     * Whether PiP should exit automatically when the app comes to the foreground.
     */
    let shouldExitOnForeground: Bool
}
```

ios/RCTConvert+BitmovinPlayer.swift
```diff
@@ -1413,6 +1419,11 @@ internal struct RNPlayerViewConfig {
      */
     let pictureInPictureConfig: PictureInPictureConfig?
 
+    /**
+     * Whether PiP should exit automatically when the app comes to the foreground.
+     */
+    let shouldExitPictureInPictureOnForeground: Bool
```
123mpozzi: **RN-only config Part 3** - see [part1](https://github.com/bitmovin/bitmovin-player-react-native/pull/920#discussion_r3113853069) for context

```suggestion
```

Finally, we just need to change



```diff
- let pictureInPictureConfig: PictureInPictureConfig?
+ let pictureInPictureConfig: RNPictureInPictureConfig?
```


And ensure `playerViewConfig` uses the native one:

```diff
    var playerViewConfig: PlayerViewConfig {
        let config = PlayerViewConfig()
-        if let pictureInPictureConfig {
-            config.pictureInPictureConfig = pictureInPictureConfig
+        if let nativePictureInPictureConfig = pictureInPictureConfig?.nativeConfig {
+            config.pictureInPictureConfig = nativePictureInPictureConfig
        }
```

123mpozzi (commented):
Looks good, thanks for your contributions! I manually tested on iOS, and works fine 💪

Just left some comments 🙂 , mostly on encapsulating the serialization into a ReactNative-only wrapper

ibocharnikov-3ss-tv:
Hello @123mpozzi !
Thanks a lot for the review and findings!

All the fixes and adjustments you've been suggested are in place now

ios/RNPlayerView.swift
```diff
@@ -44,6 +44,8 @@ public class RNPlayerView: ExpoView {
     private var requestedPictureInPictureValue: Bool?
     private var isPictureInPictureEnabledValue: Bool?
     private var avPlayerViewControllerTransitionForced = false
+    private var shouldExitPictureInPictureOnForeground = false
+    private var foregroundObserver: NSObjectProtocol?
```
123mpozzi, line 48: We never remove the block-based notification observer on view deallocation - can we add that? Ideally by placing something like this after the `init` block?

```swift
    deinit {
        if let foregroundObserver {
            NotificationCenter.default.removeObserver(foregroundObserver)
            self.foregroundObserver = nil
        }
    }
```

123mpozzi (commented):
Thanks for addressing all my previous comments, the feature looks good overall! I just left one final comment.

We're also discussing it internally, since the process for React Native-only additions is not fully defined yet. I'll keep you posted, but from my side this looks like a solid and valuable contribution.

123mpozzi (approved):

PR #921, opened 2026-04-20
Fix android build failing due to wrong kotlin requirements

123mpozzi:
## Description

This PR fixes possible build issues on Android example/integration apps.
Error was due to RN declared kotlin version being incompatible with the required one in the underlying Android Play SDK.

## Changes

Raise the Android module’s fallback Kotlin version from `1.9.21` to `2.2.20`, which is the latest version supported by the current Expo toolchain. This avoids build failures in environments where Android dependencies/tooling are already on Kotlin `2.2.x`, while staying within Expo’s supported KSP mapping.

Furthermore, it also fixes possible cache invalidation misses. In fact, CI didn't fail the [build](https://github.com/bitmovin/bitmovin-player-react-native/actions/runs/24667480214/job/72128812521) because it restored cache.
By investigating a little, I found out we're trying to cache the wrong app config file: `app.json` instead of `app.config.ts/js`; so changes to apps config were not visible to cache invalidation mechanism (and rebuild was skipped).

## Checklist
- [x] 🗒 `CHANGELOG` entry - just added the kotlin version bump

### DIFF
--- a/example/app.config.ts
+++ b/example/app.config.ts
@@ -79,7 +79,6 @@ const config: ExpoConfig = {
       'expo-build-properties',
       {
         android: {
-          kotlinVersion: '2.2.20',
           buildToolsVersion: '35.0.0',
         },
         ios: {
--- a/example/app.config.ts
+++ b/example/app.config.ts
@@ -79,6 +79,7 @@ const config: ExpoConfig = {
       'expo-build-properties',
       {
         android: {
+          kotlinVersion: '2.2.21',
           buildToolsVersion: '35.0.0',
         },
         ios: {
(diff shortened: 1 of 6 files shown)


example/app.config.ts
rolandkakonyi, line 1: Is this necessary for any integrating app?
If yes, then we need to update our Getting Started Guide as well.

Please make sure the Android team has feedback on this!

strangesource: I think we might be able to avoid this by just using the latest supported version `2.2.20`. Metadata should be compatible.

123mpozzi: Just confirmed via manual testing, it looks not necessary - removed in e9e892d7f6b6b664e066d7c12062f2bfe410983e

strangesource (approved):
Nice!

PR #923, opened 2026-04-22
Fix android video black screen when entering Picture in Picture

ibocharnikov-3ss-tv:
## Description
When switching from fullscreen to Picture-in-Picture mode on Android with the old React Native architecture (Bridge-based), the video picture disappears (black screen) while audio continues playing. The player then shows a loading/stall state until the decoder reconnects to the new surface. The issue does not reproduce with the new RN architecture.

## Changes
Remove `removeView`/`addView` from `reparent()`. Keep the `globalLayoutListener` attachment so `RNPlayerView` is still measured at `reactRoot` dimensions during PiP. Simplify `tryRestore()` to just call `dispose()` since no view was moved.

## Checklist
- [x] 🗒 `CHANGELOG` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Android: Video black screen when entering Picture in Picture mode on old React Native architecture
+
 ### Changed
 
 - Update Bitmovin's native Android SDK version to `3.150.0+jason`
--- a/android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
+++ b/android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
@@ -639,10 +639,12 @@ class RNPlayerView(context: Context, appContext: AppContext) : ExpoView(context,
         }
     }
 
-    // React Native doesn't properly handle PiP layout transitions.
-    // During PiP mode, we temporarily move the view higher up the hierarchy to the ReactRoot.
-    // This prevents fragmented rendering when the user resizes the PiP window.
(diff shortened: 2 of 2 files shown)


testcenter:
@ibocharnikov-3ss-tv could you please provide a sample to reproduce?
I gave it a try but could not reproduce the behavior you mentioned.
I tested with a physical Pixel 6a running Android 16, and an emulator running android 14.

To reproduce, I used our full screen sample and extended with the pip feature. Also I am using a DRM protected stream.
I set the `newArchEnabled` flag in the expo config to  `false` to use the old architecture.

for details see attached patch file.

[reproduction-steps.patch](https://github.com/user-attachments/files/27635765/reproduction-steps.patch)

ibocharnikov-3ss-tv:
hi @testcenter !

We don't have any specific sample app, we used generic ones with old architecture enabled.
I assume it must be enough to use old architecture, afair.

Also it was reproducible not on all the devices, we reproduced it on Samsung Galaxy A13 (Android 14) - when triggering fullscreen player to PiP player, and vice-versa only on tablet devices: Xiaomi Pad 6 (Android 14), Samsung Galaxy Tab A9 (Android 16)

testcenter:
@ibocharnikov-3ss-tv Thanks for the informations.
can you please try to reproduce the behavior with your devices and the patch file I attached in my last comment? (you should delete and recreate the `example/android` directory via `yarn bootstrap` in order to enable the old architecture)
btw: I am out for 2.5 weeks. I don't know who will take over this ticket (if at all)

testcenter:
closing this, as there is no input since weeks.
in case this is still important, re-open the PR

PR #927, opened 2026-04-24
Android: Add PiP transition ended events

testcenter:
## Changes
- Add support for `PictureInPictureEntered` and `PictureInPictureExited` Events on Android
- Reorder call to `playerView.onPictureInPictureModeChanged` to correctly trigger `onPictureInPictureExited` event.

## Checklist
- [x] 🗒 `CHANGELOG` entry


android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
```diff
@@ -449,6 +449,12 @@ class RNPlayerView(context: Context, appContext: AppContext) : ExpoView(context,
 
             reparentHelper.tryRestore()
         }
+
+        // We must "delay" the onPictureInPictureModeChanged callback via posting a runnable.
```
zigavehovec, line 453: Do you know why do we need to delay here? I don't think we need to do this in the native implementations.

zigavehovec (commented):
Only a clarification from my side. Looks good otherwise!

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
testcenter: Not 100% sure, but it could be for 
1. the layout transactions we are doing (re-attaching the PiP to its parent)
or
2. we are handling the PiP changed callback different in react native to our native implementation
  - native: the client must call `PlayerView.onPictureInPictureModeChanged` from the `Activity.onPictureInPictureChanged` callback
  - react-native: we override `View.onConfigurationChanged` to call `PlayerView.onPictureInPictureModeChanged`  (since we don't have an activity with the expo setup, this is a workaround)

ibocharnikov-3ss-tv:
Hello @testcenter !
I tested example app from current PR branch, and it seems that pip enter/exit event on android have not "equal" behaviour to ios and to what we expected.

This is **android app logs** from the flow: start app, start playback, press android HOME button, press "fullscreen" button in the PiP overlay.
As you can see, app state is equal in all the enter(ed) and exit(ed) events.

Android PIP IN- OUT

The expectation was to have "exit" event flow, in case when app is expanding from pip to fullscreen mode, to be like:
```
app state -> background
pip exit -> active
pip exited -> active
app state -> active
```

testcenter:
Hey @ibocharnikov-3ss-tv 
Can you reproduce this behavior with the example app? (`BasicPictureInPicture.tsx`)
If so, can you please clear the caches, remove the example app and retry? (also with `yarn bootstrap` to recreate the example app)

I added some logs in the `BasicPictureInPicture.tsx` and got these states (pip exited was received after the app is active again):

```log
04-28 13:45:11.856 22123 22282 I ReactNativeJS: RNTEST - appStateChange background
04-28 13:45:12.448 22123 22282 I ReactNativeJS: RNTEST - pip enter
04-28 13:45:12.485 22123 22282 I ReactNativeJS: RNTEST - pip entered
04-28 13:45:16.735 22123 22282 I ReactNativeJS: RNTEST - pip exit
04-28 13:45:16.760 22123 22282 I ReactNativeJS: RNTEST - appStateChange active
04-28 13:45:16.786 22123 22282 I ReactNativeJS: RNTEST - pip exited
```

ibocharnikov-3ss-tv:
Hi @testcenter !

I'm using current PR branch with its latest changes.
I removed installed example app from my Samsung Galaxy A23 device, and installed example app from the very beginning: `yarn bootstrap && yarn example android`.

To the existing `BasicPictureInPicture.tsx` I've added only single line of log at line #73: `prettyPrint('appStateChange', nextState);`. In result:

I'm getting app crashed as soon as I pressed native HOME button, and app moved to PiP view:
`
Your app just crashed. See the error below.
java.lang.IllegalArgumentException: Cannot coerce value to an empty range: maximum -14.75 is less than minimum 0.0.
  kotlin.ranges.RangesKt___RangesKt.coerceIn(_Ranges.kt:1476)
  expo.modules.devmenu.fab.FabUtilsKt.coerceIn-ULxng0E(FabUtils.kt:49)
  expo.modules.devmenu.fab.FabUtilsKt.coerceIn-ULxng0E$default(FabUtils.kt:46)
  expo.modules.devmenu.fab.FabUtilsKt.calculateTargetPosition-IEwrmTk(FabUtils.kt:35)
  expo.modules.devmenu.fab.MovableFloatingActionButtonKt$MovableFloatingActionButton$3$1$1.invokeSuspend(MovableFloatingActionButton.kt:80)
  kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:34)
  kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:101)
  androidx.compose.ui.platform.AndroidUiDispatcher.performTrampolineDispatch(AndroidUiDispatcher.android.kt:79)
  androidx.compose.ui.platform.AndroidUiDispatcher.access$performTrampolineDispatch(AndroidUiDispatcher.android.kt:41)
  androidx.compose.ui.platform.AndroidUiDispatcher$dispatchCallback$1.run(AndroidUiDispatcher.android.kt:57)
  android.os.Handler.handleCallback(Handler.java:958)
  android.os.Handler.dispatchMessage(Handler.java:99)
  android.os.Looper.loopOnce(Looper.java:230)
  android.os.Looper.loop(Looper.java:319)
  android.app.ActivityThread.main(ActivityThread.java:8934)
  java.lang.reflect.Method.invoke(Native Method)
  com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:578)
  com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1103)
`.

I can't fully test the process because of the crash... sorry...

ibocharnikov-3ss-tv:
@testcenter , I managed to fix the crash by patching expo-dev-menu package.

My JS logs are:
Screenshot 2026-04-28 at 15 30 12

testcenter:
Good morning @ibocharnikov-3ss-tv 
about that exception, I have never seen this before :grimacing: ... Anyway, it seems like this is not a problem anymore, right?

hm... thats weird. I double checked (cleaned caches, removed the example and let it recreate).
The only change I have locally is the log for appState (`BasicPictureInPicture.tsx` line `73`). I have different state logs than you (see screenshot below)

did you test with this branch and no modifications on it?

image

ibocharnikov-3ss-tv:
Good morning @testcenter !

Yes, crash is not the problem anymore - I found a workaround.

Yes, I'm testing on this branch, on any other branch I won't have "entered" or "exited" events triggered on Android, so I'm sure I'm using correct branch.
In what way do you moving the app into PiP mode: in app button in the header or native android "Home" button? I'm using native android "Home" button... Maybe this is the diff?

testcenter:
@ibocharnikov-3ss-tv 
the native events should be available already on the `development` branch, but not yet on react-native. :+1: 

I just gave it a try, and for all 3 methods, it produced the same state sequence with being `Exited` event being the very last, after the app state changed.
(native home button, "ENTER PIP" button, and the player pip mode button)

fyi: I am using a Pixel 6a device (real device) with Android 16 as a test device. (also tested with an emulator: Pixel 9 Pro, android 16)

ibocharnikov-3ss-tv:
Hi @testcenter .
Unfortunately, on my side, with my Samsung Galaxy A23 Android 14 I still see the same behaviour mentioned [here](https://github.com/bitmovin/bitmovin-player-react-native/pull/927#issuecomment-4335236038)

testcenter:
Hi @ibocharnikov-3ss-tv !
I did some tests, and noticed that the behavior changed with the change of Android 14 -> Android 15.
I will keep you updated about this :+1:

android/src/main/java/com/bitmovin/player/reactnative/RNPlayerView.kt
zigavehovec: Could you improve the comment a bit please? We state what we have to do but the why is not very clear. Just stating that the enter/exit event would be emitted too fast adds some value IMO.

testcenter: Update the comment: https://github.com/bitmovin/bitmovin-player-react-native/pull/927/commits/97a69bba31b47b6a9528374f75b21051b1fb11a5

zigavehovec (approved):

testcenter:
@ibocharnikov-3ss-tv fyi:

I will merge this PR as is with the known limitation of the Android <15 behavior.
We have a follow up work-item for devices running on Android <15. (though I don't know for when this is scheduled)

testcenter:
fyi: the known limitation got fixed and released with [1.20.1](https://github.com/bitmovin/bitmovin-player-react-native/releases/tag/v1.20.1)

PR #933, opened 2026-05-04
Expose Android shouldPlayAdBreak callback

strangesource:
## Description
This change exposes the Android `AdvertisingConfig.shouldPlayAdBreak` callback through the React Native bridge.

Previously, React Native integrations could filter ad items before loading (`shouldLoadAdItem`), but could not decide at ad-break playback time whether a scheduled break should still play. This prevented runtime discard logic for already-scheduled breaks (for example when playback starts from an offset).

## Changes
- Added `AdvertisingConfig.shouldPlayAdBreak?: (adBreak: AdBreak) => boolean` in `src/advertising.ts` (Android-only API surface).
- Extended the JS/native module contract in `src/modules/PlayerModule.ts`:
  - new event: `onShouldPlayAdBreak`
  - new bridge method: `setShouldPlayAdBreak(id, shouldPlay)`
- Updated `src/player.ts`:
  - added Android-only listener wiring for `onShouldPlayAdBreak`
  - clones callback payload before invoking user callback
  - defaults to `true` on callback errors/non-boolean return, matching existing callback behavior
  - ensures listener lifecycle cleanup on player destroy
- Updated Android native bridge in `android/src/main/java/com/bitmovin/player/reactnative/PlayerModule.kt`:
  - added `ResultWaiter<Boolean>` for should-play responses
  - added `onShouldPlayAdBreak` event registration
  - added `setShouldPlayAdBreak` async function
  - wired `setupShouldPlayAdBreak(...)` into player initialization and mapped callback to `AdvertisingConfig.shouldPlayAdBreak`
- Added Android integration test coverage in `integration_test/tests/advertisingTest.ts`:
  - verifies preroll ad break is skipped when `shouldPlayAdBreak` returns `false`
  - verifies callback invocation and non-ad playback state
- Added changelog entry under `## [Unreleased]`.

### API validation
- Verified Android API availability/signature against the official Bitmovin Android API docs (`ShouldPlayAdBreakCallback.shouldPlayAdBreak(adBreak: AdBreak): Boolean` and `AdvertisingConfig.shouldPlayAdBreak`).

### Testing
Executed in this environment:
- `corepack yarn typecheck` ✅
- `corepack yarn lint` ✅
- `corepack yarn lint:android` ✅
- `corepack yarn build` ✅
- `corepack yarn integration-test typecheck` ✅
- `corepack yarn --cwd integration_test eslint tests/advertisingTest.ts --ext .ts --quiet` ✅

Android instrumentation attempts (`yarn integration-test test:android`):
-  ✅ suite executed end-to-end on emulator
  - Result: `50 examples, 7 failures` (existing unrelated failures)
  - **New test passed**: `filtering ad breaks via shouldPlayAdBreak: skips preroll ad break on Android when callback returns false`

## Checklist
- [x] 🗒 `CHANGELOG` entry

### DIFF
--- a/integration_test/tests/advertisingTest.ts
+++ b/integration_test/tests/advertisingTest.ts
@@ -191,6 +191,49 @@ export default (spec: TestScope) => {
         });
       }
     );
+
+    spec.describe(
+      `filtering scheduled ${name} ad breaks via shouldPlayAdBreak`,
+      () => {
+        spec.it('skips ad break when callback returns false', async () => {
+          if (currentPlatform !== 'android') {
+            return;
+          }
+
+          let callbackInvocationCount = 0;
+          const advertisingConfig: AdvertisingConfig = {
+            schedule: adItems,
+            shouldPlayAdBreak: () => {
+              callbackInvocationCount += 1;
+              return false;
+            },
+          };
+          const playbackConfig: PlaybackConfig = {
+            isAutoplayEnabled: true,
+          };
+
+          await startPlayerTest(
+            { playbackConfig, advertisingConfig },
+            async () => {
+              await callPlayerAndExpectEvent(
+                (player) => {
+                  player.load(Sources.artOfMotionHls);
+                },
+                FilteredEvent<TimeChangedEvent>(
+                  EventType.TimeChanged,
+                  (event) => event.currentTime > 0.25
+                )
+              );
+
+              expect(callbackInvocationCount).toBeGreaterThan(0);
+              await callPlayer(async (player) => {
+                expect(await player.isAd()).toBeFalsy();
+              });
+            }
+          );
+        });
+      }
+    );
   }
 
   const adScenarios: AdScenario[] = [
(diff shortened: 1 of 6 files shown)


integration_test/tests/advertisingTest.ts
```diff
@@ -282,6 +282,54 @@ export default (spec: TestScope) => {
     .filter(({ enabledOn }) => isEnabledForCurrentPlatform(enabledOn))
     .forEach(commonAdvertisingTests);
 
+  spec.describe('filtering ad breaks via shouldPlayAdBreak', () => {
```
123mpozzi: Can we move it inside `commonAdvertisingTests`, so e.g.  after the `filtering scheduled ${name} ads via shouldLoadAdItem` block, so that it automatically covers BAM as well?


Then the only necessary changes would be:


```diff

-     spec.it(
-     'skips preroll ad break on Android when callback returns false',
+    spec.describe(
+      `filtering scheduled ${name} ad breaks via shouldPlayAdBreak`,
+      () => {
+        spec.it('skips ad break when callback returns false', async () => {
          if (currentPlatform !== 'android') {
            return;
          }

        let callbackInvocationCount = 0;
        const advertisingConfig: AdvertisingConfig = {
-          schedule: [
-            {
-              sources: [{ tag: AdTags.vastSkippable, type: AdSourceType.IMA }],
-              position: 'pre',
-            },
-          ],
+         schedule: adItems,
          shouldPlayAdBreak: () => {

```

src/advertising.ts
```diff
@@ -113,6 +113,21 @@ export interface AdvertisingConfig {
    * - `false` → the ad item will be skipped and removed from the playback schedule.
    */
   shouldLoadAdItem?: (adItem: AdItem) => boolean;
+  /**
+   * Called right before a scheduled ad break starts playback.
+   *
+   * Use this callback to conditionally allow or skip ad breaks at runtime
+   * (e.g., when starting playback from a time offset and discarding past breaks).
+   *
```
123mpozzi, line 121: Nit: maybe we should mention there's an internal timeout, after which the default behavior is applied (ad plays). This can be interesting to know if the integration callback is doing heavy operations

Internally, the timeout is defined via `shouldPlayAdBreakWaiter.make(250)`

```suggestion
   *
   * @remarks
   * If the callback does not return within the platform's internal timeout, the ad break will play.
   * The callback must be synchronous.
   *
```

If we proceed, I'll also do the same for `shouldLoadAdItem` in a separate PR

123mpozzi (approved):
Looks good and manual testing also passes! Just left a few comments

integration_test/tests/advertisingTest.ts
strangesource: Great suggestion, thanks! ✅ Moved the `shouldPlayAdBreak` test into `commonAdvertisingTests` right after the `shouldLoadAdItem` block, and switched it to scenario-driven setup (`schedule: adItems`) with the generic test wording (`skips ad break when callback returns false`). This now runs across the ad scenarios (including BAM/BITMOVIN).

src/advertising.ts
strangesource: Great point, thank you! ✅ Added `@remarks` to `AdvertisingConfig.shouldPlayAdBreak` documenting the internal timeout fallback (ad break plays if callback does not return in time) and that the callback must be synchronous.

testcenter (approved):

PR #965, opened 2026-06-15
Fix Android build could miss necessary desugaring plugin

123mpozzi:
## Description

Android app builds can fail when a dependency requires Java library desugaring because the config plugin only enabled desugaring while adding optional Cast or offline dependencies.

## Changes

- Enable core library desugaring independently from optional feature dependencies
- Update existing disabled desugaring settings instead of duplicating Gradle entries
- Add plugin regression coverage and run it in TypeScript CI

## Manual testing

Before the fix:

- Check out `development`
- In the example app, configure Android without `features.offline` and without `features.googleCastSDK.android`
- Build the android app, it should fail because Java library desugaring is not enabled while a dependency requires it

After the fix:

- Check out this branch
- In the example app, configure Android without `features.offline` and without `features.googleCastSDK.android`
- Build the android app, it should work

Additional check:

- Run `yarn test:plugin`

## Checklist
- [x] 🗒 `CHANGELOG` entry

### DIFF
--- a/plugin/src/withAppGradleDependencies.ts
+++ b/plugin/src/withAppGradleDependencies.ts
@@ -18,6 +18,8 @@ const defaultProps: PluginProps = {
 
 const CORE_LIBRARY_DESUGARING_DEPENDENCY =
   "coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:2.1.5'";
+const CORE_LIBRARY_DESUGARING_COORDINATES =
+  'com.android.tools:desugar_jdk_libs:2.1.5';
 
 const stripBlockComments = (contents: string) =>
   contents.replace(/\/\*[\s\S]*?\*\//g, '');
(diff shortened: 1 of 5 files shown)


plugin/src/withAppGradleDependencies.ts
```diff
@@ -92,22 +191,43 @@ const withAppGradleDependencies: ConfigPlugin<PluginProps> = (
       );
       return config;
     }
-    const position = dependenciesBlockStart + dependenciesBlockEnd;
-    let insertedDependencies: string[] = [];
-    insertedDependencies.push('\n');
-    insertedDependencies.push(
-      `${combinedProps.spacing}coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:2.1.5'`
+    const androidPosition = androidBlockStart + androidBlockEnd;
+    config.modResults.contents = ensureCoreLibraryDesugaringCompileOptions(
```
123mpozzi, line 193: The key behavior change here is that core library desugaring is treated as required Gradle setup, independent of the optional Cast/offline dependencies. Previously, an empty feature dependency list could skip this setup entirely; this call keeps the desugaring compileOptions path unconditional while the dependency list below still only adds optional feature dependencies when requested.

123mpozzi (commented):
Added implementation context for the desugaring setup change.

plugin/src/withAppGradleDependencies.ts
```diff
@@ -9,16 +9,140 @@ export type PluginProps = {
...
 };
 
+const CORE_LIBRARY_DESUGARING_DEPENDENCY =
+  "coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:2.1.5'";
+const CORE_LIBRARY_DESUGARING_ARTIFACT = 'com.android.tools:desugar_jdk_libs';
+
+const hasCoreLibraryDesugaringEnabled = (contents: string) =>
+  /^\s*(?:setCoreLibraryDesugaringEnabled\s*\(\s*true\s*\)|coreLibraryDesugaringEnabled(?:\s*=\s*|\s+)true\b)/m.test(
+    contents
+  );
```
krocard:
```suggestion
const hasCoreLibraryDesugaringEnabled = (contents: string) =>
  /^\s*(?:set)?CoreLibraryDesugaringEnabled\b.*?\btrue\b/m.test(contents);
```

krocard:
@codex review

krocard (approved):
This was a surface review only. Let's get this merged to unblock the client and it will be removed once IMA becomes optional.
+1 for adding tests.
I feel like this code would be much simpler if it was added at the gradle level.

PR #971, opened 2026-06-19
Update iOS player to 3.115.0

bitPlayerGHActions:
Automated iOS player version update to 3.115.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Removed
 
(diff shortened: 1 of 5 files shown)


123mpozzi (changes requested):
We need to update the now playing API wiring since we removed a deprecated API - I'll take care of that.

Update: this PR has to merge after #973

123mpozzi (approved):

PR #988, opened 2026-06-26
Split caption metadata integration tests

123mpozzi:
## Description

Splits cue metadata assertions out of the general caption integration test file.

## Changes

No production logic change. It is only integration-test restructuring:
- Move Android and iOS cue metadata coverage to a dedicated caption metadata test file
- Keep `captionTest.ts` focused on caption workflow and basic CueEnter/CueExit event shape
- Use `toBeCloseTo` for numeric cue layout assertions


integration_test/tests/captionTest.ts
```diff
@@ -14,47 +14,7 @@ import { expect } from './helper/Expect';
 import {
   CueEnterEvent,
```
123mpozzi: ` captionTest.ts` becomes the behavior/workflow file:
  - captions emit enter/exit events
  - available subtitle tracks have expected basic properties
  - `CueEnter` / `CueExit` have the basic event shape
  - selecting/disabling subtitle tracks works

integration_test/tests/captionMetadataTest.ts
```diff
@@ -0,0 +1,235 @@
+import { TestScope } from 'cavy';
+import {
```
123mpozzi, line 2: `  captionMetadataTest.ts` becomes the payload-detail file:
  - cue layout
  - cue html
  - platform differences in cue metadata shape

integration_test/tests/index.ts
```diff
@@ -12,6 +13,7 @@ import videoQualityTest from './videoQualityTest';
 export default [
   AdvertisingTest,
   CaptionTest,
```
jarhoax, line 15: nit: I guess `CaptionTest` (and others) should also be lower camel case for consistency, but not a blocker.

jarhoax (approved):

integration_test/tests/index.ts
123mpozzi: Good point. I’ll leave this unchanged for this PR since `captionMetadataTest` already follows lower camel case and renaming the older existing imports would create unrelated churn.

We can clean up the remaining test import casing separately if we want

PR #1028, opened 2026-07-29
Simplify Android Gradle caching

123mpozzi:
## Description

Simplifies Android CI caching by using Gradle's supported cache integration and generating clean Android projects with Expo before each build.

## Changes

- Remove manual Gradle `actions/cache` steps from Android build jobs
- Remove Expo prebuild caches for generated native projects
- Generate only the required Android projects with `expo prebuild --clean --platform android`
- Remove unused build-cache artifact uploads
- Let `gradle/actions/setup-gradle` own Gradle cache handling
- Update `gradle/actions/setup-gradle` from v4 to v6

## Validation

- Verified the example, integration-test, and Android TV prebuild commands route to Expo CLI
- Checked the workflow with Prettier and YAML assertions

### DIFF
--- a/.github/workflows/ci-android.yml
+++ b/.github/workflows/ci-android.yml
@@ -81,16 +81,6 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v7
 
-      - name: Cache Android Prebuild
-        id: android-prebuild-cache
-        uses: actions/cache@v6
-        with:
-          path: |
-            example/android
-          key: ${{ runner.os }}-android-prebuild-${{ hashFiles('package.json', 'yarn.lock', 'expo-module.config.json', 'example/package.json', 'example/yarn.lock', 'example/app.config.*', 'plugin/**', 'android/**') }}
-          restore-keys: |
-            ${{ runner.os }}-android-prebuild-
-
       - name: Setup Environment
         uses: ./.github/actions/setup-environment
         with:
@@ -99,9 +89,8 @@ jobs:
           subprojects: true
           skip-prebuild: true
 
-      - name: Prebuild Android Example (on cache miss)
-        if: ${{ steps.android-prebuild-cache.outputs.cache-hit != 'true' }}
-        run: .github/scripts/smart-prebuild.sh example
+      - name: Prebuild Android Example
+        run: yarn example prebuild --clean --platform android
 
       - name: Build Android example
         id: build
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-android.yml
```diff
@@ -91,20 +91,6 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-android-prebuild-
 
-      - name: Setup Gradle Build Cache
```
123mpozzi, line 94: Seems like `gradle/action/setup-gradle` also has buildin caching for gradle , so these steps are [unnecessary](https://github.com/gradle/actions/blob/main/docs/setup-gradle.md)

> More sophisticated and more efficient caching of Gradle User Home between invocations, compared to setup-java and most custom configurations using actions/cache

.github/workflows/ci-android.yml
```diff
@@ -87,24 +87,10 @@ jobs:
         with:
           path: |
             example/android
-          key: ${{ runner.os }}-android-prebuild-${{ hashFiles('example/yarn.lock', 'example/app.config.*', 'plugin/**', 'android/**') }}
+          key: ${{ runner.os }}-android-prebuild-${{ hashFiles('package.json', 'yarn.lock', 'expo-module.config.json', 'example/package.json', 'example/yarn.lock', 'example/app.config.*', 'plugin/**', 'android/**') }}
```
matamegger: Is this hasing the full android/ subtree i.e. all the source code for android? Is this working as intended? (I know the `android/**` was already there before)

matamegger (approved):

.github/workflows/ci-android.yml
123mpozzi: I pushed an update based on your comment. `android/**` was indeed hashing the full native Android subtree, so I removed the Expo prebuild caches and now run a clean Android-only prebuild in each job.

Gradle caching is handled by `gradle/actions/setup-gradle`, and I also removed unused artifact uploads that had no consumers. Could you please take another look?

matamegger: If removing everything besides the gradle cache is working for this setup, this is great.

I am not very familiar with what else might be worth caching in these builds.

matamegger (approved):

.github/workflows/ci-android.yml
123mpozzi: Yeah, apparently we were uploading artifacts that were never consumed (I think phase 2 was supposed to download and reuse them but that was left out of development).

I think using `setup-gradle` is a good starting point (builds looks stable around `~4` minutes, which is similar to pre-fix, maybe a little quicker), but of course we can always look into more optimization in the future 🚀

PR #1029, opened 2026-07-31
Support selective running of integration tests

123mpozzi:
## Description

Integration test runs currently execute the complete Cavy suite. This adds command-line suite selection so contributors can run only the coverage relevant to their changes.

## Changes

- Add `--tags` filtering for Android, iOS, and combined integration test runs
- Support selecting one or more suites as a union
- Define selector names and platform support alongside suite registration
- Validate unknown and platform-incompatible selectors before starting tests
- Add `yarn integration-test list-tags` for selector discovery
- Document how contributors register new test suites
- Add runner, launcher, registry, and tagging checks to TypeScript CI

## Validation

- `yarn integration-test check:runner`
- Manual iOS single-selector and multi-selector runs. E.g.: `yarn integration-test test:ios --tags playback,unloading`
  - list tags via: `yarn integration-test list-tags`


integration_test/scripts/start-test-ios.sh
```diff
@@ -1,15 +1,13 @@
...
-ARGS=$@
-
 # Get the first available iPhone simulator name
 SIMULATOR_NAME=$(xcrun simctl list devices available -e -j | \
     jq --raw-output -s '[.[].devices.[].[] | select( .deviceTypeIdentifier | contains("iPhone") )|.name] | first')
 
 if [ -n "$SIMULATOR_NAME" ] && [ "$SIMULATOR_NAME" != "null" ]; then
     echo "Running tests on iOS simulator: $SIMULATOR_NAME"
-    yarn cavy run-ios --no-screenshots --keep-alive-timeout=300 --terminal="terminal" --simulator "$SIMULATOR_NAME" $ARGS
+    yarn cavy run-ios --no-screenshots --keep-alive-timeout=300 --terminal="terminal" --simulator "$SIMULATOR_NAME" "$@"
```
123mpozzi, line 10: This handles whitespaces better, e.g.: `./scripts/start-test-android.sh --file "specs/caption metadata.js"`

Old version args:
  --file
  specs/caption
  metadata.js ⚠️ 

New version args:
  --file
  specs/caption metadata.js

integration_test/tests/index.js
rolandkakonyi, line 1: Why is this JavaScript now? We already had an `index.ts`

integration_test/types/cavy.d.ts
```diff
@@ -23,6 +23,7 @@ declare module 'cavy' {
     waitTime?: number;
     startDelay?: number;
     clearAsyncStorage?: boolean;
+    only?: string[];
```
rolandkakonyi, line 26: Is this a supported parameter for Cavy that was not exposed to TS before?

123mpozzi: Yes. Cavy already supports Tester’s `only` prop and uses it to filter tagged tests. Our local handwritten Cavy declaration did not expose that existing runtime parameter

integration_test/tests/index.js
123mpozzi: The registry is now consumed by both Metro and the plain Node test runner. Keeping it as CommonJS avoids adding a TypeScript runtime or duplicating the registry, while the lazy `require()` callbacks prevent Node from loading the React Native test modules.

`index.d.ts` provides the TypeScript surface

rolandkakonyi (approved):

PR #1036, opened 2026-08-05
Prevent iOS startup crash without Google Cast

rolandkakonyi:
## Description

Apps that do not configure Google Cast can crash during iOS startup because `AppLifecycleDelegate` invokes `BitmovinCastManager` even though the `google-cast-sdk` pod is optional.

## Changes

- Guard Cast manager initialization on the existing Google Cast application-ID Info.plist marker.
- Preserve Cast initialization and configured options when Google Cast is enabled.
- Add an unreleased changelog entry.

## Validation

- Podspec syntax and whitespace validation passed.
- SwiftLint could not complete because the environment could not save its temporary plist and reports unrelated existing violations.
- TypeScript checks were unavailable because dependencies are not installed in the checkout.
- No native XCTest or plugin test harness exists for this app-delegate startup path.

## Checklist

- [x] 🗒 `CHANGELOG` entry

### DIFF
--- a/ios/AppLifecycleDelegate.swift
+++ b/ios/AppLifecycleDelegate.swift
@@ -17,12 +17,10 @@ public class AppLifecycleDelegate: ExpoAppDelegateSubscriber {
         }
 
         // Only initialize Cast when it is configured via the Expo plugin.
-        if infoDictionary["BitmovinPlayerGoogleCastApplicationId"] is String,
+        if let applicationId = infoDictionary["BitmovinPlayerGoogleCastApplicationId"] as? String,
            !BitmovinCastManager.isInitialized() {
             let options = BitmovinCastManagerOptions()
-            if let applicationId = infoDictionary["BitmovinPlayerGoogleCastApplicationId"] as? String {
-                options.applicationId = applicationId
-            }
+            options.applicationId = applicationId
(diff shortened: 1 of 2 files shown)


ios/AppLifecycleDelegate.swift
```diff
@@ -16,11 +16,11 @@ public class AppLifecycleDelegate: ExpoAppDelegateSubscriber {
             OfflineManager.initializeOfflineManager()
         }
 
-        if !BitmovinCastManager.isInitialized() {
+        // Only initialize Cast when it is configured via the Expo plugin.
```
123mpozzi: Nit: This ties the behavior to the Expo plugin, but the runtime check also supports manually configured `Info.plist` values. Could we capture the reason instead?


```suggestion
        // The Cast SDK is optional; only attempt initialization when Cast configuration is present.
```

123mpozzi:
> No native XCTest or plugin test harness exists for this app-delegate startup path.

Nit: Both current iOS test apps enable Cast, so CI does not exercise this exact startup path. We could add a small no-Cast simulator boot smoke? The existing Cavy boot timeout should detect a crash before JavaScript mounts

123mpozzi (approved):
Good find!!

PR #1037, opened 2026-08-05
Prevent opening new terminal windows during integration tests

123mpozzi:
## Description

Integration test runs currently ask Cavy/React Native to open a separate terminal for Metro. This keeps Metro in the terminal that started the tests.

The launch scripts reuse a matching integration-test Expo packager or start one in the background. They clean up only the packager they own when the test command exits.

## Changes

- pass `--no-packager` to Cavy so React Native does not open a separate terminal for Metro
- start or reuse the integration-test packager on the configured Metro port, using `--port`, `RCT_METRO_PORT`, or `8081` by default
- reject listeners that do not belong to the current integration-test worktree
- pass the configured port through pre-run cleanup so stale packagers cannot affect the selected test suites
- prefer a booted iOS simulator or boot and wait for one before starting Metro and Cavy
- start or reuse an Android emulator and run Cavy against its device ID
- cover packager ownership, cleanup, custom ports, device setup, and argument forwarding in the existing TypeScript CI job

## Validation

- `yarn integration-test check:runner` (45 tests)
- `RCT_METRO_PORT=9090 yarn integration-test test:runner`

Manual runs:

```
yarn integration-test test:android --tags playback
yarn integration-test test:ios --tags playback
```

No new Metro terminal should open. Test results should stay in the terminal that started the command.


integration_test/scripts/packager-utils.sh
```diff
@@ -0,0 +1,123 @@
+#!/bin/bash
+
+SCRIPT_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd -P)
+INTEGRATION_TEST_DIR=$(cd "$SCRIPT_DIR/.." && pwd -P)
+PACKAGER_PORT=8081
```
rolandkakonyi: [P2] This fixed port breaks the runner's documented argument-forwarding contract. For example, `yarn integration-test test:ios --port 9090` starts Expo on 8081 here, but forwards `--port 9090` to `react-native run-ios`, so the app looks for Metro on 9090 and cannot load the bundle. `RCT_METRO_PORT` has the same mismatch. Please derive the managed packager port from the forwarded CLI option/environment (with 8081 as the default) and cover both forms in the launcher tests.

integration_test/scripts/script-test-harness.js
```diff
@@ -0,0 +1,322 @@
...
+  printf "%s\\n" "node $cwd/node_modules/expo/bin/cli start --port ${PACKAGER_PORT} --localhost"
+  exit 0
+fi
+printf "%s\\n" "$STUB_PROCESS_LIST"
+`
+  );
+
+  writeFakeCommand(
+    binDir,
+    'sleep',
```
rolandkakonyi, line 128: [P3] Stubbing `sleep` as an immediate exit makes the fake Expo process's `while true; do sleep 1; done` loop spin continuously for every `STUB_EXPO_START_MODE=hold` test. Keep a real short blocking wait in this stub so these tests do not peg a CPU core. This is separate from the existing Copilot comment about `waitForFile` also busy-waiting.

rolandkakonyi (changes requested):
The custom Metro-port contract is currently broken: the managed packager and the React Native app can be configured to use different ports. Please address that before merging. I also left a separate non-blocking test-harness comment.

integration_test/scripts/script-test-harness.js
123mpozzi: Fixed in e061d9ac. The fake sleep now blocks briefly, so the held Expo process no longer spins continuously.

integration_test/scripts/packager-utils.sh
123mpozzi: Good catch. Fixed in d00bd31b and e45a783c. Cleanup and startup now use `--port` or `RCT_METRO_PORT`, with `8081` as the default. Both forms are covered by the launcher tests.

integration_test/scripts/script-test-harness.js
```diff
@@ -0,0 +1,322 @@
...
+  const start = Date.now();
+
+  while (Date.now() - start < timeoutMs) {
+    if (fs.existsSync(filePath)) {
+      return true;
+    }
+  }
+
+  return false;
+}
```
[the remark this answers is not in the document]
123mpozzi: Fixed in e061d9ac. waitForFile now blocks briefly between checks instead of busy-waiting.

rolandkakonyi (approved):
Verified the follow-up fixes on the current head: custom Metro ports are propagated through startup and cleanup, the fake packager sleep blocks, and waitForFile yields between polls. The focused runner suite passes all 45 tests, shell syntax checks pass, and the latest CI checks are green.

PR #1066, opened 2026-09-08
Expose Android subtitle bitmap height

strangesource:
## Description

Expose the Android Player SDK's `Cue.bitmapHeight` value through `CueEnterEvent.layout` and `CueExitEvent.layout` for custom image-based TTML/IMSC subtitle rendering.

## Changes

- Add the optional Android-only `SubtitleCueLayout.bitmapHeight` field.
- Normalize the native viewport fraction to a `0`–`100` percentage while omitting `Cue.DIMEN_UNSET`.
- Add an Android end-to-end test using an embedded-image TTML fixture.
- Add a changelog entry.

## Checklist

- [x] 🗒 `CHANGELOG` entry


android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
```diff
@@ -1112,6 +1112,7 @@ private fun Cue.toLayoutJson(): Map<String, Any>? {
     val line = toLayoutLineJson()
     val position = fractionalPosition.takeIf { it != Cue.DIMEN_UNSET }?.toPercent()
     val size = size.takeIf { it != Cue.DIMEN_UNSET }?.toPercent()
+    val bitmapHeight = bitmapHeight.takeIf { it != Cue.DIMEN_UNSET }?.toDouble()
```
123mpozzi: Should this use `.toPercent()` rather than `.toDouble()`? The existing normalized layout contract converts Android’s native fractional `position`, `size`, and percentage-based `line` values into 0–100 percentages. Keeping `bitmapHeight` as 0–1 would give adjacent geometry fields different numeric scales.

For example, the fixture’s native `0.12` would be exposed as `12`, matching `size` and `position`. If we agree, the TS docs, PR description, and integration-test expectation should be updated too.

src/events.ts
```diff
@@ -794,6 +794,17 @@ export interface SubtitleCueLayout {
    * Width of the cue box as a percentage of the viewport dimension in the writing direction.
    */
   size?: number;
+  /**
+   * Height of an image/bitmap cue as a fraction of the viewport size (typically `0`–`1`).
+   *
+   * Only meaningful for image-based cues (e.g. IMSC1 image profile / DVB subtitles). It is
+   * omitted for text cues, and also for bitmap cues that should be displayed at their natural
+   * height. Because it is a viewport fraction rather than a pixel height, it is not equivalent
+   * to measuring the decoded bitmap exposed via the event's `image` field.
```
123mpozzi: Thanks for adding context! 💪

integration_test/tests/captionMetadataTest.ts
```diff
@@ -110,6 +132,39 @@ export default (spec: TestScope) => {
...
+
+              const cueEnterEvent =
+                await callPlayerAndExpectEvent<CueEnterEvent>(
+                  (player) => player.seek(1),
+                  FilteredEvent<CueEnterEvent>(
+                    EventType.CueEnter,
+                    (event) => event.layout?.bitmapHeight !== undefined
+                  )
+                );
+              expectBitmapCue(cueEnterEvent, 'CueEnter');
```
123mpozzi, line 158: nit: Could we also assert in the existing Android text-cue test that `bitmapHeight` is omitted when unavailable?

```ts
expect(
  cueEnterEvent.layout?.bitmapHeight,
  'Text cues should omit bitmapHeight'
).toBeUndefined();
```

123mpozzi (commented):
Thanks for taking care of this! 🚀  just left a few comments

android/src/main/java/com/bitmovin/player/reactnative/converter/JsonConverter.kt
strangesource: Updated to use `toPercent()`; the public docs, PR description, and end-to-end expectation now use the normalized `0`–`100` scale.

integration_test/tests/captionMetadataTest.ts
strangesource: Added the omission assertion to the shared Android text-cue layout helper, so it now covers both `CueEnter` and `CueExit`.

123mpozzi (approved):
Looking good! 💪

PR #1075, opened 2026-09-22
Migrate iOS Player and IMA dependencies to Swift Package Manager

jarhoax:
## Description

Move native Player and Google IMA dependencies to Swift Package Manager so the React Native SDK can consume releases without CocoaPods distribution. The Expo plugin also attaches package products to the application target so Xcode embeds the dynamic frameworks required at launch.

This is a breaking installation change: apps using Expo prebuild must rerun it after upgrading. Manually maintained apps must set `BITMOVIN_APPLE_PLATFORM` before `pod install` and add the SPM packages to their application targets. See the [iOS and tvOS migration guide](docs/ios-dependencies.md).

Addresses [PRN-224](https://bitmovin.atlassian.net/browse/PRN-224).

Builds on @stonko1994's PoCs: [SPM dependency migration (#1042)](https://github.com/bitmovin/bitmovin-player-react-native/pull/1042) and [app-target framework embedding (#1043)](https://github.com/bitmovin/bitmovin-player-react-native/pull/1043).

## Changes

- Resolve Player `3.123.0` and IMA (`3.26.1` on iOS, `4.15.1` on tvOS) from a shared manifest read by the podspec and Expo plugin. IMA versions remain aligned with the native Player SDK.
- Configure app targets during prebuild, avoid duplicate references, update plugin-owned dependencies, and preserve customer-managed or shared references.
- Keep the React Native/Expo bridge and optional Google Cast dependency on CocoaPods.
- Run plugin build and prebuild on CI cache misses or partial restores, and check embedded SDK framework binaries in the iOS and tvOS example apps.
- Add tests for the iOS SDK version updater. Separate integrator migration instructions from maintenance and testing guidance in `CONTRIBUTING.md`.

## Validation

- All 13 config-plugin tests and 50 Python automation tests pass.
- `yarn build`, `yarn typecheck:all`, and `yarn lint` pass.
- iOS and tvOS simulator builds succeeded.
- Physical iPhone app launch, playback, and ads were tested successfully. 

## Checklist

- [x] 🗒 `CHANGELOG` entry

[PRN-224]: https://bitmovin.atlassian.net/browse/PRN-224?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ios/dependencies.json
```diff
@@ -0,0 +1,19 @@
+{
+  "player": {
+    "url": "https://github.com/bitmovin/player-ios.git",
+    "version": "3.123.0",
+    "product": "BitmovinPlayer"
+  },
+  "ima": {
+    "ios": {
+      "url": "https://github.com/googleads/swift-package-manager-google-interactive-media-ads-ios",
+      "version": "3.31.0",
```
123mpozzi: IMA version in native SDK is `3.26.1` ⚠️ - Since we build native against that version, this can be breaking!

ios/dependencies.json
```diff
@@ -0,0 +1,19 @@
...
+      "product": "GoogleInteractiveMediaAds"
+    },
+    "tvos": {
+      "url": "https://github.com/googleads/swift-package-manager-google-interactive-media-ads-tvos",
+      "version": "4.16.0",
```
123mpozzi: Same here, we cannot declare a higher version than native since native is build against an older one: `4.15.1`!

ios/RNBitmovinPlayer.podspec
```diff
@@ -28,9 +28,23 @@ Pod::Spec.new do |s|
   s.static_framework = true
 
   s.dependency 'ExpoModulesCore'
-  s.dependency "BitmovinPlayer", "3.123.0"
-  s.ios.dependency "GoogleAds-IMA-iOS-SDK", "3.26.1"
-  s.tvos.dependency "GoogleAds-IMA-tvOS-SDK", "4.15.1"
```
123mpozzi: Here you can see the older configuration matched native's versions

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
 
 ### Changed
 
+- iOS: Resolve Player and Google IMA through Swift Package Manager, with automatic app integration during Expo prebuild
+- iOS: Update Google IMA to `3.31.0`; tvOS: Update Google IMA to `4.16.0`
+
```
123mpozzi: can we remove the whiteline?
```suggestion
```

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
 
 ### Changed
 
+- iOS: Resolve Player and Google IMA through Swift Package Manager, with automatic app integration during Expo prebuild
+- iOS: Update Google IMA to `3.31.0`; tvOS: Update Google IMA to `4.16.0`
```
123mpozzi: Here: since we should not update these dependencies as I explained above, we should remove the entry as well
```suggestion
```

ios/RNBitmovinPlayer.podspec
```diff
@@ -28,9 +28,23 @@ Pod::Spec.new do |s|
...
+    spm_dependency(
+      s,
+      url: dependency.fetch('url'),
+      requirement: { kind: 'exactVersion', version: dependency.fetch('version') },
+      products: [dependency.fetch('product')]
+    )
+  end
 
   if podfile_properties['BITMOVIN_GOOGLE_CAST_SDK_VERSION'].to_s != ''
     s.ios.dependency "google-cast-sdk", podfile_properties['BITMOVIN_GOOGLE_CAST_SDK_VERSION'].to_s
```
123mpozzi: Why was google cast SDK not migrated? This is inconsistent and would break once we diverge natively.

Was this on purpose? I know Google is taking its time to update Cast to SPM, but we can consider alternatives like https://github.com/SRGSSR/google-cast-sdk rather than ignoring the dependency switch

.github/scripts/update_ios_sdk_version.py
123mpozzi: Can we cover this logic? I'd at least cover a valid update, an unchanged version, an invalid version, and preservation of the IMA entries.

The repo already has [Python script tests](https://github.com/bitmovin/bitmovin-player-react-native/blob/development/.github/workflows/ci-github-scripts.yml) discovered by `test_*.py`

.github/workflows/ci-ios-tvos.yml
```diff
@@ -108,23 +108,34 @@ jobs:
           ios: true
           skip-prebuild: true
 
-      - name: Prebuild iOS Example (on cache miss)
-        if: ${{ steps.ios-example-prebuild-cache.outputs.cache-hit != 'true' }}
-        run: .github/scripts/smart-prebuild.sh example
+      - name: Build config plugin
+        run: yarn build:plugin
+
+      - name: Configure iOS Example
```
123mpozzi: Could we make both "Build config plugin" and "Configure …" conditional on an exact `prebuild-cache` miss in all three jobs?

The cache key includes the plugin and native-project inputs, so an exact hit already has the generated Xcode project. We should still run the direct Expo prebuild command on misses and partial restores.

The old `smart-prebuild.sh` can skip a partial restore merely because a Podfile exists.

.github/workflows/ci-ios-tvos.yml
```diff
@@ -108,23 +108,34 @@ jobs:
...
 
+      - name: Resolve Swift packages
+        run: xcodebuild -resolvePackageDependencies -workspace example/ios/BitmovinPlayerReactNativeExample.xcworkspace -scheme BitmovinPlayerReactNativeExample
+
+      - name: Save resolved Swift package versions
+        uses: actions/upload-artifact@v4
+        with:
+          name: ios-example-swift-packages
+          path: example/ios/BitmovinPlayerReactNativeExample.xcworkspace/xcshareddata/swiftpm/Package.resolved
+
```
123mpozzi: Could we add a post-build check that the required dynamic frameworks are present in the example `.app`, for iOS and tvOS? The plugin tests check project references, while the failure described in the plugin (`withBitmovinIosDependencies.ts` L3-8) can still pass compilation and fail at launch

.github/scripts/update_ios_sdk_version.py
123mpozzi: nit: Splitting validation, path construction, reading, and updating into named functions would make the intent easier to scan - can we at least add a few white lines to split up the flow?

docs/apple-dependencies.md
```diff
@@ -0,0 +1,59 @@
...
+   exact versions. Add their named products to the application target's
+   Frameworks, Libraries, and Embedded Content. Let Xcode embed the dynamic
+   frameworks supplied by these products.
+4. Build and launch the app. A successful compilation alone does not verify
+   framework embedding. Repeat this check for Release archives and tvOS if used.
+
+Repeat the package-version updates when upgrading the library. Apps with multiple
+application targets must integrate the products into each target using the bridge.
+
+## Maintaining the integration
```
123mpozzi: Who's the target of this new documentation file (and folder)? Is it the integrators, or is it the maintainers of the project?

The README explicitly links to it, so integrators could end up here, but this note is useless to them, unless they're also maintainers.

Should it be moved in `CONTRIBUTING.md` instead? E.g. we can add a short "Updating the native iOS SDK" section after the existing "Scripts" section

docs/apple-dependencies.md
```diff
@@ -0,0 +1,59 @@
...
+
+Use `python3 .github/scripts/update_ios_sdk_version.py <version>` to update the Player
+pin. The SDK update workflow uses this command too. Change IMA pins in the same
+manifest when updating IMA.
+
+Run `yarn test:plugin` to check repeated prebuilds, SDK upgrades, platform switching,
```
123mpozzi: This could move to Testing section, specifically, of `CONTRIBUTING.md`  as it fits better there

docs/apple-dependencies.md
```diff
@@ -0,0 +1,59 @@
...
+3. In Xcode, add the Player package and the appropriate IMA package from
+   `node_modules/bitmovin-player-react-native/ios/dependencies.json`, using their
+   exact versions. Add their named products to the application target's
+   Frameworks, Libraries, and Embedded Content. Let Xcode embed the dynamic
+   frameworks supplied by these products.
+4. Build and launch the app. A successful compilation alone does not verify
+   framework embedding. Repeat this check for Release archives and tvOS if used.
+
+Repeat the package-version updates when upgrading the library. Apps with multiple
+application targets must integrate the products into each target using the bridge.
```
123mpozzi: Could we separate the initial migration from later upgrades? Projects using the config plugin need to run Expo prebuild after upgrading to this release. Projects maintained without prebuild need to set the platform and add the SPM packages manually for this migration.


On later releases, they should check the release notes and `dependencies.json` for changes to their native integration. The current "Repeat the package-version updates" sentence does not make that distinction clear.

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
 
 ### Changed
 
+- iOS: Resolve Player and Google IMA through Swift Package Manager, with automatic app integration during Expo prebuild
```
123mpozzi: Could we call out the upgrade action here and link the migration guide? Apps using Expo prebuild need to rerun it after upgrading to this release. Apps maintained without prebuild need the manual SPM migration described in the guide. In particular, the new podspec requires `BITMOVIN_APPLE_PLATFORM` before `pod install`.

This reads like a routine dependency change at present, while the repo has used a "Breaking Change" entry for an installation migration before.

123mpozzi (changes requested):

ios/dependencies.json
jarhoax: interesting, I've tested this locally and Ads worked just fine 🤔 Will downgrade though

123mpozzi: It can work fine, but it can break if the symbols we're building against natively mismatch with newer versions - it depends on what's changed on IMA's side

ios/RNBitmovinPlayer.podspec
jarhoax: I'm not sure if we should use a third party wrapper to have SPM support for Google Cast. We can still use it from CocoaPods without much hassle until there is official support. 

So yes, this was intentional.

ios/dependencies.json
jarhoax: Yeah you're definitely right, good catch 👍

jarhoax: 2c146274cec8f92a6e4cfaf0ac52b69bb187440d

ios/dependencies.json
jarhoax: 2c146274cec8f92a6e4cfaf0ac52b69bb187440d

ios/RNBitmovinPlayer.podspec
jarhoax: 2c146274cec8f92a6e4cfaf0ac52b69bb187440d

CHANGELOG.md
jarhoax: e9b713adc815fdd66f8ad0109ff41a784f3426ed

CHANGELOG.md
jarhoax: e9b713adc815fdd66f8ad0109ff41a784f3426ed

.github/scripts/update_ios_sdk_version.py
jarhoax: 331d979099efaa892fff4feb195e076e3a0c56d2

.github/workflows/ci-ios-tvos.yml
jarhoax: 74adf64208cc8999c801dcb5ed099792f9cc8a77

.github/workflows/ci-ios-tvos.yml
jarhoax: acd1c445d96822b0fe7cfcce5953942f0d31130f

.github/scripts/update_ios_sdk_version.py
jarhoax: 331d979099efaa892fff4feb195e076e3a0c56d2

docs/apple-dependencies.md
jarhoax: 74fab3d23e844b3d9373dae3cc0d937981880651

docs/apple-dependencies.md
jarhoax: 74fab3d23e844b3d9373dae3cc0d937981880651

docs/apple-dependencies.md
jarhoax: 74fab3d23e844b3d9373dae3cc0d937981880651

CHANGELOG.md
jarhoax: e9b713adc815fdd66f8ad0109ff41a784f3426ed

.github/scripts/update_ios_sdk_version.py
123mpozzi: Thanks, much clearer! 🙌

CHANGELOG.md
123mpozzi: This works, but it differs in style from the past Breaking Change style in `1.0.0` , which was a `## Breaking Change` section if I'm not wrong.

Not a blocker

jarhoax: Yes I know, but that does not adhere to my knowledge of keep-a-changelog 😅 :

```
### Changed
- **Breaking:** Retired the tagline "Don't let your friends dump git logs into changelogs" for "Clearly document the evolution of your projects." Earlier versions keep the original.
```

docs/ios-dependencies.md
```diff
@@ -0,0 +1,61 @@
+# iOS SDK dependencies
+
+This guide covers iOS and tvOS apps integrating the React Native Player SDK.
+Player and Google IMA are installed through Swift Package Manager (SPM).
+CocoaPods still installs the React Native bridge, ExpoModulesCore, and optional
+Google Cast SDK.
+
+## Migrating an existing app to this release
```
123mpozzi: Could we name the migration instead of referring to "this release", which is going to be stale pretty soon?

This guide will still be used by customers upgrading from a pre-SPM version after future releases ship. "Migrating from CocoaPods to SPM" and "Upgrading after the migration" would keep both sections clear.

CHANGELOG.md
123mpozzi: Fair take, then fine to keep as is 👍

docs/ios-dependencies.md
jarhoax: Good catch, will do

jarhoax: df56dcc2cea7e0ef82a26e157d4904a58d0ab645

123mpozzi (approved):
All good to go! Thanks for addressing all feedback so quickly! Please wait for CI checks though before merging

PR #1076, opened 2026-09-23
Use Player SDKs email for release automation

jarhoax:
## Description
Release automation currently publishes commits using the obsolete support email. Change to `player-sdks@bitmovin.com`

Related: https://bitmovin.atlassian.net/browse/PI-2277

## Changes
Update release workflows, the release-sync author-email guard, and its test fixtures together. 

Validation: all four release-sync tests and shell syntax checks pass; no obsolete email matches remain.


123mpozzi (approved):
Thanks for taking care of this!

matamegger (approved):
There are many occurrences.

Could we easily move it into an environment variable?

jarhoax:
> There are many occurrences.
> 
> Could we easily move it into an environment variable?

good point, will check

.github/scripts/test_push_finish_release_sync_branch.py
```diff
@@ -160,6 +162,7 @@ def push_branch(self, repository_path: Path) -> None:
     def run_script(self, repository_path: Path) -> subprocess.CompletedProcess[str]:
         return subprocess.run(
             ["bash", str(SCRIPT_PATH), BRANCH_NAME],
+            env={**os.environ, "RELEASE_AUTOMATION_EMAIL": AUTOMATION_EMAIL},
```
matamegger: Maybe let's not use a non existing default. I guess failing if the environment variable/argument is missing would be fine as well.

.github/workflows/finish-release-train.yml
```diff
@@ -69,7 +70,7 @@ jobs:
       - name: Setup Git User
         run: |
           git config --global user.name "Bitmovin Release Automation"
-          git config --global user.email "support@bitmovin.com"
+          git config --global user.email "$RELEASE_AUTOMATION_EMAIL"
```
matamegger: Seems inconsistent.

Why checking and failing with a message on the one job, but not on the other.

matamegger (commented):
Nice!

Just some small cleanup please :)

.github/workflows/finish-release-train.yml
jarhoax: I found it sufficient to fail on the prepare if its missing, no need to repeat the message in the other usages imo

matamegger: Fair

.github/scripts/test_push_finish_release_sync_branch.py
jarhoax: that's the test explicitely providing a value for it 👀 Anything I'm missing?

matamegger: Sorry 😅

matamegger (approved):

.github/scripts/test_push_finish_release_sync_branch.py
jarhoax: all good 😅 🫶

